PR #3, opened 2020-01-06
NIST curve crate boilerplate

tarcieri:
Boilerplate for the `p256`, `p384`, and `p521` crates


newpavlov:
It may be worth to also add license files. (we probably should unify them across all crates one day...)

PR #15, opened 2020-04-18
p256: Group operation

str4d:
Implements the complete addition formulas from [Renes-Costello-Batina 2015](https://eprint.iacr.org/2015/1060).


paulmillr:
What is the benefit of using those new formulas instead of classical projective formulas via hyperelliptic from 2008 etc?

str4d:
They aren't that much slower, and may in fact be competitive on some platforms.

The formulas from Renes-Costello-Batina 2015 have the following costs:
- Projective addition: 12M+ 2m<sub>b</sub>+ 29a
- Mixed addition: 11M+ 2m<sub>b</sub>+ 23a
- Doubling: 8M+ 3S+ 2m<sub>b</sub>+ 21a

According to the [Explicit-Formulas Database](https://www.hyperelliptic.org/EFD/g1p/auto-shortw-projective-3.html) (also shown in Table 1 of Renes-Costello-Batina 2015), the most efficient formulas for each case are (counting field element doublings and triplings as additions):
- Projective addition: 12M + 2S + 7a (Cohen–Miyaji–Ono 1998)
- Mixed addition: 9M + 2S + 7a (Cohen–Miyaji–Ono 1998)
- Doubling: 7M + 3S + 11a (Bernstein–Lange 2007)

Currently S and m<sub>b</sub> are implemented in terms of M, so RCB has the same multiplication cost as CMO for regular addition; with improved implementations of the field arithmetic, RCB would use fewer general multiplications than CMO. The tradeoff is the much higher number of additions, but that may work in RCB's favour on low-end platforms where the M/a ratio is higher (multiplications being more costly).

In any case, my goal for this implementation is to be safe and understandable first, followed by fast. The benefits of working with complete addition formulas IMHO outweighs the performance penalty (maybe 1.4x slower than it could be going from RCB's numbers), and I'd prefer that we have a quality base implementation, and then benchmark and optimise where it makes sense.

str4d:
Rebased on master because GitHub was complaining the PR was out of date (despite there being no merge conflicts).

str4d:
The PR that adds scalar multiplication will include tests against the [Project Wycheproof ECPoint test vectors for ASN encoded points](https://github.com/google/wycheproof/blob/master/testvectors/ecdh_secp256r1_ecpoint_test.json), which should catch any missing edge cases.

PR #18, opened 2020-04-25
p256: Constant-time scalar multiplication

str4d:
Implements the basic double-and-add formula.


str4d:
I implemented w-NAF scalar multiplication on today's stream, but disabled no-std for debugging and then completely forgot that `Vec` requires allocations 😅 so I need to rework it to use fixed-size arrays. That can be in a subsequent PR.

tarcieri:
@str4d you can always use `heapless::Vec` if you’d like `Vec`-like ergonomics but with fixed capacities

PR #19, opened 2020-04-27
Implement secp256k1 with field arithmetic, point addition/doubling, and scalar multiplication

tuxxy:
This introduces changes made by @str4d for p256, but modifies them for
secp256k1.

Additionally, this implementation primarily uses montgomery multiplication
rather than schoolbook multiplication + a montgomery reduction for added
performance gains. However, it lacks a more efficient squaring
implementation.

EDIT: I've expanded the scope of the PR to incorporate point addition/doubling. It also includes recent additions from @str4d for scalar multiplication. [This PR has a strong "draw the rest of the fucking owl" vibe to it.](https://external-preview.redd.it/DodWFQ9mQkVyWoKFa0ZIu12PYrPo3P2T0taaK-lgJCo.png?auto=webp&s=c180684f48b01ff6f2cbc72e080067039943de07)

For the point arithmetic, I used algorithms 7 (addition), 8 (mixed addition) and 9 (doubling) from the [Renes-Costello-Batina 2015 paper](https://eprint.iacr.org/2015/1060) that @str4d also used.


tarcieri:
First, thanks for submitting this! Looks good at first glance.

Unfortunately, unless it's GF(2^128) I'm not a good person to review field arithmetic code. It'd be nice if @str4d could glance over it.

Another thing you could do which would improve my confidence in the implementation would be to `proptest` and/or fuzz test it against the [secp256k1 field implementation from `fiat-rust`](https://github.com/mit-plv/fiat-crypto/blob/master/fiat-rust/src/secp256k1_64.rs).

hdevalence:
This is super cool!

Montgomery arithmetic doesn't require any special properties of the modulus, but the secp256k1 modulus has somewhat of a nice form: p = 2^256 - 2^32 - 977, so 2^256 = 977 + 2^32 mod p. So I wonder whether it would be higher-performance to write a multiply-and-reduce implementation that uses the special structure, rather than using Montgomery arithmetic.  If you'd be interested in working on that I'd be happy to help.

tuxxy:
Yes, that would be great to test! @fjarri from our team would also be
interested in keeping up with that.

tuxxy:
I have implemented the point addition algorithms and incorporated @str4d's work on scalar multiplication now. So I'm going to be expanding the scope of this PR in a second (cause why not?).

tuxxy:
@tarcieri As of 0696d7f I've added proptests that leverage fiat-crypto. Does that look like about what you expected?

tarcieri:
Yep, looks great! I'll go ahead and merge this.

k256/src/arithmetic.rs
```diff
@@ -0,0 +1,730 @@
+//! A pure-Rust implementation of group operations on secp256k1.
+//!
+//! # Status
+//!
+//! Currently, no actual group operations are implemented. Only point compression and
+//! decompression is supported.
```
daira, line 6: This comment seems obsolete.

Issue #22, opened 2020-04-28
Generic curve arithmetic


tarcieri:
One of the goals of the `elliptic-curve` crate is to be able to write code that's generic over (at least Weierstrass) elliptic curve types.

Now that the `p256` and `k256` crates have arithmetic implementations (thanks @str4d and @tuxxy!), an open question is how to write code that's generic over the underlying `AffinePoint` and/or `ProjectivePoint` types.

It seems like `AffinePoint` and `ProjectivePoint` could benefit from having traits which at the very least are bounded by the point arithmetic they must support.

It also seems like there needs to be some trait connecting the two of them which can be used to access the `From` impls and arithmetic (e.g. `Add`) between the two different point types.

Finally, it seems like there needs to be a trait a `weierstrass::Curve` type can impl which provides that curve's `AffinePoint` and `ProjectivePoint` types as associated types.

Curious if people think this is a good idea and what those traits might look like.

cc @tuxxy

tarcieri:
@tuxxy brought up scalars on #23 

It might make sense to replace the existing `Scalar` struct in the `elliptic-curve` crate with a trait as well:

https://docs.rs/elliptic-curve/0.3.0/elliptic_curve/scalar/struct.Scalar.html

tuxxy:
I have some curve generic traits that I used for a library I am building with @nucypher called Twinkle. I've changed a lot of the backend since, but I have some trait definitions for `Scalar` and `Point`.

As an example, here's the definition for `Point`:

```rust
pub trait Point: Sized {
    type Backend: Backend<Point = Self>;
    /// This can be used to swap out the library point type with a custom type.
    /// If no such swapping is done, this should simply be Self.
    type BackendPoint: Sized;

    // Arithmetic functions
    fn point_mul(lhs: <Self::Backend as Backend>::Scalar, rhs: Self) -> Self;
    fn point_add(lhs: Self, rhs: Self) -> Self;
    fn point_sub(lhs: Self, rhs: Self) -> Self;
    fn point_neg(point_val: Self) -> Self;

    fn point_eq(lhs: &Self, rhs: &Self) -> bool;

    // Constants
    fn basepoint() -> Self;
    fn identity() -> Self;

    // Helper functions
    fn new() -> Self {
        Self::identity()
    }
    fn from_backend_point(point: Self::BackendPoint) -> Self;
    fn get_backend_point(self) -> Self::BackendPoint;
}
```

[...]

PR #26, opened 2020-04-30
p256 v0.2.0

tarcieri:
## Added
- Constant time scalar multiplication ([#18])
- Group operation ([#15])

[#18]: https://github.com/RustCrypto/elliptic-curves/pull/18
[#15]: https://github.com/RustCrypto/elliptic-curves/pull/15


tarcieri:
If @str4d approves I can do the same for `k256` /cc @tuxxy 

After that is probably a good time to circle back on a new set of traits or pulling them in from other crates (#22)

tarcieri:
:shipit:

PR #27, opened 2020-05-01
k256 v0.2.0

tarcieri:
### Added
- Field arithmetic, point addition/doubling, and scalar multiplication ([#19])

[#19]: https://github.com/RustCrypto/elliptic-curves/pull/19


tarcieri:
cc @tuxxy

PR #36, opened 2020-05-26
Implemented SecretKey -> Scalar conversion in p256

rozbb:
This is just a small ergonomics issue I ran into when implementing [hpke](https://github.com/rozbb/rust-hpke)

cc @str4d

### DIFF
--- a/p256/src/arithmetic/scalar.rs
+++ b/p256/src/arithmetic/scalar.rs
@@ -45,10 +45,8 @@ impl Scalar {
     ///
     /// Returns None if the secret's underlying value does not represent a field element.
     pub fn from_secret(s: SecretKey) -> CtOption<Scalar> {
-        let mut buf = [0u8; 32];
-        buf.copy_from_slice(s.secret_scalar().as_ref());
         // We can't unwrap() this, since it's not guaranteed that s represents a valid field elem
-        Self::from_bytes(buf)
+        Self::from_bytes((*s.secret_scalar().as_ref()).into())
     }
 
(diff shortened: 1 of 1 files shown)


p256/src/arithmetic/scalar.rs
```diff
@@ -40,6 +40,17 @@ impl Scalar {
...
 
+    /// Attempts to convert a `SecretKey` (defined in the more generic `elliptic_curve` crate) to a
+    /// `Scalar`
+    ///
+    /// Returns None if the secret's underlying value does not represent a field element.
+    pub fn from_secret(s: SecretKey) -> CtOption<Scalar> {
+        let mut buf = [0u8; 32];
+        buf.copy_from_slice(s.secret_scalar().as_ref());
+        // We can't unwrap() this, since it's not guaranteed that s represents a valid field elem
+        Self::from_bytes(buf)
```
tarcieri, line 51: I believe you can write this as...

```suggestion
        Self::from_bytes(s.secret_scalar().into())
```

...or failing that:

```suggestion
        Self::from_bytes(s.secret_scalar().as_ref().try_into().unwrap())
```

rozbb: Neither of these worked for me. It doesn't look like `elliptic_curve::Scalar` implements `Into` or `TryInto` for _anything_.

rozbb: Fixed!

tarcieri:
Thanks!

PR #49, opened 2020-06-19
Add `weierstrass::FixedBaseScalarMul` trait

tarcieri:
Adds a trait for fixed-base scalar multiplication which accepts `&ScalarBytes` as input and returns an associated point type, whose bounds allow for a `From` conversion to either the `CompressedCurvePoint` or `UncompressedCurvePoint` for a given curve.

Using this trait, a `weierstrass::PublicKey<C>::from_secret_key` method is conditionally implemented when the curve `C` impls the `FixedBaseScalarMul` trait, allowing generic computation of a public key from a secret key, with optional point compression (selected via a `compress` argument).


tarcieri:
@str4d @tuxxy pushing this up early for a bit of design review. I implemented `FixedBaseScalarMul` for `k256` so far (but not for `p256` yet), which is interesting because it demonstrates not all curves are obliged to implement it which is nice.

TODO: test with test vectors before merging

elliptic-curve-crate/src/weierstrass.rs
```diff
@@ -4,8 +4,30 @@ pub mod curve;
...
+use subtle::{ConditionallySelectable, CtOption};
+
+/// Fixed-base scalar multiplication
+pub trait FixedBaseScalarMul: Curve
+where
+    <Self::ScalarSize as Add>::Output: Add<U1>,
+    CompressedCurvePoint<Self>: From<Self::Point>,
+    UncompressedCurvePoint<Self>: From<Self::Point>,
+    CompressedPointSize<Self::ScalarSize>: ArrayLength<u8>,
+    UncompressedPointSize<Self::ScalarSize>: ArrayLength<u8>,
```
tarcieri, line 25: Oh `generic-array` trait bounds 😢

tarcieri, line 25: Oh `generic-array` trait bounds 😢

elliptic-curve-crate/src/weierstrass.rs
```diff
@@ -4,8 +4,30 @@ pub mod curve;
...
+{
+    /// Elliptic curve point type
+    type Point: ConditionallySelectable + Default;
```
tarcieri, line 28: The `Default` bound is necessitated by the current use of `CtOption`, namely `and_then`/`map`:

https://docs.rs/subtle/2.2.3/subtle/struct.CtOption.html#method.map

I think it'd be nice to get rid of it if possible...

k256/src/arithmetic.rs
```diff
@@ -53,6 +56,12 @@ impl PartialEq for AffinePoint {
     }
 }
 
+impl Default for AffinePoint {
+    fn default() -> AffinePoint {
+        AffinePoint::generator()
+    }
+}
```
tarcieri, line 63: As [mentioned above](https://github.com/RustCrypto/elliptic-curves/pull/49/files#r442975417), this was necessitated by the current usage of `CtOption`. I'd be interested in finding a workaround so it isn't necessary.

That said, this implementation seems reasonable to me...

k256/src/arithmetic.rs
```diff
@@ -124,36 +133,66 @@ impl AffinePoint {
...
 
+    /// Returns a [`PublicKey`] with the SEC-1 compressed encoding of this point.
+    pub fn to_compressed_pubkey(&self) -> PublicKey {
+        PublicKey::Compressed(self.clone().into())
+    }
+
+    /// Returns a [`PublicKey`] with the SEC-1 uncompressed encoding of this point.
+    pub fn to_uncompressed_pubkey(&self) -> PublicKey {
+        PublicKey::Uncompressed(self.clone().into())
+    }
```
tarcieri, line 144: I changed these to return an actual `PublicKey`, and moved their previous implementations which return a `CompressedCurvePoint` or `UncompressedCurvePoint` into the `From` impls below.

That's because I'm using `From` bounds for `FixedBaseScalarMul::Point`, but also I think having the `to_*_pubkey` methods return a `PublicKey` makes sense.

k256/src/arithmetic.rs
```diff
@@ -124,36 +133,66 @@ impl AffinePoint {
...
         encoded[0] = 0x04;
-        encoded[1..33].copy_from_slice(&self.x.to_bytes());
-        encoded[33..65].copy_from_slice(&self.y.to_bytes());
+        encoded[1..33].copy_from_slice(&affine_point.x.to_bytes());
+        encoded[33..65].copy_from_slice(&affine_point.y.to_bytes());
 
         UncompressedCurvePoint::from_bytes(GenericArray::clone_from_slice(&encoded[..]))
             .expect("we encoded it correctly")
     }
 }
```
tarcieri, line 185: It seems like if we had an `AffineCoordinates` trait which provided access to the equivalent of `x.to_bytes()` and `y.to_bytes()`, we could have generic `CompressedCurvePoint::from_affine_point` and  `UncompressedCurvePoint::from_affine_point` methods which could DRY out this logic, since it's otherwise identical between the two crates.

If we had that, the bounds used on `FixedBaseScalarMul::Point` could be dramatically simpler since we wouldn't need to bound on the `From` impls.

tarcieri:
Calling this effectively complete. I'm going to open another PR (which incorporates this one) to add an `AffineCoordinates` trait and see if I can use that to further DRY out some of the copy-and-pasted code between the `k256` and `p256` crates.

PTAL @str4d and @tuxxy !

elliptic-curve-crate/src/weierstrass.rs
tarcieri: By introducing an `AffineCoordinates` trait I was able to eliminate the `CompressedCurvePoint<Self>: From<Self::Point>` and `UncompressedCurvePoint<Self>: From<Self::Point>` bounds 🎉 

See #50

k256/src/arithmetic.rs
tarcieri: I implemented the above in #50 and it totally works and DRYs out all of this ugly serialization logic into the `elliptic-curve` crate itself so it doesn't have to be reimplemented on a curve-by-curve basis 🎉

k256/src/arithmetic.rs
str4d: It doesn't matter specifically for `CtOption`'s usage, but in all my implementations, `Default` is the identity, not the generator. This seems like it could quickly cause issues...

tarcieri: Sounds good, I can change it to the identity

tarcieri: Err, seems `identity()` is only defined for `ProjectivePoint`

tarcieri: I eliminated the `Default` bound and deleted the `impl Default for AffinePoint` blocks

elliptic-curve-crate/src/weierstrass.rs
tarcieri: Got rid of it 🎉

elliptic-curve-crate/src/weierstrass/public_key.rs
```diff
@@ -101,6 +102,35 @@ where
...
+            let affine_point = ct_option.unwrap();
+
+            if compress {
+                Some(PublicKey::Compressed(affine_point.into()))
+            } else {
+                Some(PublicKey::Uncompressed(affine_point.into()))
+            }
+        } else {
+            None
+        }
```
tarcieri, line 130: I was able to eliminate the `Default` bound on `AffinePoint` by having this return an `Option` rather than a `CtOption` as soon as scalar multiplication is complete (i.e. unwrapping the `CtOption` containing the public key).

Since we're dealing with a public key after scalar multiplication is complete, this seems ok to me.

tarcieri:
Going to go ahead and merge this as I think it's a pretty basic and important feature.

There are some potential alternative designs to explore/consider here, but I think this is a reasonable start.

PR #50, opened 2020-06-19
Add `AffineCoordinates` trait + DRY out point serialization

tarcieri:
*NOTE: this PR includes a commit from #49 which should get merged first*

Adds a trait for obtaining the x and y-coordinates of an affine point.

So as to avoid exposing field elements as part of the public API, the coordinates are first serialized as bytes.

Using this trait, it's possible to generically implement compressed and uncompressed point serialization according to the SEC-1 Elliptic-Curve-Point-to-Octet-String encoding.

This commit also modifies the trait bounds of `FixedBaseScalarMul` to bound on the `AffineCoordinates` trait, allowing the result to be an affine point for the curve which can be passed as input to the newly added `from_affine_point` methods on `CompressedCurvePoint` and `UncompressedCurvePoint`.


elliptic-curve-crate/src/coordinates.rs
```diff
@@ -0,0 +1,16 @@
...
+pub trait AffineCoordinates {
+    /// Size of a byte array representing an affine coordinate
+    type ScalarSize: ArrayLength<u8>;
+
+    /// x-coordinate
+    fn x(&self) -> ScalarBytes<Self::ScalarSize>;
+
+    /// y-coordinate
+    fn y(&self) -> ScalarBytes<Self::ScalarSize>;
+}
```
str4d, line 16: This abstraction directly exposes the coordinates instead of the effect of using them. Making it a separate trait is better than having it be part of the core traits, but it still implies that coordinates are an API surface that is supported and encouraged.

I would prefer instead e.g. an `EcdsaPrimitive` trait that requires implementing the inner part of the ECDSA algorithm that requires coordinates, and then `EcdsaCurve: EcdsaPrimitive` that provides the full algorithm, which users can import without exposing the inner API unnecessarily. This is similar to the `rsa::PrivateKey: rsa::DecryptionPrimitive` split.

Separately, using `ScalarBytes` for the coordinates strongly encourages type errors during implementation of the trait. The coordinates are field elements, and scalars (of the affine point's curve) are field elements (of a different field), but the coordinates are not scalars.

elliptic-curve-crate/src/weierstrass.rs
```diff
@@ -4,8 +4,28 @@ pub mod curve;
...
+
+/// Fixed-base scalar multiplication
+pub trait FixedBaseScalarMul: Curve
+where
+    <Self::ScalarSize as Add>::Output: Add<U1>,
+    CompressedPointSize<Self::ScalarSize>: ArrayLength<u8>,
+    UncompressedPointSize<Self::ScalarSize>: ArrayLength<u8>,
+{
+    /// Affine point type for this elliptic curve
+    type Point: AffineCoordinates + ConditionallySelectable + Default;
```
str4d, line 26: Fixed-base scalar multiplication doesn't require exposing affine coordinates, so this is a more restrictive abstraction than necessary. I suspect we can address the type bounds issue in another way.

tarcieri:
Okay, based on @str4d's comments this is too leaky a design, so let me propose an alternative.

> Fixed-base scalar multiplication doesn't require exposing affine coordinates, so this is a more restrictive abstraction than necessary.

What if we had something like:

```rust
pub struct Coordinate<C: Curve> {
    bytes: ScalarBytes<C::ScalarSize>
}
```

...but with a twist: while it's possible for library-level code to instantiate, say with this sort of API:

```rust
impl<C: Curve> From<ScalarBytes<C::ScalarSize>> {
...
}
```

...to everything but the `elliptic-curve` crate itself, the contents would be opaque. This would allow elliptic curve implementations to provide the `elliptic-curve` crate with serialized field elements, but nothing else would be directly able to consume them.

This could be used by the `elliptic-curve` crate to expose an `EcdsaPrimitive`, as well as the current usage in this PR: generic serializers for the SEC-1 `Elliptic-Curve-Point-to-Octet-String` encoding.

tarcieri:
Alternatively to that, we could keep the bounds from #49, but still add `from_affine_coordinates` methods to `CompressedCurvePoint` and `UncompressedCurvePoint` which take `GenericArray`-based `x` and `y` coordinates.

Concrete curve implementations could then invoke those methods directly, rather than exposing them as a public trait.

This keeps the coordinates encapsulated while still allowing for generic SEC-1 serialization, which seems like a nice middle ground.

I might open a separate PR based on #49 which implements that idea.

tarcieri:
Closing out this PR as this does not seem like the right abstraction.

I'll open another PR to DRY out SEC-1 encoding by incorporating it directly into the `elliptic-curve` crate, but in a way that doesn't leak affine coordinate field elements in the public API.

PR #52, opened 2020-06-20
Factor SEC-1 serialization into `elliptic-curve` crate

tarcieri:
Adds a `from_affine_coords` method to `CompressedPoint` and `UncompressedPoint` which accepts a `ScalarBytes` for the x and y-coordinates.

This is then used in the `k256` and `p256` crates to consolidate the duplicated SEC-1 serialization logic.


tarcieri:
This (hopefully) accomplishes one of the goals of #50 without a leaky abstraction /cc @str4d

Issue #54, opened 2020-06-22
Scalar arithmetic


tomtau:
It'd be good to (optionally) expose arithmetic operations on scalars.
@nickray has this for `p256` feature-guarded: https://github.com/nickray/elliptic-curves/commit/3e45bdece6db39a2b30ad247bfc395c740935c75
Not sure if it is OK to upstream this?

str4d:
I was planning to implement scalar arithmetic myself for `p256` later this week 😝 so yes, it would be good to have. It will be required in any case for implementing the `ff` and `group` crate traits.

nickray:
I'm happy to PR my scalar arithmetic if so desired. In any case, there are decisions to make:
- I implement scalars via Barrett reduction on `pub struct Scalar([u64; 4])`, is this what is desired?
- Scalar field inversion via Euler's theorem is unbearably slow on microcontrollers. My current implementation (not published) uses Stein's theorem instead, which is highly not-constant time and would necessitate a masking scalar in signature generation (calculate `r*(1/(k*r))` for random `r`).
- Would we want both implementations, and if so, configure via feature flag, or separate methods, or...?

My current signature implementation ("`hazmat`", entropy to be generated and hashes to be calculated externally, e.g. with hardware acceleration) looks like
```
impl Scalar {
    pub fn try_sign_prehashed(&self,    
        ephemeral_scalar: Scalar,    
        // for side channel protection
        masking_scalar: Option<Scalar>,    
        hashed_msg: &[u8; 32],        
    ) -> Result<[u8; 64], Error> { (...) }
}
```
It's not clear to me what the intended implementation approach for DH and signatures is in `p256`, but that would be a next step anyway.

tarcieri:
> My current signature implementation ("hazmat", entropy to be generated and hashes to be calculated externally, e.g. with hardware acceleration) looks like...

Neat! How did you implement `masking_scalar: Option<Scalar>`?

I'm thinking it probably makes the most sense to implement [RFC 6979](https://tools.ietf.org/html/rfc6979), but providing that sort of `masking_scalar`, using techniques like the ones described in [draft-mattsson-cfrg-det-sigs-with-noise Section 3](https://tools.ietf.org/html/draft-mattsson-cfrg-det-sigs-with-noise-00#section-3).

[...]

PR #56, opened 2020-06-22
Implement scalar arithmetic via Barrett reduction

nickray:
To make https://github.com/RustCrypto/elliptic-curves/issues/54 more concrete.

### DIFF
--- a/p256/src/arithmetic/scalar.rs
+++ b/p256/src/arithmetic/scalar.rs
@@ -1,34 +1,115 @@
-//! Scalar field arithmetic.
+//! Scalar field arithmetic modulo n = 115792089210356248762697446949407573529996955224135760342422259061068512044369
 
 use core::convert::TryInto;
-use elliptic_curve::subtle::{Choice, ConditionallySelectable, CtOption};
+use core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+use elliptic_curve::subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
 
-use crate::{arithmetic::util::sbb, SecretKey};
+use super::util::{adc, mac, sbb};
+use crate::SecretKey;
 
 /// The number of 64-bit limbs used to represent a [`Scalar`].
 const LIMBS: usize = 4;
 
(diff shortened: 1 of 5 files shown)


p256/src/arithmetic/scalar.rs
```diff
@@ -76,6 +157,24 @@ impl Scalar {
         CtOption::new(Scalar(w), Choice::from(is_some))
     }
 
+    /// Parses the given byte array as hashed message.
+    ///
+    /// Subtracts the modulus when the byte array is larger than the modulus.
+    pub fn from_hash(bytes: [u8; 32]) -> Self {
```
tarcieri, line 163: I would expect a `from_hash` method to accept a `Digest` instance. How about naming it `from_bytes_mod_order` instead?

```suggestion
    pub fn from_bytes_mod_order(bytes: [u8; 32]) -> Self {
```

nickray: I'm not sure what the right name is. Thinking generically, this is tied to "Let *z* be the  leftmost bits of , where  is the bit length of the group order *n*.", so I'd like a method and name that captures this requirement. Ideas?

tarcieri: I think `from_hash` is fine if the argument is a `Digest` instance. There's a bit of a rationale for why e.g. `signature::DigestSigner` takes one (as opposed to a byte array) here:

https://docs.rs/signature/1.1.0/signature/trait.DigestSigner.html#notes

tarcieri: Okay, so my original suggestion was definitely off... I'm now thinking (as I working on #68) this is actually the opposite of what I originally suggested, and something more like: `from_bytes_unreduced`

tarcieri: @nickray since this method is specific to `p256` (and after #73 we no longer need to worry about coming up with cross-curve abstractions) how about just calling this `from_bytes_reduced`?

nickray: Sure, I'll follow along if/when https://github.com/RustCrypto/elliptic-curves/pull/76 is merged :+1:

tarcieri: I'm actually liking #78 better at this point

tarcieri:
Merged as 061a1a2 in #83

PR #58, opened 2020-06-23
Re-export types from `arithmetic` module at toplevel

tarcieri:
This commit removes the `arithmetic` module from the public API of the `k256` and `p256` crates, re-exporting the `AffinePoint`, `ProjectivePoint`, and `Scalar` types from the crate's toplevel.

Additionally, it moves the `test_vectors` module to the toplevel, gated under the `test-vectors` feature (or `cfg(test))`).

To make it clear these types are available conditionally based on the `arithmetic` and `test-vectors` features, `doc_cfg` is used to tag them in rustdoc as rendered by https://docs.rs


tarcieri:
Rendered rustdoc:

## `k256`

Screen Shot 2020-06-23 at 9 19 48 AM

## `p256`

Screen Shot 2020-06-23 at 9 15 30 AM

nickray:
What is the reasoning for arithmetic being a feature (crate is rather barren without, and the feature doesn't trigger additional dependencies AIUI)? My instinct would be to organize the library around modules field, point/curve, scalar, signature, agree, with randomness the core optional feature.

That said, I do prefer your suggestion over status quo.

A remark that perhaps doesn't quite fit here but seems relevant: For my microcontroller stuff, FieldElement newtypes [u32; 8] instead of [u64; 4], which impacts all the other types (and I deal with it via features that switch the entire type, not traits/generics).

tarcieri:
> What is the reasoning for arithmetic being a feature (crate is rather barren without, and the feature doesn't trigger additional dependencies AIUI)? 

These crates are used in two different ways at the moment:

1. With the `arithmetic` feature on (when available) it can function in a self-contained manner
2. With it off, the `PublicKey`, `CompressedPoint`/`UncompressedPoint`, and `weierstrass::Curve` types can still be used  in conjunction with the `ecdsa` crate to provide common key and signature types (generic around the underlying `weierstrass::Curve`) in cases where it's wrapping existing ECDSA implementations in a common trait-based API.

This includes things like:

- Wrappers for Rust libraries which provide their own arithmetic implementation e.g. `secp256k1`  the C FFI wrapper for libsecp256k1, *ring* for NIST curves
- Hardware devices as in the `yubihsm` and `yubikey-piv` crates
- Cloud KMS services

With the types defined in the curve crates, it's possible to use trait objects as a common interface for representing various ECDSA providers. For example, in the Tendermint KMS project I work on:

https://github.com/iqlusioninc/tmkms/blob/develop/src/keyring/ecdsa.rs#L24

...we are able to support a libsecp256k1-based signer, Ledgers, and YubiHSM2s through a common keyring.

Arguably we could remove the `arithmetic` feature, and in these cases the curve arithmetic would be compiled even when it isn't needed, which isn't the end of the world. However, I imagine as more (e.g. table-based, possibly using tables computed as part of a build step) complex curve arithmetic is added, it will be increasingly nice to be able to skip all of that in the event that all you care about are the curve and "wire" types for interoperability between crates.

PR #59, opened 2020-06-29
New arithmetic for k256

fjarri:
This PR implements "lazy reduction" arithmetic for SECP256k1, based on [libsecp256k1](https://github.com/bitcoin-core/secp256k1).

(Disclaimer: this is my first non-trivial piece of Rust code. Don't hesitate to point out my mistakes).

- adds 5x52bit and 10x26bit arithmetic for the internal field (the old Montgomery one is preserved)
- adds 4x64bit and 8x32bit arithmetic for scalars
- adds more tests for fields and scalars (using `num-bigint` as a reference)
- adds a fixed window point-scalar multiplication

More advanced multiplication routines (wNAF, endomorphism, batch multiplication) will not be included in this PR — it has too much stuff already. Even the windowed multiplication already gives a good performance boost. As compared to the baseline performance, I currently see an almost 3x speedup (300us vs 106us per mul). 

The old Montgomery field can be enabled with `field-montgomery`, 32-bit fields & scalars on a 64-bit target can be enabled with `force-32-bit`.


tarcieri:
> Note: there is a #[bench] test in field.rs, so the whole thing only compiles for nightly now. Not sure what is the best practice here.

A couple options:

1) Add `#[cfg(nightly)]` gating, and run the benchmarks as `cargo +nightly bench -- --cfg nightly`. Somewhat annoying.
2) On some other issues (#57) we've discussed adding an `expose-arithmetic` cargo feature (potentially hidden in the https://docs.rs docs, and with appropriate scare language documenting it) which would be useful in this case. It would allow you to use nicer tools for benchmarking too (e.g. criterion)

tarcieri:
Looks like some recent changes to `arithmetic.rs` are now conflicting. Can you rebase please?

Overall this looks exciting!

fjarri:
Sure, I'm doing some abstracting now, will rebase after.

fjarri:
Sorry for a bit of a delay on this. I was playing with endomorphism-based multiplication from the original libsecp256k1, but the code related to it is quite large in size and very convoluted, so it is better left for the next PR. For now I'll settle for a simple fixed-window multiplication (which is still an almost 2x improvement over the previous double-and-add) and limit this PR to scalar arithmetic.

fjarri:
@tarcieri I added the `nightly-bench` feature as you suggested, and it seems to work (tests run on my machine both on stable and nightly), but CI still fails because this feature gets enabled by `--all-features`. Is there a way to avoid that somehow? This will also be a problem when I add 32-bit implementations for FieldElement and Scalar, which I planned to make available as features.

tarcieri:
My suggestion was to add a `cfg` attribute, not a cargo feature:

```rust
#[cfg(nightly)]
```

fjarri:
For some reason it doesn't work for me. The marked functions/imports are not picked neither `stable` nor in a `nightly` run. `--cfg` option is not recognized either:

```
> cargo +nightly bench -- --cfg nightly
    Finished bench [optimized] target(s) in 0.03s
     Running /Users/bogdan/wb/github/elliptic-curves/target/release/deps/k256-5d871111fa650de9
error: Unrecognized option: 'cfg'
error: bench failed
```

tarcieri:
Hmm, that's unfortunate.

I think a better option in general would be finding a way to make it work with criterion, which in addition to working on stable provides a much better benchmarking tool in general.

fjarri:
I see. I guess I'll just scrap the benchmark in the final version then.

tarcieri:
Sounds good for now. We can explore adding an `expose-arithmetic` feature in a follow-up PR which would permit a benchmark using criterion.

tarcieri:
Not sure why the codecov check is failing, but regardless, don't worry about it

tarcieri:
@fjarri sorry about #66. After landing it, I realized I'll need a bunch of stuff in this branch anyway for what I'm presently working on (public key recovery from ECDSA) and should probably be using this branch as a base instead

fjarri:
No worries, seems like a small change anyway. I know I've been taking some time with this PR, but I'm really hoping to get it ready for review this weekend.

tarcieri:
@fjarri fantastic! I can keep working off your WIP branch in the meantime.

Also: do you plan on adding scalar inversions? It looks like you have most of the prerequisites in place.

fjarri:
@tarcieri , uhh, I just realized you merged some of my commits to upstream. I was intending to squash them, they're quite unkempt, but I guess it's too late now...

k256/Cargo.toml
```diff
@@ -23,13 +23,22 @@ optional = true
...
 default = ["arithmetic", "std"]
+all-nonconflicting = ["default", "rand", "test-vectors", "64bit"]
 arithmetic = []
+field-10x26 = []
+field-5x52 = []
+field-montgomery = []
+scalar-8x32 = []
+scalar-4x64 = []
+64bit = ["field-5x52", "scalar-4x64"]
+32bit = ["field-10x26", "scalar-8x32"]
```
tarcieri, line 41: Would it make sense to select the scalar/field backends automatically based on `cfg(target_pointer_width = ...)` instead of using Cargo features? As you're encountering here, they're really intended for additive features and not ones that are mutually exclusive.

Here's an example of the `polyval` crate doing this for its GF(2^128) field backends:

https://github.com/RustCrypto/universal-hashes/blob/master/polyval/src/field.rs#L46-L50

(still kind of gross, but at least automatic based on the target)

fjarri:
Also, I added an addition-table based inversion for scalars, which should be faster than a generic vartime power.

tarcieri:
> I just realized you merged some of my commits to upstream. I was intending to squash them, they're quite unkempt, but I guess it's too late now...

@fjarri  I used them in a draft PR (#68). None of them have been merged to the `master` branch.

fjarri:
Ah, my bad, I'm going to do some squashing then.

k256/src/arithmetic/field.rs
```diff
@@ -1,583 +1,289 @@
...
-    type Output = FieldElement;
-
-    fn neg(self) -> FieldElement {
-        FieldElement::zero() - self
+#[cfg(test)]
+impl ToBigUint for FieldElement {
+    fn to_biguint(&self) -> Option<BigUint> {
+        self.normalize().0.to_biguint()
     }
 }
```
tarcieri, line 264: I believe you can move these `#[cfg(test)]`-gated impls into the `tests` module so everything test-related is in one module

k256/Cargo.toml
fjarri: Ok, I'll try it out. There's still a problem of how to fit the old (Montgomery) field arithmetic. Without it, we could just make 64-bit backends default, and 32-bit ones features that would turn off the 64-bit ones (that's the way I had it initially).

k256/src/arithmetic/field.rs
fjarri: Would it be useful to expose conversion to/from BigUint in the main library?

k256/src/arithmetic/field/field_impl.rs
```diff
@@ -0,0 +1,209 @@
...
+
+#[cfg(debug_assertions)]
+impl Default for FieldElementImpl {
+    fn default() -> Self {
+        Self::zero()
+    }
+}
+
+#[cfg(debug_assertions)]
+#[cfg(test)]
```
tarcieri, line 195: `debug_assertions` defaults to on for the `test` profile, so this feels a little redundant:

https://doc.rust-lang.org/cargo/reference/profiles.html#test

It seems like you could either just use `#[cfg(test)]` or something like the following:

```suggestion
#[cfg(all(debug_assertions, test))]
```

k256/src/arithmetic/scalar/scalar_4x64.rs
```diff
@@ -0,0 +1,459 @@
...
+            }
+        }
+
+        Self(res)
+    }
+
+    #[cfg(feature = "zeroize")]
+    pub fn zeroize(&mut self) {
+        self.0.as_mut().zeroize()
+    }
```
tarcieri, line 289: Any reason why you're defining `fn zeroize` as an inherent method rather than as `impl Zeroize`? (I see it in a few other places as well, e.g. the 32-bit field and the scalar backends.

k256/src/arithmetic/field.rs
tarcieri: I don't see a reason to. You're just using it for equivalence testing, right?

k256/src/arithmetic/scalar/scalar_4x64.rs
fjarri: That's one thing I need to clean up. I was initially confused by all the `add()`/`mul()` etc methods, but it seems that those only exist as inherent methods to simplify defining traits for different combinations of reference/not reference.

k256/Cargo.toml
tarcieri: The `polyval` example above is trying to juggle three field arithmetic backends as well (32-bit, 64-bit, and one based on Intel AVX2). To make that a bit more manageable it's using the  [`cfg-if`](https://docs.rs/cfg-if/) crate to manage some of the gating.

It seems like you'll need at least one Cargo feature in the mix to keep the Montgomery field arithmetic around, but unless there's a good reason to do otherwise it seems like you can autodetect the 32/64-bit distinction based on the target.

k256/src/arithmetic/field.rs
fjarri: Got it. Yes, just as a reference in tests instead of fiat-crypto (which is Montgomery-based).

tarcieri:
Another small change you might consider making (or otherwise I can) is wiring up the `GenerateSecretKey` implementation to use your newly added `Scalar::generate` method:

https://github.com/RustCrypto/elliptic-curves/blob/42dcac1/k256/src/arithmetic.rs#L519-L535

k256/Cargo.toml
fjarri: Hm, in this case how would I go around running tests for 32 bit target on my machine? I'm on Macos, and I don't see any mac-32 bit target (Catalina doesn't even support 32-bit anymore afaik). Would it make sense to add a "force 32 bit" feature or something?

fjarri: And how would the CI config look like?

fjarri:
`Scalar::generate()` has a small non-uniformity, the current method is not constant time - which is preferable?

k256/Cargo.toml
fjarri: Ok, made another attempt, with a `force-32-bit` flag to test 32-bit backends.

tarcieri: The absence of a 32-bit target for macOS is a bit annoying (I sympathize as a fellow macOS user). Otherwise target-based gating could work for CI (and checking on Linux for local development).

Another option is to add `cfg(test)` to the gating for including the backends, and include both under a `test` configuration.

Since this has gotten a lot trickier than I initially thought it was going to be, I'm happy to table it for now with your `force-32-bit` feature and circle back on it myself and see if I can wire up things using `cfg-if` to test both the 32-bit and 64-bit backends, but select one of them to use automatically in builds.

tarcieri:
If `Scalar::generate()` has a negligible bias from the uniform distribution, I'd say I weakly prefer a constant-time implementation.

So I guess the question is: is the bias significant enough to actually worry about?

fjarri:
That I'm not qualified to answer. @tuxxy , what do you say? The biased generation is essentially the same as it was for field elements - 512 random bits are generated and modulo reduced.

k256/src/arithmetic/scalar/scalar_4x64.rs
fjarri: Fixed.

k256/src/arithmetic/field/field_impl.rs
fjarri: Somewhat fixed, I guess. Much less `#[cfg(test)]` around the code now.

k256/src/arithmetic/field.rs
fjarri: Fixed.

k256/Cargo.toml
fjarri: Ok, let's call it fixed for now.

k256/src/arithmetic/field/field_impl.rs
fjarri: Also, I think it is worth it to keep `debug_assertions` and `test` separate. `test` marks whatever is used in tests, `debug_assertions` marks debug assertions. These are orthogonal flags; just because by default `debug_assertions` is on when `tests` is on, does not mean that they can't be active separately under some conditions.

k256/src/arithmetic.rs
```diff
@@ -24,12 +25,41 @@ use elliptic_curve::{
...
 
-/// b = 7 in Montgomery form (aR mod p, where R = 2**256.
-const CURVE_EQUATION_B: FieldElement = FieldElement([
-    0x0000_0007_0000_1ab7,
-    0x0000_0000_0000_0000,
-    0x0000_0000_0000_0000,
-    0x0000_0000_0000_0000,
+const CURVE_EQUATION_B_SINGLE: u32 = 7u32;
+
+const CURVE_EQUATION_B: FieldElement = FieldElement::from_bytes_unchecked(&[
```
tarcieri, line 30: Looks like `rustfmt` is... not being very helpful here. You might want to:

```suggestion
#[rustfmt::skip]
const CURVE_EQUATION_B: FieldElement = FieldElement::from_bytes_unchecked(&[
```

...and format it slightly better than what `rustfmt` is doing.

k256/src/arithmetic/field/field_10x26.rs
```diff
@@ -0,0 +1,681 @@
+use elliptic_curve::subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
+
+#[cfg(feature = "getrandom")]
+use getrandom::getrandom;
```
tarcieri, line 4: There isn't currently a `getrandom` feature... see #47 (and #62 where I got rid of a few usages that were missed)

Using `CryptoRng + RngCore` is a bit friendlier to embedded devices as it works with non-global RNGs. `rand_core` also has a `getrandom` feature you can use to access the OS RNG via `rand_core::OsRng`

tarcieri:
@fjarri if you rebase, I believe #70 will address the codecov action failing

k256/src/arithmetic.rs
fjarri: Thanks, fixed

k256/src/arithmetic/field/field_10x26.rs
fjarri: Fixed

tarcieri:
@fjarri I'd say this is ready to merge unless there are any objections.

Normally we use GitHub's "Squash and Merge" however if you'd like me to preserve the commit history here (or squash it yourself) you can.

fjarri:
Squash and merge sounds good to me.

Please give me until the end of the day before merging, I just want to add a bit more comments.

tarcieri:
@fjarri regarding the `SecretKey` issue, after talking with some people it sounds like the current method it's using (rejection sampling) is the safest.

Notably it can also be made constant time by always generating a fixed number of samples. The probability of a non-biased RNG generating multiple consecutive candidates which are out of range vanishes rather quickly (after ~4 candidates).

fjarri:
Ok, I think it's ready now. I exposed both `Scalar::generate_vartime()` (the algorithm from `generate_secret_key()`) and `Scalar::generate_biased()`, and used the former in `generate_secret_key()`. We can remove the biased version later if no-one needs it, but at least its drawback is right there in the name.

tarcieri:
Merged! 🎉

Thanks so much!

tuxxy:
Hey, y'all! Sorry I didn't see these mentions (GitHub notifications don't exactly pull my attention anymore :sweat_smile:), but it sounds like y'all figured it out anyway.

This PR looks great @fjarri. Excited to build on this for sure!

daira:
> That I'm not qualified to answer. @tuxxy , what do you say? The biased generation is essentially the same as it was for field elements - 512 random bits are generated and modulo reduced.

In my professional opinion that is the best way to do it. It's what we use in Zcash for RedDSA signatures. There is no reason to worry about a ~2<sup>-256<sup> bias in any plausible cryptographic context.

tarcieri:
Interesting article on reduction sampling vs modular reduction for randomly generating in-range values with a uniform distribution. The conclusion was interesting: why not both?

https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/

PR #60, opened 2020-06-29
k256: add `Scalar::is_high`

tarcieri:
Adds a method for determining if a scalar is greater than or equal to the modulus divided by 2 (in constant time).

The ultimate purpose of this method is implementing non-malleable ECDSA signatures as described in [BIP 0062: Dealing with Malleability](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki), namely ensuring signatures are [low-S normalized](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures).

In the event the S-component of an ECDSA signature is "high", it should be normalized (to be low) by subtracting it from the modulus.

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -1,7 +1,7 @@
 //! Scalar field arithmetic.
 
 use core::convert::TryInto;
-use elliptic_curve::subtle::{Choice, ConditionallySelectable, CtOption};
+use elliptic_curve::subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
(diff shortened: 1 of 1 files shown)


k256/src/arithmetic/scalar.rs
```diff
@@ -77,6 +85,15 @@ impl Scalar {
         ret[24..32].copy_from_slice(&self.0[0].to_be_bytes());
         ret
     }
+
+    /// Is this scalar greater than or equal to n / 2?
+    pub fn is_high(&self) -> Choice {
```
tarcieri, line 90: Perhaps this could use a comment similar to the one in `from_bytes` regarding how an overflow results in a borrow?

k256/src/arithmetic/scalar.rs
```diff
@@ -20,6 +20,14 @@ const MODULUS: [u64; LIMBS] = [
     0xFFFF_FFFF_FFFF_FFFF,
 ];
 
+/// Constant representing the modulus / 2
+const FRAC_MODULUS_2: [u64; LIMBS] = [
```
tarcieri, line 24: I borrowed the `FRAC_*_N` naming from [core::f64::consts](https://doc.rust-lang.org/core/f64/consts/index.html). Not sure if there's a better convention I should be using?

PR #61, opened 2020-06-30
k256: impl Neg for Scalar

tarcieri:
Implement negation (modulo n) for `Scalar`, for the purposes of implementing low-S normalization of ECDSA signatures (see #60).

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -1,6 +1,6 @@
 //! Scalar field arithmetic.
 
-use core::convert::TryInto;
+use core::{convert::TryInto, ops::Neg};
 use elliptic_curve::subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 
 #[cfg(feature = "zeroize")]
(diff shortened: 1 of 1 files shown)


k256/src/arithmetic/scalar.rs
```diff
@@ -90,9 +90,56 @@ impl ConditionallySelectable for Scalar {
...
+    type Output = Scalar;
+
+    fn neg(self) -> Scalar {
+        let (w0, borrow) = sbb(MODULUS[0], self.0[0], 0);
+        let (w1, borrow) = sbb(MODULUS[1], self.0[1], borrow);
+        let (w2, borrow) = sbb(MODULUS[2], self.0[2], borrow);
+        let (w3, _) = sbb(MODULUS[3], self.0[3], borrow);
+
+        let is_zero = self.0.ct_eq(&[0u64; LIMBS]);
+        Scalar::conditional_select(&Scalar([w0, w1, w2, w3]), &Scalar::zero(), is_zero)
```
tarcieri, line 103: Curious if there's a better solution for this...

PR #65, opened 2020-07-07
Add `legacy` feature with `AffineCoordinates` trait

tarcieri:
Protocols like ECDSA need access to affine coordinates: the x-coordinate for ECDSA itself, but also the y-coordinate for things like recovering public keys from ECDSA signatures (as seen in Ethereum).

This commit adds support for accessing affine coordinates for use in these protocols, taking into account some feedback from the previous attempt to do this (#50):

- Trait is placed inside a `legacy` module with comments strongly suggesting it should only be used for implementing legacy protocols which explicitly need coordinate access
- Feature is hidden from https://docs.rs to discourage (mis)use
- Coordinates are represented using a `FieldElementBytes` type alias which is also defined in the `legacy` module

/cc @nickray

### DIFF
--- a/elliptic-curve-crate/Cargo.toml
+++ b/elliptic-curve-crate/Cargo.toml
@@ -37,10 +37,10 @@ default-features = false
 hex = "0.4"
 
 [features]
-default = []
+legacy = []
 weierstrass = []
 std = []
 
 [package.metadata.docs.rs]
-all-features = true
+features = ["weierstrass", "std"]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/elliptic-curve-crate/src/legacy.rs
+++ b/elliptic-curve-crate/src/legacy.rs
@@ -1,0 +1,26 @@
+//! Support for implementing legacy protocols that require direct access to
+//! coordinates of affine points on elliptic curves.
+//!
+//! These APIs violate the group abstraction and expose coordinate field
+//! elements which could be potentially misused when designing new protocols
+//! based on elliptic curve groups.
+//!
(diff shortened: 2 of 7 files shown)


tarcieri:
I've been poking at using this in some local branches and thought I'd spell out the immediate concrete use cases for each of the coordinates.

## ECDSA signing

Needs access to the x-coordinate.

@nickray's quick 'n' dirty ECDSA signing implementation (#57) serializes the field element to an integer-as-bytestring  and then reduces it to an element of the `Scalar` field for use in subsequent operations:

https://github.com/RustCrypto/elliptic-curves/pull/57/files#diff-e6b3b4deef9b63467f2493d3bb13a7b2R58-R65

```rust
        // Geometrically, x-coordinate is an element of the base field.
        // to_bytes lifts this to a big-endian integer.
        // This integer is then reduced to an element of the scalar field (n < p).
        let x_coordinate: [u8; 32] = ephemeral_point.x.to_bytes();
        let r = Scalar::from_hash(x_coordinate);
        if bool::from(r.is_zero()) {
            return Err(Error::RIsZero);
        }
```

## ECDSA public key recovery

Needs access to the y-coordinate to compute the "recovery ID", but like point compression only really needs to know if the y-coordinate is odd.

tarcieri:
I prefer https://github.com/RustCrypto/signatures/pull/96 to this PR for the purposes of ECDSA.

tarcieri:
The `elliptic-curve` crate has been moved to the https://github.com/RustCrypto/traits repo.

If anyone is still interested in this feature, this PR needs to be reopened there.

PR #67, opened 2020-07-11
[WIP] Add `legacy` feature with ECDSA-related traits

tarcieri:
This is an alternative to #65 which scopes the "legacy" functionality specifically to what is needed to implement ECDSA.

ECDSA signing computes the `r` component of a signature by reducing the affine x-coordinate of an ephemeral point mod n.

This adds a `ReduceAffineX` trait for performing this reduction, with the goal of facilitating an ECDSA implementation which is generic over (prime order) elliptic curves.

Additionally, it adds an `IsAffineYOdd` trait which is useful for computing the "recovery ID" used to recover the public key which created a signature (ala point compression).

### DIFF
--- a/elliptic-curve-crate/Cargo.toml
+++ b/elliptic-curve-crate/Cargo.toml
@@ -37,10 +37,10 @@ default-features = false
 hex = "0.4"
 
 [features]
-default = []
+legacy = []
 weierstrass = []
 std = []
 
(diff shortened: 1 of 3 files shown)


tarcieri:
I prefer https://github.com/RustCrypto/signatures/pull/96 to this

Issue #55, opened 2020-06-22
Microcontroller optimizations.


nickray:
I have a rather fast implementation of the base field for Cortex-M4/M33 microcontrollers, ~wrapping~ stealing the assembly routines in https://github.com/Emill/P256-cortex-ecdh/blob/master/P256-cortex-m4-ecdh-speedopt-gcc.s. Would there be interest to include platform-specific arithmetic implementations in this crate, or should I focus on a "lean and mean" MCU fork?

The complications are:
- use `[u32; 8]` for base field
- assemble routines and commit "binary" to repository (avoids need for assembler in dependent crates)
- pick implementation in `build.rs`, possibly overridable
- I'm experimenting with inline assembly as well, which might be pulled in if compiled on nightly

tarcieri:
> Would there be interest to include platform-specific arithmetic implementations in this crate, or should I focus on a "lean and mean" MCU fork?

Thus far we've generally focused on pure Rust implementations (cc @newpavlov) however some precedent for ASM can be found in https://github.com/RustCrypto/asm-hashes, which we then conditionally pull into the otherwise pure Rust hash implementation gated under an `asm` feature:

https://github.com/RustCrypto/hashes/blob/master/sha2/Cargo.toml#L37

To get the ball rolling, I'd suggest adding the ASM directly to the `p256` crate, and then we can discuss how to split it up.

I'm curious how @newpavlov feels about including some inline assembly directly into the same crate if it's gated under an `asm` feature (in addition to target-specific gating), now that it seems inline ASM is headed towards stabilization. IMO that would be fine, especially if it requires explicit opt-in via a feature.

newpavlov:
If it will be implemented using the new `asm!` macro, then I also think it should be fine to keep it as part of the `p256` crate, also it would help to test the new macro. But requiring nightly compiler is pretty strong requirement, especially nowadays, so I guess we could start with linking the `s` file using `cc`. Personally I would prefer the `asm!` option, but since I don't plan to use it in the near future, I will leave decision to you.

P.S.: I also would like to keep licensing simple, so it would be nice if the assembly code will be re-licensed under MIT/Apache as part of the PR.

PR #69, opened 2020-07-12
[WIP] Add ECDSA primitive traits to the `elliptic-curve` crate

tarcieri:
(Another alternative to #65 and #67 with the goal of generic cross-curve ECDSA support)

Adds a set of low-level "hazmat" traits to the `elliptic-curve` crate for implementing ECDSA signing, verification, and public key recovery.

These primitives allow crates containing implementations of specific curves to define core ECDSA functionality in a coarse-grained manner which can potentially leverage things like assembly or hardware cryptographic accelerators.

### DIFF
--- a/elliptic-curve-crate/Cargo.toml
+++ b/elliptic-curve-crate/Cargo.toml
@@ -38,9 +38,10 @@ hex = "0.4"
 
 [features]
 default = []
+ecdsa = ["weierstrass"]
 weierstrass = []
 std = []
 
 [package.metadata.docs.rs]
-all-features = true
+features = ["weierstrass", "std"]
 rustdoc-args = ["--cfg", "docsrs"]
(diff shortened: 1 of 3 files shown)


tarcieri:
I put these traits in this crate because presently the `ecdsa` crate consumes the various curve implementations, however as I look at them it feels a bit gross and like this is the wrong place for them and they should go in the `ecdsa` crate.

Another option is to put these traits in the `ecdsa` crate and reverse the relationship between the `ecdsa` crate and the individual curve implementations, so the curves pull in `ecdsa` as an optional feature.

I think that approach would actually make the most sense: it means anyone could potentially implement ECDSA for curves in third party crates.

All the rest of that said, I really like this approach over #65 and #67. I think it addresses @nickray's concerns in #55 about assembly-based optimizations by having the trait express a coarse-grained ECDSA implementation.

tarcieri:
Here's similar (actually, much simpler!) traits implemented in the `ecdsa` crate:

https://github.com/RustCrypto/signatures/pull/96

I definitely prefer that approach over this one, so I'm going to go ahead and close this out.

Issue #71, opened 2020-07-12
Move `elliptic-curve` crate to RustCrypto/traits?


tarcieri:
It's been pretty convenient having the `elliptic-curve` crate in the same repo as the curve implementations so far.

However, all of the other trait crates are presently in https://github.com/RustCrypto/traits

Aside from the consistency issue, there's a notable practical one: if we land https://github.com/RustCrypto/signatures/pull/96 to reverse the relationship between the `ecdsa` crate and the curve implementation crates, there's now a circular dependency between crates in https://github.com/RustCrypto/elliptic-curves and the `ecdsa` crate in https://github.com/RustCrypto/signatures

- `ecdsa` depends on `elliptic-curve`
- https://github.com/RustCrypto/signatures/pull/96 would have the `k256`, `p256`, `p384` depend on both the `elliptic-curve` and `ecdsa` crates

This would make it difficult to use `[patch.crates-io]` directives to make incremental progress on unreleased versions of crates.

The most straightforward solution I can think of is to eliminate this circular dependency by moving `elliptic-curve` into the traits repo. That would allow both `ecdsa` and the curve implementations to use a `[patch.crates-io]` directive resolving to `elliptic-curve` in https://github.com/RustCrypto/traits

tarcieri:
Went ahead and made this move. I think it works out for the best.

PR #72, opened 2020-07-13
elliptic-curve-crate: move to RustCrypto/traits

tarcieri:
This commit completes a move of the `elliptic-curve` crate to:

https://github.com/RustCrypto/traits/tree/master/elliptic-curve

See #71


tarcieri:
Note: I plan on landing #59 first before moving forward with this

PR #76, opened 2020-07-16
k256: add Scalar::from_bytes_reduced

tarcieri:
This method is intended for use when implementing ECDSA, namely for computing a `Scalar` from the hash of the message:

https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm#Signature_generation_algorithm

> Let 𝐳 be the 𝐋𝑛 leftmost bits of 𝑒, where 𝐋𝑛 is the bit length of the group order 𝑛

See related discussion here: https://github.com/RustCrypto/elliptic-curves/pull/56/files#diff-c8893dd2f9cf0dff9549c297fc437664R160-R176

cc @fjarri @nickray

### DIFF
--- a/k256/src/arithmetic/scalar/scalar_8x32.rs
+++ b/k256/src/arithmetic/scalar/scalar_8x32.rs
@@ -207,6 +207,26 @@ impl Scalar8x32 {
         CtOption::new(Self(w), underflow)
     }
 
+    /// Parses the given byte array as a scalar.
+    ///
+    /// Subtracts the modulus when the byte array is larger than the modulus.
+    pub fn from_bytes_reduced(bytes: &[u8; 32]) -> Self {
+        // Interpret the bytes as a big-endian integer w.
+        let w7 = u32::from_be_bytes(bytes[0..4].try_into().unwrap());
+        let w6 = u32::from_be_bytes(bytes[4..8].try_into().unwrap());
+        let w5 = u32::from_be_bytes(bytes[8..12].try_into().unwrap());
+        let w4 = u32::from_be_bytes(bytes[12..16].try_into().unwrap());
+        let w3 = u32::from_be_bytes(bytes[16..20].try_into().unwrap());
+        let w2 = u32::from_be_bytes(bytes[20..24].try_into().unwrap());
+        let w1 = u32::from_be_bytes(bytes[24..28].try_into().unwrap());
+        let w0 = u32::from_be_bytes(bytes[28..32].try_into().unwrap());
+        let w = [w0, w1, w2, w3, w4, w5, w6, w7];
+
+        // If w is in the range [0, n) then w - n will underflow
+        let (r2, underflow) = sbb_array_with_underflow(&w, &MODULUS);
+        Self(conditional_select(&w, &r2, !underflow))
+    }
+
     /// Returns the SEC-1 encoding of this scalar.
     pub fn to_bytes(&self) -> [u8; 32] {
         let mut ret = [0; 32];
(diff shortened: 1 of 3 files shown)


k256/src/arithmetic/scalar/scalar_8x32.rs
```diff
@@ -207,6 +207,26 @@ impl Scalar8x32 {
...
+        // Interpret the bytes as a big-endian integer w.
+        let w7 = u32::from_be_bytes(bytes[0..4].try_into().unwrap());
+        let w6 = u32::from_be_bytes(bytes[4..8].try_into().unwrap());
+        let w5 = u32::from_be_bytes(bytes[8..12].try_into().unwrap());
+        let w4 = u32::from_be_bytes(bytes[12..16].try_into().unwrap());
+        let w3 = u32::from_be_bytes(bytes[16..20].try_into().unwrap());
+        let w2 = u32::from_be_bytes(bytes[20..24].try_into().unwrap());
+        let w1 = u32::from_be_bytes(bytes[24..28].try_into().unwrap());
+        let w0 = u32::from_be_bytes(bytes[28..32].try_into().unwrap());
+        let w = [w0, w1, w2, w3, w4, w5, w6, w7];
```
nickray, line 223: A bit off-topic, but what is the motivation for this level of explicitness vs. loops or iterator approaches in cryptographic code? When I try this kind of thing, I tend to always get at least one index wrong :)

tarcieri: This is actually copypasta from both the `from_bytes` method as well as the `from_overflow` method.

I'd agree it could be simplified and potentially composed in terms of `from_overflow` too.

tarcieri: Perhaps some time after https://github.com/rust-lang/rust/pull/74060 is merged this could be:

```rust
let w: [u32; 8] = bytes.chunks(4).map(|chunk| u32::from_be_bytes(chunk).try_into().unwrap()).collect();
```

tarcieri:
I'm going to go with #78 instead

k256/src/arithmetic/scalar/scalar_8x32.rs
nickray: AIUI, in some cases the p256 arithmetic uses "many variables" to allow functions to be const, not in this case though. I was wondering if the use for this was motivated by consistency and subjective preference, or some kind of cryptographic reason, would you know?

tarcieri: I don't think there's any specific reason for it, but perhaps @str4d or @fjarri could chime in

fjarri: There is a minor performance benefit as compared to loops, but that can be negated by using an unrolling macro (there are several packages available for that). I have plans to do that, but just didn't get to it yet.

nickray: Ah, loop unrolling. Thanks!

tarcieri: As far as macros for loop unrolling go, I've seen [`unroll`](https://docs.rs/unroll/), but that said so far in my usage of [godbolt](https://godbolt.org/) to optimize cryptographic code I still haven't seen a case where unrolling a loop by hand generates more efficient assembly.

Granted that's been almost exclusively with x86(-64) targets, but...

tarcieri:
Haha, and now poking at implementing ECDSA this seems nice, especially in conjunction with the `ecdsa::hazmat::SignPrimitive` I defined, so I guess I'll be reviving it in the PR where I'm trying to implement ECDSA for secp256k1 😅

Issue #24, opened 2020-04-29
Improve secp256k1 field multiplication speeds


tuxxy:
Presently, the `k256` crate implements montgomery mulmod for multiplication/reduction. @hdevalence brought up a point in #19 that there may be a method that leverages the special form of the secp256k1 modulus to improve multiplication performance.

Performance of a scalar multiplication for the `k256` crate:
```
test tests::k256_scalar_mul    ... bench:     197,252 ns/iter (+/- 2,838)
```
It would be nice to see a significant performance increase for this.

tarcieri:
@tuxxy would you consider this addressed by #59?

tuxxy:
Not quite. I'd like to write some benchmarks for myself and talk with @fjarri a bit before it gets closed.

fjarri:
What multiplication are we talking about here? Scalar-scalar, field-field, or point-scalar? The latter can certainly be improved, I'm planning to look into it.

Where's this benchmark you're quoting, by the way?

tuxxy:
I've unfortunately lost the benchmark I originally wrote, but it measured point-scalar. It would be nice to see two benchmarks, 1) comparing point-scalar operations to `libsecp256k1` and 2) measuring scalar-scalar operations (maybe against a variety of implementations?).

tarcieri:
I think we can go ahead and close this out. Further optimizations would certainly be appreciated, but I think we're ok for now.

PR #82, opened 2020-07-18
Improved k256 point-scalar multiplication

fjarri:
May fix issue #24, or at least come close to fixing it. On my machine, the multiplication speed without endomorphism is 104us, and with endomorphism 75us.

Current version uses a simple windowed multiplication. `libsecp256k1` has the following improvements on top of that:

1. a use of the endomorphism

This is implemented, and enabled by `endomorphism-mul`. I have some doubts about the validity of the approximation used in `libsecp256k1`, see my comment in `mul.rs` ("@fjarri"). 

2. a more advanced windowing algorithm, requiring precomputation of only odd multiples of the point

I checked it out, and it performs the same as the one from Ristretto, with the latter being considerably simpler. So I used the Ristretto one. (It does have a hardcoded window size, but generalizing to other window sizes is trivial, if needed).

3. a use of isomorphism to calculate precomputed multiples in the affine form, and use `add_mixed()` later to add them to the (projective) accumulator

This introduces non-constant-timeness wrt the point, and requires a check for the point not being infinite. Not for this PR.

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -66,7 +66,8 @@ impl Scalar {
 
     /// Attempts to parse the given byte array as a scalar.
     /// Does not check the result for being in the correct range.
-    pub const fn from_bytes_unchecked(bytes: &[u8; 32]) -> Self {
+    #[cfg(feature = "endomorphism-mul")]
+    pub(crate) const fn from_bytes_unchecked(bytes: &[u8; 32]) -> Self {
         Self(ScalarImpl::from_bytes_unchecked(bytes))
     }
 
@@ -510,6 +511,7 @@ mod tests {
         }
 
         #[test]
+        #[cfg(feature = "endomorphism-mul")]
         fn fuzzy_roundtrip_to_bytes_unchecked(a in scalar()) {
             let bytes = a.to_bytes();
             let a_back = Scalar::from_bytes_unchecked(&bytes);
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -64,6 +64,12 @@ impl Scalar {
         self.0.truncate_to_u32()
     }
 
+    /// Attempts to parse the given byte array as a scalar.
+    /// Does not check the result for being in the correct range.
+    pub const fn from_bytes_unchecked(bytes: &[u8; 32]) -> Self {
+        Self(ScalarImpl::from_bytes_unchecked(bytes))
+    }
+
     /// Attempts to parse the given byte array as an SEC-1-encoded scalar.
     ///
     /// Returns None if the byte array does not contain a big-endian integer in the range
@@ -224,6 +230,18 @@ impl Scalar {
             }
         }
     }
+
+    /// If `flag` evaluates to `true`, adds `(1 << bit)` to `self`.
+    pub fn conditional_add_bit(&self, bit: usize, flag: Choice) -> Self {
+        Self(self.0.conditional_add_bit(bit, flag))
+    }
+
+    /// Multiplies `self` by `b` (without modulo reduction) divide the result by `2^shift`
+    /// (rounding to the nearest integer).
+    /// Variable time in `shift`.
+    pub fn mul_shift_var(&self, b: &Scalar, shift: usize) -> Self {
+        Self(self.0.mul_shift_var(&(b.0), shift))
+    }
 }
 
 impl Shr<usize> for Scalar {
@@ -261,6 +279,14 @@ impl PartialEq for Scalar {
 }
 
 impl Neg for Scalar {
+    type Output = Scalar;
+
+    fn neg(self) -> Scalar {
+        self.negate()
+    }
+}
+
+impl Neg for &Scalar {
     type Output = Scalar;
 
     fn neg(self) -> Scalar {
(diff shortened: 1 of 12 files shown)


tarcieri:
My understanding is the libsecp256k1 C library (and its Rust wrapper) disable this optimization by-default out of concern for patents, notably https://patentimages.storage.googleapis.com/bb/75/9e/a4353a2158ea1a/US7995752.pdf (which looks like it might expire this September, but might also get renewed).

While I think it's okay to include in the codebase, I think like libsecp256k1 it needs to be gated by a Cargo feature and disabled by-default such that it can only be used in jurisdictions where the patent does not apply.

fjarri:
Hm, what exactly is protected by this patent? The whole usage of the endomorphism, or just the specific method of decomposing the scalar (as described in the paper of the patent holders, R. J. Lambert - which was published in 2001, so I don't know how this patent, filed in 2005, is even valid)? In the latter case, I would prefer not to use that anyway, since, as I mentioned earlier, it doesn't actually have a bound on its results. There is an [alternative method](https://link.springer.com/chapter/10.1007%2F3-540-45664-3_23), published in 2002 - would it be free of the patent?

tarcieri:
I’m certainly not a patent lawyer. All I know is the libsecp256k1 developers have been concerned about this particular patent and how it applies to endomorphism optimizations.

fjarri:
Some more investigation results:

- While the patent does describe an older [GLV method](https://www.iacr.org/archive/crypto2001/21390189.pdf), `libsecp256k1` uses the more reliable [PJKL method](https://link.springer.com/chapter/10.1007%2F3-540-45664-3_23) of scalar decomposition.

Now I'm not a patent lawyer myself, but I find it highly objectionable that a patent could be filed several years after the research is published, and cover not just a specific algorithm, but the general idea of doing a decomposition (otherwise `libsecp256k1` would be free of it). I wonder how enforceable it actually is.

- The decomposition method requires the computation of a couple of expressions like `round(k * a / modulus)` where `k` is the scalar, and `a` is a known constant. In `libsecp256k1` they are approximated as `round(k * g / 2^n)`, where `g = a * 2^n // modulus` (floor division). The code references [this paper](https://link.springer.com/article/10.1007%2Fs13389-012-0029-z), but it doesn't provide any clues on how `n` was chosen. I did some experiments, and for the `n=272` used in the code it is quite common to get off-by-1 errors. 

It does not lead to the breach of decomposition contract, since the code only calculates one part (`k2`) and calculates the other part as `k1 = k - lambda * k2 mod modulus`, instead of calculating `k1` as the original algorithm suggests. But an off-by-1 in `k2` can lead to the upper bound on `k1` not being `sqrt(modulus)` anymore (which will lead to the multiplication failure because it assumes that `k1` and `k2` are both fit into 128 bits). Not sure what was the reason for this choice - 1) a mistake; 2) the probability of the error is low enough to be negligible; 3) I am misunderstanding something.

tarcieri:
> Now I'm not a patent lawyer myself, but I find it highly objectionable that a patent could be filed several years after the research is published

That's interesting. I do recall this claim from djb's patent research:

https://cr.yp.to/patents/us/4200770.html

> Under United States case law, a document has been published if it ``has been disseminated or otherwise made available to the extent that persons interested and ordinarily skilled in the subject matter or art, exercising reasonable diligence, can locate it.'' A patent is automatically invalid if the patented invention was published more than a year before the patent's filing date.

tarcieri:
> This introduces non-constant-timeness wrt the point, and requires a check for the point not being infinite. Not sure if this should be implemented or not.

A variable time scalar mult which is faster than the constant time version is still useful for signature verification

k256/src/arithmetic/scalar.rs
```diff
@@ -65,6 +65,12 @@ impl Scalar {
         self.0.truncate_to_u32()
     }
 
+    /// Attempts to parse the given byte array as a scalar.
+    /// Does not check the result for being in the correct range.
+    pub const fn from_bytes_unchecked(bytes: &[u8; 32]) -> Self {
```
tarcieri, line 70: Is there a specific reason why this needs to be `pub` as opposed to `pub(crate)`?

fjarri: Not really, I didn't think about what should be exposed to the user. This method is useful for defining `Scalar` constants, which I suppose may be needed sometimes? But if you want to preserve the current public API, this can be set to `pub(crate)` for now.

tarcieri: Given that this method can violate the invariant of what a `Scalar` is supposed to contain, I'd suggest making it private for now.

I think ideally a `pub const fn from_bytes(...)` method would be possible.

fjarri: Yes, I think `from_bytes()` can be easily made into a `const fn` (by using the existing `unchecked` code). Probably can wait for another PR though.

k256/src/arithmetic/scalar/scalar_4x64.rs
```diff
@@ -313,6 +331,53 @@ impl Scalar4x64 {
...
+            (((bit >> 6) == 3) as u64) << bit_lo,
+        ]);
+
+        Self::conditional_select(self, &(self.add(&w)), flag)
+    }
+
+    pub fn mul_shift_var(&self, b: &Self, shift: usize) -> Self {
+        debug_assert!(shift >= 256);
+
+        fn ifelse(c: bool, x: u64, y: u64) -> u64 { if c {x} else {y} }
```
tarcieri, line 354: That's uhh... interesting 😉

k256/src/lib.rs
```diff
@@ -16,6 +16,8 @@
 
 #[cfg(feature = "arithmetic")]
 mod arithmetic;
+#[cfg(feature = "arithmetic")]
+mod mul;
```
tarcieri, line 20: I feel like there's maybe a larger discussion to be had here about refactoring the `arithmetic` module, but we can save that for another PR.

PR #83, opened 2020-07-22
p256: Implement scalar arithmetic via Barrett reduction

tarcieri:
This PR cherry picks @nickray's original #56 onto the latest code.

I also removed the `expose-arithmetic` feature/gating as I think it should no longer be necessary.

With this I think it should be possible to implement `ecdsa::hazmat::SignPrimitive` ala the implementation in #79 for secp256k1.

### DIFF
--- a/p256/src/arithmetic/field.rs
+++ b/p256/src/arithmetic/field.rs
@@ -9,6 +9,9 @@ use elliptic_curve::rand_core::{CryptoRng, RngCore};
 
 use super::util::{adc, mac, sbb};
 
+/// The number of 64-bit limbs used to represent a [`FieldElement`].
+const LIMBS: usize = 4;
+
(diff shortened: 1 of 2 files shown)


tarcieri:
@nickray was there anything else incomplete about this PR?

I'd like to go ahead and merge it and cherry-pick parts of #57 onto the `SignPrimitive` interface ala #79

nickray:
@tarcieri sure, go ahead with both, happy to see signatures appear!

Nothing obvious missing, apart from a potential code detail review and more testing, which perhaps comes later?

tarcieri:
Sounds good, will go ahead and move forward with this.

I think if we can get ECDSA to a PoC stage, there are some other interested parties who can help work on improving P-256 scalar arithmetic (and ideally with that, improving testing).

PR #85, opened 2020-07-23
k256: use `ecdsa/dev` test macros + API support

tarcieri:
Converts ECDSA tests to use the `new_signing_test!` and `new_verification_test!` macros added to the new `dev` feature of the `ecdsa` crate in RustCrypto/signatures#103.

The macros use a few new APIs which were also added in this PR, namely `AffinePoint::{from_compressed_point, from_uncompressed_point}` which provide a more direct conversion path from SEC-1 encodings to `AffinePoint` which skip the `PublicKey` interface (which is more of a high-level, user-facing API).

### DIFF
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -40,49 +40,49 @@ jobs:
           command: fmt
           args: --all -- --check
 
-  codecov:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-
-      - name: Cache cargo registry
-        uses: actions/cache@v1
-        with:
-          path: ~/.cargo/registry
-          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('Cargo.lock') }}
-
-      - name: Cache cargo index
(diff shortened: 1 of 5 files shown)


.github/workflows/workspace.yml
```diff
@@ -40,49 +40,49 @@ jobs:
...
+#          args: --all -- --test-threads 1
+#
+#      - name: Upload to codecov.io
+#        uses: codecov/codecov-action@v1.0.10
+#
+#      - name: Archive code coverage results
+#        uses: actions/upload-artifact@v1
+#        with:
+#          name: code-coverage-report
+#          path: cobertura.xml
```
tarcieri, line 88: Not sure why but `cargo-tarpaulin` is failing:

https://github.com/RustCrypto/elliptic-curves/pull/85/checks?check_run_id=904067088

I attempted to reproduce it locally but cannot. It doesn't appear to be cache-related as this commit is a cache breaker due to the `Cargo.lock` changes.

I'd like to do a pass over `p256` to get it in to a similar state as this PR (as part of #84) after which I can try re-enabling this.

PR #98, opened 2020-07-28
Deactivate default features of sha2 in Cargo.toml

Sawchord:
Set default-features = false for sha2 in Cargo.toml such that the crate builds in no_std environment.


tarcieri:
Sorry about that! Before doing the next release I plan on adding CI checks for more feature combinations.

PR #99, opened 2020-07-28
k256/p256: use new `SignPrimitive` API; add `BlindedScalar`

tarcieri:
Updates both crates with the `SignPrimitive` changes from RustCrypto/signatures#107.

These changes removed the `masking_scalar` argument from the trait, and replaced it with trait bounds that make it possible to substitute a blinded scalar.

In the `p256` crate (which is the only one that presently implements a variable-time inversion) the `masking_scalar` is replaced with a `BlindedScalar` type that implements the previous blinded inversion.

The implementation it uses (including `Scalar::invert_vartime`) could potentially be made generic and extracted into the `elliptic-curve` crate, allowing it to be used with any curve which implements the arithmetic primitives used in the blinded inversion implementation.

However, for now, this PR leaves it in `p256`, and therefore at least has feature parity with the old implementation.

### DIFF
--- a/p256/src/ecdsa.rs
+++ b/p256/src/ecdsa.rs
@@ -9,7 +9,7 @@ use {
         hazmat::{SignPrimitive, VerifyPrimitive},
         Error,
     },
-    elliptic_curve::subtle::CtOption,
+    elliptic_curve::{ops::Invert, subtle::CtOption},
 };
 
 /// ECDSA/P-256 signature (fixed-size)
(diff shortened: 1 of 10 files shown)


p256/src/ecdsa.rs
```diff
@@ -107,6 +105,24 @@ mod tests {
...
+        let d = Scalar::from_bytes(vector.d.try_into().unwrap()).unwrap();
+        let k = Scalar::from_bytes(vector.k.try_into().unwrap()).unwrap();
+        let k_blinded = BlindedScalar::new(k, &mut OsRng);
+        let sig = d
+            .try_sign_prehashed(&k_blinded, GenericArray::from_slice(vector.m))
+            .unwrap();
+
+        assert_eq!(vector.r, sig.r().as_slice());
+        assert_eq!(vector.s, sig.s().as_slice());
+    }
```
tarcieri, line 127: FYI @nickray, here's the proposed alternative for `masking_scalar`.

I had it take an RNG argument directly, but we could also add an API to construct one from two `Scalar` values you generate however you want (perhaps that could be `new`, and the RNG-based one could be `generate`)

PR #122, opened 2020-08-06
A change corresponding to normalize_low() becoming pure in `ecdsa`

fjarri:
See https://github.com/RustCrypto/signatures/pull/119


tarcieri:
I think you need to `cargo update` to get it to pick up the merged changes to the other crates

tarcieri:
This was addressed by #123

Issue #119, opened 2020-08-05
New releases of elliptic curve crates


tarcieri:
I'd like to cut releases of the following crates relatively soon (next few days / week):

- [`elliptic-curve`](https://github.com/RustCrypto/traits/tree/master/elliptic-curve) v0.5.0
- [`ecdsa`](https://github.com/RustCrypto/signatures/tree/master/ecdsa) v0.7.0
- `k256` v0.4.0
- `p256` v0.4.0
- `p384` v0.3.0

These will be the first releases supporting an integrated ECDSA implementation as well as initial (and rather rudimentary) support for implementing algorithms generically over elliptic curve groups.

If there's anything else you'd like to get into these releases, let me know, otherwise I plan on doing it soon.

When I say that, I'm also looking for small items rather than more significant changes. There's plenty of bigger issues to address (e.g. eliminating fallible APIs via eliminating the invalid representations that cause them in `PublicKey`/`SecretKey`) but I'd rather not make any bigger items release blockers at this point and we can continue making large and breaking changes in the next release.

cc @str4d @tuxxy @fjarri @nickray

fjarri:
I've got some problems in my library that uses `ecdsa` and `k256` - when I updated to the current state of `elliptic-curves`, `traits` and `signatures` from ~week old commits, I started getting incorrect results (some internal verification fails). Currently investigating what the reason.

Also, there is a bug in `k256/ecdsa/signer.rs`: in `try_sign_recoverable_prehashed()`, `let r_is_odd = R.y.is_odd();` should read `let r_is_odd = R.y.normalize().is_odd();`. There is an assertion that checks that you don't call `is_odd()` for non-normalized `FieldElement`, so it was supposed to be caught in testing. But, I guess, we only run tests for two feature sets:
- default features (which doesn't include `ecdsa` and therefore does not cover `signer.rs`, and
- all features (which includes `field-montgomery` and therefore does not check for normalization)
I can make a separate PR for this, but I wanted to figure out the remaining problems first.

tarcieri:
Okay, sounds good. I'll try to wrap up what I'm working on and let you have at it.

If there's no pressing reason not to, I might cut releases of the `elliptic-curve` and `ecdsa` crates in the meantime.

fjarri:
[...]

PR #123, opened 2020-08-07
k256: impl new RecoverableSignPrimitive API

tarcieri:
Implements the changes made in RustCrypto/signatures#120.

These changes have the `k256` always compute a `recoverable::Signature` via `RecoverableSignPrimitive` and use the blanket impl of `SignPrimitive` for `RecoverableSignPrimitive` to convert it to a regular `k256::ecdsa::Signature` in `ecdsa::signer::Signer<Secp256k1>`.

Ideally the `k256::ecdsa::Signer` newtype could go away and this could all be handled by `ecdsa::signer::Signer` via blanket and conditional impls, however I'm not sure how to prevent the impls of `RandomizedSigner` from overlapping in a generic context.

### DIFF
--- a/k256/src/ecdsa/normalize.rs
+++ b/k256/src/ecdsa/normalize.rs
@@ -6,12 +6,11 @@ use crate::Scalar;
 use ecdsa_core::NormalizeLow;
 
 impl NormalizeLow for Scalar {
-    fn normalize_low(&mut self) -> bool {
+    fn normalize_low(&self) -> (Self, bool) {
         if self.is_high().into() {
-            *self = -*self;
-            true
+            (-self, true)
         } else {
-            false
+            (*self, false)
         }
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -227,7 +227,7 @@ dependencies = [
 [[package]]
 name = "ecdsa"
 version = "0.7.0-pre"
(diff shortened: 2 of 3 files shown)


tarcieri:
I tried moving this into the `ecdsa` crate, but as noted in the description, to prevent the two `RandomizedSigner` impls from overlapping, I'd need to ensure `RecoverableSignPrimitive<C>::RecoverableSignature` is not the same type as `ecdsa::Signature<C>` to prevent the impls from overlapping.

k256/src/ecdsa/normalize.rs
```diff
@@ -6,12 +6,11 @@ use crate::Scalar;
...
         if self.is_high().into() {
-            *self = -*self;
-            true
+            (-self, true)
         } else {
-            false
+            (*self, false)
         }
     }
 }
```
tarcieri, line 16: NOTE: this PR incorporates #122 (although I needed to deref in the `else` clause).

tarcieri:
Going to merge this as I believe it addresses the main problem we've been discussing in #119.

I think there are a lot of other ways to potentially go with this, either possibly hoisting the whole thing up into the `ecdsa` crate somehow, or if that isn't feasible, getting rid of `RecoverableSignPrimitive` and just having a private function at the core of both implementations which returns the `recoverable::Id` (and discards it in `SignPrimitive`). But for now, I consider this progress.

PR #162, opened 2020-09-04
Remove `rand` feature; bump `elliptic-curve` and `ecdsa`

tarcieri:
Updates to use the `group` crate. See: RustCrypto/traits#287.

This crate has a hard `rand_core` dependency so this commit gets rid of the `rand` features across the board and makes them mandatory.

(Even if we don't end up shipping the `group` crate this release, that's probably for the best to keep the number of features down)


tests/README.md
```diff
@@ -0,0 +1,16 @@
...
+It contains small test crates in their own isolated workspace which ensure that
+these features are not activated when linking and therefore that the crates
+will link in `no_std` environments when consumed as a dependency of another
+`no_std`-compatible crate.
+
+Here are upstream issues tracking the problem:
+
+- [#7914: Tracking issue for `-Z features=itarget`](https://github.com/rust-lang/cargo/issues/7914)
+- [#7915: Tracking issue for `-Z features=host_dep`](https://github.com/rust-lang/cargo/issues/7915)
+- [#7916: Tracking issue for `-Z features=dev_dep`](https://github.com/rust-lang/cargo/issues/7916])
```
tarcieri, line 16: A different solution it's probably worth exploring here is moving `proptest`s and criterion benchmarks into their own crates so the `[dev-dependencies]` of the root workspace remain clean.

newpavlov: I am not sure about the `proptest`, but I think that the criterion benchmarks should definitely be moved into a separate crate with its own workspace as was done in other repositories.

PR #164, opened 2020-09-04
k256: impl `ff` and `group` traits

tarcieri:
- Impls `ff::{Field, PrimeField}` on `Scalar`
- Impls `group::{Group, Curve}` on `ProjectivePoint`

### DIFF
--- a/k256/Cargo.toml
+++ b/k256/Cargo.toml
@@ -16,6 +16,7 @@ categories = ["cryptography", "cryptography::cryptocurrencies", "no-std"]
 keywords = ["bitcoin", "crypto", "ecc", "ethereum", "secp256k1"]
 
 [dependencies]
+byteorder = { version = "1", default-features = false }
 cfg-if = "0.1"
 ecdsa-core = { version = "0.7", package = "ecdsa", optional = true, default-features = false }
 elliptic-curve = { version = "0.5", default-features = false }
--- a/k256/src/arithmetic/projective.rs
+++ b/k256/src/arithmetic/projective.rs
@@ -2,7 +2,7 @@
 
 use super::{AffinePoint, FieldElement, Scalar, CURVE_EQUATION_B_SINGLE};
 use core::{
-    fmt::{self, Display},
+    fmt,
     iter::Sum,
     ops::{Add, AddAssign, Neg, Sub, SubAssign},
 };
@@ -448,12 +448,9 @@ impl<'a> Neg for &'a ProjectivePoint {
     }
 }
 
-// TODO(tarcieri): double check how this should be implemented
-impl Display for ProjectivePoint {
+impl fmt::Display for ProjectivePoint {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{:X}", self.x.to_bytes())?;
-        write!(f, "{:X}", self.y.to_bytes())?;
-        write!(f, "{:X}", self.z.to_bytes())
+        write!(f, "{:?}", self)
     }
 }
 
--- a/k256/src/arithmetic/projective.rs
+++ b/k256/src/arithmetic/projective.rs
@@ -241,7 +241,7 @@ impl Group for ProjectivePoint {
     type Scalar = Scalar;
 
     fn random(rng: impl RngCore) -> Self {
-        Self::generator() * &Scalar::generate_vartime(rng)
+        Self::generator() * Scalar::generate_vartime(rng)
     }
 
     fn identity() -> Self {
@@ -265,9 +265,8 @@ impl Group for ProjectivePoint {
 impl group::Curve for ProjectivePoint {
     type AffineRepr = AffinePoint;
 
-    // TODO(tarcieri): avoid panicking on identity
     fn to_affine(&self) -> AffinePoint {
-        ProjectivePoint::to_affine(self).unwrap()
+        ProjectivePoint::to_affine(self)
     }
 }
 
@@ -522,7 +521,7 @@ mod tests {
                 ADD_TEST_VECTORS[i]
             );
 
-            p = p + &generator;
+            p += &generator;
         }
     }
 
@@ -541,7 +540,7 @@ mod tests {
                 ADD_TEST_VECTORS[i]
             );
 
-            p = p + &generator;
+            p += &generator;
         }
     }
 
--- a/k256/src/arithmetic/projective.rs
+++ b/k256/src/arithmetic/projective.rs
@@ -1,9 +1,15 @@
 //! Projective points
 
-use super::{AffinePoint, FieldElement, CURVE_EQUATION_B_SINGLE};
-use core::ops::{Add, AddAssign, Neg, Sub, SubAssign};
+use super::{AffinePoint, FieldElement, Scalar, CURVE_EQUATION_B_SINGLE};
+use core::{
+    fmt::{self, Display},
+    iter::Sum,
+    ops::{Add, AddAssign, Neg, Sub, SubAssign},
+};
 use elliptic_curve::{
+    group::{self, Group},
     point::Generator,
+    rand_core::RngCore,
     subtle::{Choice, ConditionallySelectable, ConstantTimeEq},
 };
 
@@ -57,6 +63,8 @@ impl PartialEq for ProjectivePoint {
         self.ct_eq(other).into()
     }
 }
+
+impl Eq for ProjectivePoint {}
 
 impl ProjectivePoint {
     /// Returns the additive identity of SECP256k1, also known as the "neutral element" or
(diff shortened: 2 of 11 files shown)


k256/Cargo.toml
```diff
@@ -16,6 +16,7 @@ categories = ["cryptography", "cryptography::cryptocurrencies", "no-std"]
 keywords = ["bitcoin", "crypto", "ecc", "ethereum", "secp256k1"]
 
 [dependencies]
+byteorder = { version = "1", default-features = false }
```
tarcieri, line 19: NOTE: this will go away in the next release of `ff`, I believe

k256/src/arithmetic/scalar.rs
```diff
@@ -405,6 +542,12 @@ impl Generate for Scalar {
     }
 }
 
+impl Display for Scalar {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:X}", self.to_bytes())
+    }
+}
```
tarcieri, line 549: Wasn't sure exactly what a `Display` impl should do here, so I had it spit out upper case serialized hex

k256/src/arithmetic/projective.rs
```diff
@@ -346,6 +444,15 @@ impl<'a> Neg for &'a ProjectivePoint {
...
 }
 
+// TODO(tarcieri): double check how this should be implemented
+impl Display for ProjectivePoint {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:X}", self.x.to_bytes())?;
+        write!(f, "{:X}", self.y.to_bytes())?;
+        write!(f, "{:X}", self.z.to_bytes())
+    }
+}
```
tarcieri, line 454: What should this `Display` impl actually look like?

k256/src/arithmetic/projective.rs
```diff
@@ -280,6 +338,32 @@ impl AddAssign<AffinePoint> for ProjectivePoint {
...
+    fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+        iter.fold(ProjectivePoint::identity(), |a, b| a + b)
+    }
+}
+
+impl<'a> Sum<&'a ProjectivePoint> for ProjectivePoint {
+    fn sum<I: Iterator<Item = &'a ProjectivePoint>>(iter: I) -> Self {
+        iter.fold(ProjectivePoint::identity(), |a, b| a + b)
+    }
+}
```
tarcieri, line 357: Not sure I implemented these correctly... I suppose the implementation could also be unified

k256/src/arithmetic/projective.rs
```diff
@@ -224,6 +232,40 @@ impl ProjectivePoint {
...
+
+impl group::Curve for ProjectivePoint {
+    type AffineRepr = AffinePoint;
+
+    // TODO(tarcieri): avoid panicking on identity
+    fn to_affine(&self) -> AffinePoint {
+        ProjectivePoint::to_affine(self).unwrap()
+    }
```
tarcieri, line 266: Presently there's no way of representing the identity as an `AffinePoint`.

Perhaps we should just turn it into an enum with an `AffinePoint::Identity` variant?

k256/src/arithmetic/scalar.rs
```diff
@@ -227,6 +238,82 @@ impl Scalar {
...
+
+    #[must_use]
+    fn square(&self) -> Self {
+        Scalar::square(self)
+    }
+
+    #[must_use]
+    fn double(&self) -> Self {
+        todo!();
+    }
```
tarcieri, line 266: Stub!

k256/src/arithmetic/scalar.rs
```diff
@@ -227,6 +238,82 @@ impl Scalar {
...
+
+    fn invert(&self) -> CtOption<Self> {
+        Scalar::invert(self)
+    }
+
+    fn sqrt(&self) -> CtOption<Self> {
+        todo!();
+    }
```
tarcieri, line 274: Stub!

k256/src/arithmetic/scalar.rs
```diff
@@ -227,6 +238,82 @@ impl Scalar {
...
+    type ReprEndianness = byteorder::BigEndian;
+
+    const NUM_BITS: u32 = 256;
+    const CAPACITY: u32 = 256;
+    const S: u32 = 6;
+
+    fn from_repr(repr: ElementBytes) -> Option<Self> {
+        let result = Scalar::from_bytes(&repr);
+
+        // TODO(tarcieri): replace with into conversion when available (see subtle#73)
```
tarcieri, line 288: I should really open a PR for this rather than copying this comment everywhere 😅

k256/src/arithmetic/scalar.rs
```diff
@@ -227,6 +238,82 @@ impl Scalar {
...
+        self.to_bytes()
+    }
+
+    fn is_odd(&self) -> bool {
+        self.0.is_odd().into()
+    }
+
+    fn char() -> Self::Repr {
+        unimplemented!(); // removed in newer versions of `ff`
+    }
```
tarcieri, line 306: Stub, but this method will be removed in future versions of `ff` I believe.

k256/src/arithmetic/scalar.rs
```diff
@@ -227,6 +238,82 @@ impl Scalar {
...
+
+    fn multiplicative_generator() -> Self {
+        todo!();
+    }
```
tarcieri, line 310: Stub!

k256/src/arithmetic/scalar.rs
```diff
@@ -227,6 +238,82 @@ impl Scalar {
...
+
+    fn root_of_unity() -> Self {
+        todo!();
+    }
```
tarcieri, line 314: Stub!

k256/src/arithmetic/scalar/scalar_8x32.rs
```diff
@@ -528,6 +533,12 @@ impl From<u32> for Scalar8x32 {
     }
 }
 
+impl From<u64> for Scalar8x32 {
+    fn from(k: u64) -> Self {
+        Self([(k & 0xFFFF) as u32, (k >> 32) as u32, 0, 0, 0, 0, 0, 0])
```
tarcieri, line 538: I think I did this right, but someone please double check me (or maybe I should write a test)

tarcieri:
Lots of clippy complaints about unnecessary borrows now. I'm wondering if we need to add `#[inline]` on the various owned arithmetic operations to remove those without impacting performance.

It also seems like there are a lot of inherent methods on `Scalar` and `ProjectivePoint` which also have similar or identical methods on the trait. Perhaps we should get rid of the duplicated inherent methods and just use the trait?

k256/src/arithmetic/projective.rs
str4d: Yes, and yes: https://github.com/zkcrypto/jubjub/blob/b7374764b47c97a126a0d171d0d0e95459fd7fdf/src/lib.rs#L168-L178

tarcieri:
Applied clippy's suggested fixes with the new arithmetic impls on owned types. I do really like how it improves the code quality (by removing "eye of sauron"-style borrowed arithmetic).

Ran the existing benchmarks... it looked like mostly improvements except for `normalize`:

Screen Shot 2020-09-04 at 11 56 37 AM

```
field element operations/normalize_weak
                        time:   [1.6940 ns 1.7020 ns 1.7113 ns]
                        change: [+4.7960% +5.6804% +6.5209%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  1 (1.00%) high severe
field element operations/normalize
                        time:   [1.7010 ns 1.7075 ns 1.7142 ns]
                        change: [+1.4923% +3.0551% +4.6100%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

cc @fjarri

fjarri:
I wouldn't worry about it. The change is small, and when we're talking about an operation that takes literally 1ns, I doubt that these measurements are very accurate. On my laptop it takes ~4ns, and I don't see any change between this PR and `master`.

k256/src/arithmetic/projective.rs
BlackHoleFox: This seems like a good case for a [debug struct](https://doc.rust-lang.org/stable/std/fmt/struct.Formatter.html#method.debug_struct).

So maybe, it could look something like this: 
```rust
 fmt.debug_struct("ProjectivePoint")
        .field("x point: {:X}", self.x.to_bytes())
        .field("y point: {:X}", self.y.to_bytes())
        .field("z point: {:X}", self.z.to_bytes())
        .finish()
```

tarcieri: This is a `Display` impl though, so I'd expect it to give something easily parsable.

Looking at `bls12_381` though, `Display` is impl'd equivalently to `Debug`, so I guess I'll just do that.

k256/src/arithmetic/scalar.rs
tarcieri: Oh hey look, someone else opened one already: https://github.com/dalek-cryptography/subtle/pull/75

k256/src/arithmetic/projective.rs
tarcieri: Updated to use the same impl as `bls12_381`

k256/src/arithmetic/projective.rs
tarcieri: Added `AffinePoint` identity support in #165

k256/src/arithmetic/scalar.rs
tarcieri: It seems there's some code we can repurpose from the `p256` crate to impl this, but in the meantime I've more clearly marked it's a stub and would like to address that in a follow-up PR.

PR #167, opened 2020-09-05
p256: implement `AffinePoint::identity()` and `::is_identity()`

tarcieri:
This is a corresponding change ala #165, but for the `p256` crate.

### DIFF
--- a/k256/src/arithmetic/affine.rs
+++ b/k256/src/arithmetic/affine.rs
@@ -139,8 +139,11 @@ impl FromEncodedPoint<Secp256k1> for AffinePoint {
                         // Check that the point is on the curve
                         let lhs = (y * &y).negate(1);
                         let rhs = x * &x * &x + &CURVE_EQUATION_B;
-                        let infinity = Choice::from(0);
-                        let point = AffinePoint { x, y, infinity };
+                        let point = AffinePoint {
+                            x,
+                            y,
+                            infinity: Choice::from(0),
+                        };
                         CtOption::new(point, (lhs + &rhs).normalizes_to_zero())
                     })
                 })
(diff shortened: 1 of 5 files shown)


nickray:
No strong opinions, but from a mathematical point of view, an "affine point" is one in the affine plane (i.e., *not* the point at infinity). So then what one might want is a kind of `CtEnum` consisting of either an affine point (the (x,y) coordinates, that must lie on the curve) or the infinite point.

tarcieri:
@nickray I agree, but unfortunately there is not such a thing as a `CtEnum` AFAIK (at least in Rust)

nickray:
Right, I made that up (as I was thinking of just an enum at first, and then saw your careful use of constant-time operations). You could get by with a `CtOption<AffinePoint>` (where AffinePoint is the coordinate pair) if you want something your ProjectivePoint maps to.

tarcieri:
That’s how things previously worked prior to this PR. 

The goal of this PR though is to support an infallible bidirectional conversion between `AffinePoint` and `ProjectivePoint` in service of supporting the `group::Curve` trait (#164).

It also makes `AffinePoint` much more `subtle::CtOption`-friendly in general, since several methods on `CtOption` are bounded on a `Default` impl (which `AffinePoint::identity()` makes possible)

nickray:
I'm just saying this AffinePoint is a misnomer (and there is no such bijection); but it's fine -- one just has to mentally substitute either AffinePointOrInfinity or ProjectivePointWithNormalizedCoordinates or some such thing wherever it says AffinePoint :)

tarcieri:
Yeah. I tried introducing a `NonZeroScalar` type to avoid it (https://github.com/RustCrypto/traits/pull/241), which as it were ended up being helpful for other reasons (https://github.com/RustCrypto/elliptic-curves/pull/144).

I think unless you're dealing explicitly with scalars which by definition must be non-zero (e.g. in the ECDSA case mentioned above) it's still really nice to avoid having a fallible conversion, and especially in the case of writing constant-time code this particular conversion has been quite annoying, especially because without an `AffinePoint::default` impl it isn't possible to use the `Option`-like combinators on `CtOption`.

PR #174, opened 2020-09-09
k256/p256: bump `ff` and `group` deps to v0.8; MSRV 1.44+

tarcieri:
Previously the `elliptic-curve` crate was depending on pre-releases of `ff` and `group` sourced via git. It was upgraded to use the final releases in RustCrypto/traits#292.

Because these crates depend on `bitvec` and its transitive dependencies, an `arithmetic` feature was added to `elliptic-curve` in RustCrypto/traits#293.

This commit upgrades the `k256` and `p256` crates to use the final releases of `ff` and `group`. This ended up actually being a relatively straightforward upgrade because we were already running off the latest versions sourced from git.

The major notable change is `PrimeField::ReprBits` and the `to_le_bits` and `char_le_bits` methods for obtaining `bitvec`-friendly types for representing field elements.

### DIFF
--- a/p256/src/arithmetic/scalar.rs
+++ b/p256/src/arithmetic/scalar.rs
@@ -5,7 +5,6 @@ pub mod blinding;
 use crate::{ElementBytes, NistP256, SecretKey};
 use core::{
     convert::TryInto,
-    fmt,
     ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
 };
 use elliptic_curve::{
@@ -69,6 +68,9 @@ pub const MU: [u64; 5] = [
 /// Non-zero scalar value.
 pub type NonZeroScalar = elliptic_curve::scalar::NonZeroScalar<NistP256>;
 
+/// NIST P-256 field element serialized as bits.
+pub type ScalarBits = elliptic_curve::scalar::ScalarBits<NistP256>;
+
 /// An element in the finite field modulo n.
 // The internal representation is as little-endian ordered u64 words.
 #[derive(Clone, Copy, Debug, Default)]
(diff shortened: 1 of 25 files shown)


tarcieri:
It seems `p256` isn't building on 32-bit targets because it lacks a u32x8 scalar field backend, and `BitView` is not impl'd for `[u64; N]` on 32-bit architectures:

```
error[E0277]: the trait bound `[u64; 4]: elliptic_curve::BitView` is not satisfied
```

Interestingly enough, since `k256` does have a 32-bit scalar field backend (thanks @fjarri), this isn't an issue for that crate!

I'll look at working around the issue for now, but it'd sure be great to have a 32-bit scalar field implementation as well, particularly if `ff_derive` can generate one.

/cc @str4d

p256/src/arithmetic/scalar.rs
```diff
@@ -829,6 +844,20 @@ impl Invert for Scalar {
     }
 }
 
+#[cfg(target_pointer_width = "32")]
+impl From<&Scalar> for FieldBits {
+    fn from(_scalar: &Scalar) -> FieldBits {
+        todo!();
```
tarcieri, line 850: TODO: will implement this before merging

p256/src/arithmetic/scalar.rs
```diff
@@ -829,6 +847,27 @@ impl Invert for Scalar {
...
+        let mut output = [0u32; 8];
+
+        for (input, output) in scalar.0.iter().zip(output.chunks_mut(2)) {
+            output[0] = (input >> 32) as u32;
+            output[1] = (input & 0xFFFF) as u32;
+        }
+
+        output.into()
+    }
+}
```
tarcieri, line 862: TODO: this could use a test

p256/src/arithmetic/scalar.rs
tarcieri: Implemented. Just needs a test now.

p256/src/arithmetic/scalar.rs
tarcieri: Test added

PR #184, opened 2020-09-12
P256 pk recovery

evan-brass:
I'm new to cryptography, but I tried to implement public key recovery for p256 from SEC 1-v2 section 4.1.6

It uses an iterator and is variable time.

### DIFF
--- a/p256/Cargo.toml
+++ b/p256/Cargo.toml
@@ -26,6 +26,8 @@ hex = "0.4" # TODO: switch to hex-literal
 hex-literal = "0.3"
 proptest = "0.10"
 rand_core = { version = "0.5", features = ["getrandom"] }
+rand = "0.7.3"
+sha2 = "0.9"
 
 [features]
 default = ["arithmetic", "std"]
--- a/p256/src/ecdsa.rs
+++ b/p256/src/ecdsa.rs
@@ -61,23 +61,6 @@ use {
 /// ECDSA/P-256 signature (fixed-size)
 pub type Signature = ecdsa_core::Signature<NistP256>;
 
-struct VerifyKeyRecoverIter {
-	r: NonZeroScalar,
-	s: NonZeroScalar,
-	e: NonZeroScalar,
-	j: Scalar,
-	invert: bool
-}
-impl VerifyKeyRecoverIter {
-	fn new(r: NonZeroScalar, s: NonZeroScalar, e: Scalar, j: Scalar) -> Self {
-		Self {
-			r, s,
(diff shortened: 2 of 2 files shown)


tarcieri:
Did you take a look at the recoverable signature support in the `k256` crate?

https://docs.rs/k256/0.4.2/k256/ecdsa/recoverable/index.html

If there's genuine interest, this could be generalized into the `ecdsa` crate, although for what it's worth I started down that road and ended up removing it:

https://github.com/RustCrypto/signatures/pull/146

evan-brass:
I looked at your k256 implementation, though I somehow missed that `from_trial_recovery` operates on untagged signatures.

The protocol I'm working with (WebPush) requires the P256 and I'm trying to make the messages as small as possible which is why public key recovery interests me.

The recovery algorithm from SEC1 seemed to lend itself to an iterator, so I thought I'd give it a shot and see what it looked like: that's the only point of this PR.

tarcieri:
The same algorithm implemented in the `k256` crate will work for `p256`. If you're really interested in this, I can revisit putting abstractions for it (back) into the "ecdsa" crate.

There's no standard definition or wire format for recoverable ECDSA/P-256 signatures though, so I'd be somewhat hesitant to include it in the `p256` crate itself. However, the `p256` crate could implement the ([previously removed](https://github.com/RustCrypto/signatures/pull/146)) `RecoverableSignPrimitive`, which you could use in a third-party crate to implement a recoverable signature scheme with a wire format you define.

If you do want to define such a wire representation for your protocol, I would strongly recommend checking out [EIP-2098](https://eips.ethereum.org/EIPS/eip-2098). If you low-S normalize an ECDSA signature (which has the added benefit of making it non-malleable), the high-bit of the S-component of the signature will always be 0, which means it can be repurposed for storing the "recovery ID" (a bitflag computed as `is_r_odd ^ is_s_high`, used ala point compression to recover 𝑘×𝑮 by solving the curve equation).

With such an approach, a "recoverable signature" can be the same size as a normal ECDSA signature (i.e. 64-bytes for ECDSA/P-256), and also non-malleable.

evan-brass:
I'm really interested, but not really skilled and you've given me a lot to look into.  Thank you!

tarcieri:
Feel free to open an issue about supporting generic signature recovery primitives on the `ecdsa` crate repo:

https://github.com/rustcrypto/signatures

tarcieri:
FYI, I'm reviving `RecoverableSignPrimitive`: https://github.com/RustCrypto/signatures/pull/175

PR #214, opened 2020-09-28
k256: 32-bit CI improvements

tarcieri:
Uses the GitHub Actions matrix feature to run the entire suite of tests on both the `i686-unknown-linux-gnu` and `x86_64-unknown-linux-gnu` targets, rather than using the `force-32-bit` feature (which we should eventually find some way to get rid of).

This allows testing on a true 32-bit target, rather than trying to enable the 32-bit backend on a 64-bit target (which really tests the `force-32-bit` Cargo feature more than how a 32-bit target behaves)


k256/Cargo.toml
```diff
@@ -27,7 +27,7 @@ sha3 = { version = "0.9", optional = true, default-features = false }
 criterion = "0.3"
 ecdsa-core = { version = "0.8.2", package = "ecdsa", default-features = false, features = ["dev"] }
 hex = "0.4" # TODO: switch to hex-literal
-hex-literal = "0.3"
+hex-literal = "0.2"
```
tarcieri, line 30: It seems like we weren't properly testing MSRV before since we didn't have `override: true`! 😬

tarcieri:
Oh snap, caught a real bug! 😅 

It's in the `ecdsa` crate (the `dev` module is broken on 32-bit platforms). Will push a fix.

PR #231, opened 2020-10-19
CI: test `k256` on PPC32 using cross

tarcieri:
This adds CI on PPC32, which should ensure that the code passes tests on a big endian platform.


tarcieri:
Not quite sure what's going on here... it's a pretty strange error.

It seems possibly related to `bitvec` (or one of its dependencies), or it's something screwed up in the `cross` environment.

PR #269, opened 2020-12-25
add impl for DigestPrimitive on keccak feature

TheRealBluesun:
## Abstract

This PR adds an `impl DigestPrimitive` when the feature `keccak256` is used. This should resolve a build error when using the feature.

### DIFF
--- a/k256/src/ecdsa.rs
+++ b/k256/src/ecdsa.rs
@@ -96,6 +96,11 @@ impl ecdsa_core::hazmat::DigestPrimitive for Secp256k1 {
     type Digest = sha2::Sha256;
 }
 
+#[cfg(all(feature = "ecdsa", feature = "keccak256"))]
+impl ecdsa_core::hazmat::DigestPrimitive for Secp256k1 {
+    type Digest = sha3::Keccak256;
+}
+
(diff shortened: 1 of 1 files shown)


tarcieri:
I think what you're looking for is `k256::ecdsa::recoverable::Signature`, which provides Ethereum-style recoverable signatures (which carry an extra byte as a "recovery ID"):

https://docs.rs/k256/0.7.1/k256/ecdsa/recoverable/index.html

Note that this signature type already has an impl equivalent to the one you're trying to add here:

https://github.com/RustCrypto/elliptic-curves/blob/80f33e8/k256/src/ecdsa/recoverable.rs#L259-L262

...also note that `DigestSignature` can only be impl'd once for any given signature type as it defines a "default" digest primitive, so the tests for this PR are failing.

If you *really* want to produce a standard (non-recoverable) ECDSA signature with Keccak256 (which I'm guessing isn't the case), the only way to do it is by passing a `Keccak256` instance directly to `DigestSigner`.

TheRealBluesun:
Thank you so much for taking the time to respond! 

You are correct, this was my misunderstanding.  Cheers!

tarcieri:
Opened #270 to hopefully clarify the documentation

PR #276, opened 2021-01-13
Bump `elliptic-curve` dependency to v0.9.0-pre

tarcieri:
...as sourced via git from the RustCrypto/traits repo.

Also bumps `ecdsa` to v0.11.0-pre (via git from RustCrypto/signatures)


tarcieri:
(wrong branch name)

PR #337, opened 2021-05-09
p256/k256: fix mixed coordinate addition with the point at infinity

### LINKED ISSUE #336: Mixed addition is not complete
The mixed addition formulas from the Renes-Costello-Batina paper assume Z2=1, which is not the case for the point at infinity.

As a result, point addition currently returns an incorrect result if the second point is in affine coordinates:

```rust
let g = ProjectivePoint::generator();
let o = AffinePoint::identity();
let p = g + o;
assert_eq!(p, g);
```

-> We get `p != g`.

jedisct1:
The mixed point additions are not complete; they assume that the point in affine coordinates is not the point at infinity.

Handle that case, if only for consistency with non-mixed addition.

Fixes #336

### DIFF
--- a/k256/src/arithmetic/projective.rs
+++ b/k256/src/arithmetic/projective.rs
@@ -186,11 +186,13 @@ impl ProjectivePoint {
             .mul_single(CURVE_EQUATION_B_SINGLE)
             .normalize_weak();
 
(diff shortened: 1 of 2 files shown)


tarcieri:
Thank you!

PR #357, opened 2021-06-12
implement compact support for p256

madninja:
This may not quite be what you had in mind, especially with the modifications to field math to expose the borrow, but it was the shortest way to show what we'd done so far using the new elliptic curve traits

### DIFF
--- a/p256/src/arithmetic/field.rs
+++ b/p256/src/arithmetic/field.rs
@@ -135,7 +135,7 @@ impl FieldElement {
         let is_some = (borrow as u8) & 1;
 
         // Convert w to Montgomery form: w * R^2 * R^-1 mod p = wR mod p
-        CtOption::new(FieldElement(w).mul(&R2), Choice::from(is_some))
+        CtOption::new(FieldElement(w).as_montgomery(), Choice::from(is_some))
     }
 
     /// Returns the SEC1 encoding of this field element.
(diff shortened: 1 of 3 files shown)


p256/src/arithmetic/field.rs
```diff
@@ -201,12 +202,20 @@ impl FieldElement {
...
     pub const fn subtract(&self, rhs: &Self) -> Self {
+        let (_, result) = Self::sub_inner(
+            self.0[0], self.0[1], self.0[2], self.0[3], 0, rhs.0[0], rhs.0[1], rhs.0[2], rhs.0[3],
+            0,
+        );
+        result
+    }
+
+    /// Returns self - rhs mod p
+    pub const fn informed_subtract(&self, rhs: &Self) -> (u64, Self) {
```
tarcieri, line 213: FWIW, our convention for this is generally to make the borrow the second element in the tuple. See `sbb`, for example.

I'm also not sure about the name here, or making it part of the public API.

Hopefully a lot of this code will be cleaner as we transition to `crypto_bigint::UInt`s for the inner type for field elements.

madninja: > FWIW, our convention for this is generally to make the borrow the second element in the tuple. See `sbb`, for example.

Makes sens, fixed

> 
> I'm also not sure about the name here, or making it part of the public API.

Yeah.. agreed, I've limited it to crate only

> 
> Hopefully a lot of this code will be cleaner as we transition to `crypto_bigint::UInt`s for the inner type for field elements.

For sure, yeah. This appears to work in the mean-time and gets of of the wild patch train we had to keep up with

tarcieri (approved):
Thank you!

tarcieri:
The cross-based test failures are not your fault.

I'm going to go ahead and merge this and attempt to address them on `master`.

PR #376, opened 2021-06-27
[WIP] p384: scalar arithmetic

tarcieri:
This is largely written in a way that it could eventually be extracted into an `elliptic_curve::Scalar<C>` type, but for now avoids being generic.

Warning: no tests yet; quite likely to be broken in some way.


tarcieri:
Closed in favor of #489

PR #385, opened 2021-07-19
k256: fix edge case in `Scalar::is_high`

tarcieri:
Per BIP62, the valid range for "low" scalars is:

https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures

> The value S in signatures must be between 0x1 and 0x7FFFFFFF FFFFFFFF
> FFFFFFFF FFFFFFFF 5D576E73 57A4501D DFE92F46 681B20A0 (inclusive).

...where the large hex constant is the scalar modulus / 2.

However, the previous implementation was exclusive. There were tests for this particular case, but the tests confirmed it was exclusive instead of inclusive.

This commit fixes both the tests and the implementation to confirm that the scalar modulus / 2 is *NOT* considered high by `Scalar::is_high`, however scalar modulus / 2 + 1 is high.

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -607,16 +607,20 @@ mod tests {
         let high: bool = Scalar::zero().is_high().into();
         assert!(!high);
 
-        let m = Scalar::modulus_as_biguint();
-        let m_by_2 = &m >> 1;
+        // 1 is not high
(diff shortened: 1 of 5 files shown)


tarcieri:
Note: I'm guessing this was an oversight in my original implementation of low-S normalization. Mea culpa.

PR #386, opened 2021-07-20
k256: add Wycheproof test case which triggers debug assert failure

tarcieri:
This test case is triggering a debug assert in the 32-bit backend only inside of `FieldElement10x26::normalize`.

The 64-bit backend works correctly.

Relevant assertion:

https://github.com/RustCrypto/elliptic-curves/blob/31bb6ac/k256/src/arithmetic/field/field_10x26.rs#L213-L215


tarcieri:
I confirmed that after merging #388 this now passes locally, so I think we should be good to go on #384 now, which includes the tests in this PR.

Going to close this.

PR #388, opened 2021-07-21
Fix a bug in overflow check during 32-bit multiplication

fjarri:
In some situations after two field elements were multiplied (with 32-bit backend), the result was such that:
- in `normalize()`, after `normalize_weak()` the top limb had 23 bits, but was not equal to `0x3FFFFF`
- in `get_overflow()`, a condition for that (`self.0[9] >> 22 != 0`) was missing, so the overflow was not detected
- as a result, the number was not reduced and a debug assertion in `normalize()` was triggered.

See the corresponding check in `libsecp`: https://github.com/bitcoin-core/secp256k1/blob/master/src/field_10x26_impl.h#L67
```c
    x = (t9 >> 22) | ((t9 == 0x03FFFFFUL) & (m == 0x3FFFFFFUL)
        & ((t1 + 0x40UL + ((t0 + 0x3D1UL) >> 26)) > 0x3FFFFFFUL));
```
The first condition was the one missing (but present in the 64-bit version).

Notes: 
- This whole condition is a little suspicious from the constant-timeness point of view, but `libsecp` uses kind of the same one... except in C comparison operators return integers right away, and I don't really know how Rust handles it. Maybe it ends up the same.


tarcieri (approved):

tarcieri:
Thank you!

PR #396, opened 2021-08-24
k256: use `crypto-bigint` for `scalar::MODULUS`

tarcieri:
This commit begins the process of starting to retrofit representing the scalar modulus as a `crypto_bigint::UInt`.

It isn't a perfect fit in the current form as the `force-32-bit` cargo feature complicates things. It would probably make sense to get rid of it in the next breaking release, especially as it isn't additive. Instead it should be easy to do 32-bit testing with `cross`.

All that said, it's now possible to calculate constants like `FRAC_MODULUS_2` at compile time using `const fn` which is pretty neat.

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -5,18 +5,22 @@ use cfg_if::cfg_if;
 cfg_if! {
     if #[cfg(any(target_pointer_width = "32", feature = "force-32-bit"))] {
         mod scalar_8x32;
-        use scalar_8x32::Scalar8x32 as ScalarImpl;
-        use scalar_8x32::WideScalar16x32 as WideScalarImpl;
+        use scalar_8x32::{
+            Scalar8x32 as ScalarImpl,
+            WideScalar16x32 as WideScalarImpl,
+        };
 
         #[cfg(feature = "bits")]
-        use scalar_8x32::MODULUS;
+        use scalar_8x32::MODULUS as MODULUS_ARR;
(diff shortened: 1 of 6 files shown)


tarcieri:
FYI @fjarri, some complications due to the `force-32-bit` feature. I'd suggest getting rid of it and using `cross` to test 32-bit targets.

k256/src/arithmetic/scalar.rs
```diff
@@ -5,24 +5,29 @@ use cfg_if::cfg_if;
...
-        use scalar_4x64::Scalar4x64 as ScalarImpl;
-        use scalar_4x64::WideScalar8x64 as WideScalarImpl;
+        use scalar_4x64::{
+            Scalar4x64 as ScalarImpl,
+            WideScalar8x64 as WideScalarImpl,
+        };
 
         #[cfg(feature = "bits")]
-        use scalar_4x64::MODULUS;
+        use scalar_4x64::MODULUS as MODULUS_ARR;
```
tarcieri, line 23: This can go away if we get rid of the `force-32-bit` feature.

k256/src/arithmetic/scalar.rs
```diff
@@ -172,7 +181,7 @@ impl PrimeFieldBits for Scalar {
     }
 
     fn char_le_bits() -> ScalarBits {
-        MODULUS.into()
+        MODULUS_ARR.into()
```
tarcieri, line 184: Ideally this would be `MODULUS.to_uint_array().into()` but it can't be yet because of the `force-32-bit` feature.

PR #399, opened 2021-08-30
k256: remove `force-32-bit` feature

tarcieri:
This feature was originally added for local testing purposes, but greatly complicates interactions with additional crates like
`crypto-bigint` which are gated entirely on the target's word size.

The testing workflow this provided can still be accomplished using a tool like `cross` by testing against a 32-bit target:

```
$ cross test --target powerpc-unknown-linux-gnu --release --all-features
```

Bonus points for testing against a big endian target to ensure there aren't any endian-related bugs as well.

Breaking change in that it removes a feature.


.gitignore
```diff
@@ -1,3 +1,3 @@
 target
-tests/Cargo.lock
+**/Cargo.lock
```
tarcieri, line 2: One annoying thing about `cross` is it leaves behind a `Cargo.lock` file in the crate's directory (and for that matter, doesn't use the existing crate resolution) but this change at least prevents them from being checked in.

PR #400, opened 2021-08-31
k256: implement `Scalar::sqrt`

tarcieri:
Implements Tonelli-Shank's algorithm for q mod 16 = 1 as synthesized using the `ff_derive` crate, similar to #392 which implements it for the `p256`.

Like in `p256`, as part of implementing this it was discovered that `root_of_unity()` was incorrect. Here it is (re)calculated with sage:

```sage
sage: n = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141
sage: GF(n).primitive_element()
7
sage: s = 6
sage: t = (n - 1) >> s
sage: power_mod(7,t,n)
5480320495727936603795231718619559942670027629901634955707709633242980176626
```

Note that the value was computed correctly originally, but the hex digits were shifted such that the resulting value was left shifted by 4-bits. This has now been corrected.

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -113,9 +113,372 @@ impl Field for Scalar {
         Scalar::invert(self)
     }
 
-    // TODO(tarcieri): stub! See: https://github.com/RustCrypto/elliptic-curves/issues/170
+    /// Tonelli-Shank's algorithm for q mod 16 = 1
+    /// https://eprint.iacr.org/2012/685.pdf (page 12, algorithm 5)
+    #[allow(clippy::many_single_char_names)]
     fn sqrt(&self) -> CtOption<Self> {
-        todo!("see RustCrypto/elliptic-curves#170");
+        // TODO(tarcieri): replace with `self.pow((t - 1) >> 1)`
+        let w = {
+            let t0 = self;
+            let t1 = t0.square();
+            let t2 = t1 * t0;
+            let t3 = t1.square();
+            let t4 = t3.square();
+            let t5 = t4 * t2;
+            let t6 = t5 * t3;
+            let t7 = t6.square();
+            let t8 = t7 * t6;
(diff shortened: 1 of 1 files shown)


k256/src/arithmetic/scalar.rs
```diff
@@ -113,9 +113,371 @@ impl Field for Scalar {
...
+    /// https://eprint.iacr.org/2012/685.pdf (page 12, algorithm 5)
+    #[allow(clippy::many_single_char_names)]
     fn sqrt(&self) -> CtOption<Self> {
-        todo!("see RustCrypto/elliptic-curves#170");
+        let w = {
+            let t0 = self;
+            let t1 = t0.square();
+            let t2 = t1 * t0;
+            let t3 = t1.square();
+            let t4 = t3.square();
```
tarcieri, line 125: Curiously I wasn't able to replace this with `pow_vartime` as copied/pasted from `p256`, which is how `sqrt` is implemented there.

I tried the following:

```
        let w = self.pow_vartime(&[
            0x77fa4bd19a06c82,
            0xd755db9cd5e91407,
            0xffffffffffffffff,
            0x1fffffffffffffff,
        ]);
```

It's possible I miscomputed the exponent. Here's my Sage:

```sage
sage: n = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141
sage: n
115792089237316195423570985008687907852837564279074904382605163141518161494337
sage: s = 6
sage: t = (n - 1) >> s
sage: t
1809251394333065553493296640760748560200586941860545380978205674086221273349
sage: (t - 1) / 2
904625697166532776746648320380374280100293470930272690489102837043110636674
sage: hex((t - 1) / 2)
'0x1fffffffffffffffffffffffffffffffd755db9cd5e9140777fa4bd19a06c82'
```

PR #408, opened 2021-09-07
p256: use `U256` as the internal representation of `Scalar`

tarcieri:
This is a minimally invasive change to use `crypto_bigint::U256` as the internal representation of `p256::Scalar`.

There are several places that various functions are still using a `U64x4` type which could be made to work over 8 x 32-bit limbs instead on 32-bit platforms, which is what `U256` uses internally.

This commit doesn't attempt to implement those changes, but rather just switch out the type used for the internal representation. This should hopefully avoid any regressions in this particular change.

### DIFF
--- a/p256/src/arithmetic/projective.rs
+++ b/p256/src/arithmetic/projective.rs
@@ -284,7 +284,7 @@ impl ProjectivePoint {
     fn mul(&self, k: &Scalar) -> ProjectivePoint {
         let mut ret = ProjectivePoint::identity();
 
-        for limb in k.0.iter().rev() {
+        for limb in k.to_u64x4().iter().rev() {
             for i in (0..64).rev() {
                 ret = ret.double();
                 ret.conditional_assign(&(ret + self), Choice::from(((limb >> i) & 1u64) as u8));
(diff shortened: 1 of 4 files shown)


p256/src/arithmetic/projective.rs
```diff
@@ -284,7 +284,7 @@ impl ProjectivePoint {
     fn mul(&self, k: &Scalar) -> ProjectivePoint {
         let mut ret = ProjectivePoint::identity();
 
-        for limb in k.0.iter().rev() {
+        for limb in k.to_u64x4().iter().rev() {
             for i in (0..64).rev() {
```
tarcieri, line 288: This would be one of the easiest things to switch to using 32-bit limbs on 32-bit targets, but I left it as close to the original implementation as possible for now.

p256/src/arithmetic/scalar.rs
```diff
@@ -366,21 +326,20 @@ impl Scalar {
...
-        ret[0..8].copy_from_slice(&self.0[3].to_be_bytes());
-        ret[8..16].copy_from_slice(&self.0[2].to_be_bytes());
-        ret[16..24].copy_from_slice(&self.0[1].to_be_bytes());
-        ret[24..32].copy_from_slice(&self.0[0].to_be_bytes());
-        ret
+        self.0.to_be_byte_array()
     }
 
     /// Returns self + rhs mod n
+    // TODO(tarcieri): use `UInt::add_mod`
```
tarcieri, line 333: @dignifiedquire originally as part of this PR I did attempt to use `add_mod` and ran into some test failures. I'll open a draft followup with that change after this gets merged

PR #426, opened 2021-09-12
k256+p256: use `UInt::add_mod` for scalar addition

tarcieri:
`crypto-bigint` v0.2.7 fixed a bug in `UInt::add_mod` in https://github.com/RustCrypto/utils/pull/619, and the `elliptic-curve` crate has been updated to depend on this version at a minimum.

This means we can now use it to implement scalar addition.

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -242,10 +242,7 @@ impl Scalar {
     /// Returns self + rhs mod n.
     // TODO(tarcieri): use `UInt::add_mod`
     pub const fn add(&self, rhs: &Self) -> Self {
-        let (w, carry) = self.0.adc(&rhs.0, Limb::ZERO);
-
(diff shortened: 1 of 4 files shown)


k256/src/arithmetic/scalar.rs
```diff
@@ -242,10 +242,7 @@ impl Scalar {
     /// Returns self + rhs mod n.
     // TODO(tarcieri): use `UInt::add_mod`
```
BlackHoleFox, line 243: I was a few seconds too slow from Zulip to point out this TODO has also been addressed.

tarcieri: Aah yeah, got it on `p256` but not `k256`. Thanks for pointing it out.

tarcieri: Removed in 8ddcc1c

PR #431, opened 2021-09-13
k256: remove `Scalar::conditional_add_bit`

tarcieri:
The case it's actually being used for in practice is much simpler and can be easily expressed using the current functionality.


fjarri:
Just as a general information: in `libsecp256k1`, the analogue of this function (`secp256k1_scalar_cadd_bit()`) is only ever used with a non-zero shift in `secp256k1_wnaf_const()`.

fjarri (approved):

PR #432, opened 2021-09-14
k256: Add non-biased/non-zero constructors to Scalar

fjarri:
This is a "spitballing" PR, intended primarily for illustration of required API and discussion.

There are two things I'm missing from the current `Scalar` API:
- an ability to generate random `NonZeroScalar`s (without unwrapping)
- an ability to generate a `Scalar`/`NonZeroScalar` from a digest safely according to [hash-to-curve standard](https://tools.ietf.org/pdf/draft-irtf-cfrg-hash-to-curve-11.pdf)

The latter needs the scalar to be reduced from `L = ceil((ceil(log2(p)) + k) / 8)` bytes (Section 5.3), where `p` is the order, and `k` is the security parameter. For `k = 256` this gives 64 bytes, for `k = 128` it's 48 bytes. I went with 64 for simplicity.

The public methods added to `Scalar`:
- `pub fn random_nonzero(rng: impl RngCore) -> NonZeroScalar`
- `pub fn from_digest_safe<D>(digest: D) -> Self where D: Digest<OutputSize = U64>`
- `pub fn from_digest_safe_nonzero<D>(digest: D) -> NonZeroScalar where D: Digest<OutputSize = U64>`

Now for the problems:
- `random()` is a part of the `Field` trait. If `NonZeroScalar` implemented `Field`, the code from `random_nonzero()` could go there.
- `FromDigest` requires `OutputSize = FieldSize`, so it can't be changed to 64 bytes output. The ways to deal with it include:
  - adding a `SafeExpansionSize` (or whatever) type and locking `FromDigest` to it
  - adding a `FromDigestSafe` trait
- `from_digest_safe_nonzero()` can be an impl of `FromDigest` for `NonZeroScalar` (but again there's the size problem)
- The added internal function `Scalar::from_wide_bytes_reduced()` assumes that `WideScalar::reduce()` can reduce any 512 bit, and not just anything below `order^2`. I'm 99.9% certain that's true, but I can't present a formal proof right now. I did test it for `0xfff...fff`, and it works.
- `WideScalar::reduce_nonzero()` assumes that `reduce()` works just as well for the modulus decreased by one. See the comment above about being 99.9% sure.
- `WideScalar::reduce_nonzero()` contains an `unwrap()`. Ideally I'd prefer to have something like `NonZeroScalar::from_bytes_unchecked()` to avoid it.

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -6,7 +6,7 @@ mod wide;
 
 pub(crate) use self::wide::WideScalar;
 
-use crate::{FieldBytes, Secp256k1, ORDER};
+use crate::{FieldBytes, NonZeroScalar, Secp256k1, ORDER};
 use core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Shr, Sub, SubAssign};
 use elliptic_curve::{
     bigint::{limb, nlimbs, ArrayEncoding, Encoding, Limb, U256},
@@ -25,7 +25,11 @@ use elliptic_curve::{
 use {crate::ScalarBits, elliptic_curve::group::ff::PrimeFieldBits};
 
 #[cfg(feature = "digest")]
-use ecdsa_core::{elliptic_curve::consts::U32, hazmat::FromDigest, signature::digest::Digest};
+use ecdsa_core::{
+    elliptic_curve::consts::{U32, U64},
+    hazmat::FromDigest,
+    signature::digest::Digest,
+};
 
 #[cfg(test)]
 use num_bigint::{BigUint, ToBigUint};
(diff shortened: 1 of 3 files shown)


tarcieri:
Lots to respond to here, but I can perhaps start here:

> FromDigest requires OutputSize = FieldSize, so it can't be changed to 64 bytes output. The ways to deal with it include:
> - adding a SafeExpansionSize (or whatever) type and locking FromDigest to it
> - adding a FromDigestSafe trait

The current `FromDigest` trait is `ecdsa::FromDigest` and very much serves ECDSA-related concerns. I specifically moved it to the `ecdsa` crate to get it out-of-the-way so we can define a new `FromDigest` crate in `elliptic-curve`.

It'd be great if we could define a better `elliptic_curve::FromDigest` trait to replace `ecdsa::FromDigest`, if we can design it in such a way that it can cover both ECDSA's usage and hash-to-curve.

Perhaps we could make the input size a generic parameter, which would permit impls for multiple sizes.

tarcieri:
I looked at trying to add a new `FromDigest` trait to the `elliptic-curve` crate which is generic with respect to the digest size and can be impl'd for multiple sizes, and then it dawned on me that the result really had nothing specifically to do with elliptic curves:

```rust
pub trait FromDigest<OutputSize: ArrayLength<u8>> {
    fn from_digest<D>(digest: D) -> Self
    where
        D: Digest<OutputSize = OutputSize>;
}
```

Given that, perhaps the trait could be added to the `digest` crate instead /cc @newpavlov

newpavlov:
Shouldn't this use-case be already covered by the [`crypto_common::InnerInit`](https://github.com/RustCrypto/traits/blob/new_traits/crypto-common/src/init.rs#L53-L59) trait? You can write a blanket impl for scalars and hashes with matching sizes

tarcieri:
In this particular case, we would need to provide several impls for different sizes of a digest, e.g. for a secp256k1 scalar it could be initialized from a digest with a 256-bit output using a narrow reduction, or a 512-bit output using a wide reduction, and those two cases will take separate codepaths

newpavlov:
Is this initialization process defined for a range of output sizes or only for those two values (e.g. would it work for SHA-1 or for SHA-384)? Do you track which hash function was used on the type level (i.e. do you get the same type when 256-bit and 512-bit  hash functions were used)?

tarcieri:
I can't speak specifically to the hash-to-curve-point case, but in the case of scalars it's effectively two values:

- narrow: bitlength equivalent to the modulus
- wide: 2 * bitlength equivalent to the modulus

>  Do you track which hash function was used on the type level

The `ecdsa` crate needs to do this, namely the default implementation uses the digest function used on the message to instantiate HMAC for RFC6979

newpavlov:
I am not sure why `InnerInit` can not be used here:
```rust
struct Foo<D>
where D: Digest, // also constrain D::OutputSize % Self::ScalarSize == 0
{ .. }

impl<D> InnerUser for Foo<D>
where // same bounds
{ type Inner = D; }

impl<D> InnerInit for Foo<D>
where // same bounds
{
    fn inner_init(digest: D) -> Self { .. }
}
```
Am I missing something?

tarcieri:
Can you write a full translation of something like this?

```
pub trait FromDigest<OutputSize: ArrayLength<u8>> {
    fn from_digest<D>(digest: D) -> Self
    where
        D: Digest<OutputSize = OutputSize>;
}

pub struct Scalar([u64; 4]);

impl Scalar {
    pub fn from_bytes_mod_order(bytes: &[u8; 32]) -> Self {
        ...
    }

    pub fn from_bytes_mod_order_wide(bytes: &[u8; 64]) -> Self {
        ...
    }
}

impl FromDigest<U32> for Scalar {
    fn from_digest<D>(digest: D) -> Self
    where
        D: Digest<OutputSize = U32>
    {
        Self::from_bytes_mod_order(&digest.finalize().into())
    }
}

impl FromDigest<U64> for Scalar {
    fn from_digest<D>(digest: D) -> Self
    where
        D: Digest<OutputSize = U64>
    {
        Self::from_bytes_mod_order_wide(&digest.finalize().into())
    }
}
```

The real issue is the overlapping impls, which you can't have without a generic parameter.

newpavlov:
Here is my take on it: https://play.rust-lang.org/?gist=6547bf4c636d51196326fd26aa4bbdee

The idea is that the trait should be implemented by a type which tracks used hash function. It's not necessary should be the scalar itself, it could be wrapper around it as well. The trait bounds are somewhat annoying, but it should be better eventually with panicking `const fn`s in where bounds.

tarcieri:
It might be good to split this off into a separate issue, possibly https://github.com/RustCrypto/traits/issues/481

But I don't get how that works @newpavlov, especially this:

>  D::OutputSize: IsEqual<U32> + IsEqual<U64>,

But then the problem further becomes, in the context of ECDSA I need to bound on support for a specific digest size:

```
pub struct SigningKey<C>
where
    C: PrimeCurve + ProjectiveArithmetic,
    Scalar<C>: FromDigest<FieldSize<C>> + Invert<Output = Scalar<C>> + SignPrimitive<C>,
    SignatureSize<C>: ArrayLength<u8>,
{
    inner: NonZeroScalar<C>,
}
```

newpavlov:
>It might be good to split this off into a separate issue

Yeah, I've answered to you here: https://github.com/RustCrypto/traits/issues/481#issuecomment-921053716

tarcieri:
Sidestepping the issue of digests for the moment, I've opened a PR which adds a `Reduce` trait for these sorts of hashing-to-a-field-element constructors:

https://github.com/RustCrypto/traits/pull/768

It's generic around a `crypto_bignum::Integer`, allowing overlapping impls for multiple integer sizes, so it can be impl'd twice for a "narrow" and "wide" reduction.

There are also provided methods for reducing from a big endian or little endian byte array.

The bounds for the ECDSA use case would look something like:

```rust
Scalar<C>: Reduce<C::Uint>
```

...to select a narrow reduction.

`NonZeroScalar` is defined in the `elliptic-curve` crate. We can add impls for it there, but they would have to be generic over curves, or use an additional "thunk" trait.

tarcieri:
#436 adds "narrow" impls of the `Reduce` trait to `k256` and `p256`

tarcieri:
Circling back on this, I think this is the current state of things:

#### an ability to generate random NonZeroScalars (without unwrapping)
Use `NonZeroScalar::random`

#### an ability to generate a Scalar/NonZeroScalar from a digest safely according to hash-to-curve standard
#472 impl'd `Reduce<U512>` for `Scalar` which has a `Reduce::from_be_bytes` method accepting a 64-bit input

A couple things are still missing though:
- Integration with `digest` (possibly via a `digest` feature on the `elliptic-curve` crate which adds methods to `Reduce`)
- Support for `NonZeroScalar`

It seems like supporting this might require changes to the `Reduce` trait, or possibly supplementing it with another trait (e.g. `ReduceNonZero`), in order to propagate the `modulus_minus_one` flag passed to the reduction.

If that existed, then it would be possible to impl `Reduce` for `NonZeroScalar` where the inner `Scalar` type impls the `Reduce`-with-modulus-minus-one flag.

tarcieri:
Proposed traits for non-zero reductions here: https://github.com/RustCrypto/traits/pull/827

tarcieri:
Went ahead and addressed what I believe are the remaining concerns, adapting this PR into #474, which is now merged.

PR #433, opened 2021-09-16
Normalize p256

x87-va:


p256/Cargo.toml
```diff
@@ -29,6 +29,8 @@ ecdsa-core = { version = "=0.13.0-pre", package = "ecdsa", default-features = fa
 hex-literal = "0.3"
 proptest = "1.0"
 rand_core = { version = "0.6", features = ["getrandom"] }
+num-bigint = "0.4"
+num-traits = "0.2"
```
tarcieri, line 33: I think it's a bit excessive to loop in `num-bigint` just for testing this. `k256` does a number of other equivalence tests of the field arithmetic with it.

I think ideally we'd even get rid of those and move all of that sort of testing into a `proptests` crate or something.

For the purposes of testing this, I think it's fine to use `crypto_bigint::UInt` to compute the various boundary conditions and then check if they're low/high.

p256/src/ecdsa.rs
```diff
@@ -128,6 +129,16 @@ impl VerifyPrimitive<NistP256> for AffinePoint {
     }
 }
 
+impl NormalizeLow for Scalar {
```
tarcieri, line 132:
```suggestion
#[cfg(feature = "ecdsa")]
impl NormalizeLow for Scalar {
```

tarcieri:
I believe this is obsoleted by https://github.com/RustCrypto/signatures/pull/393

PR #443, opened 2021-09-30
Fix `to_bytes()` for `ProjectivePoint::identity()`

### LINKED ISSUE #442: `<ProjectivePoint as GroupEncoding>::to_bytes()` can panic
For example when you use `ProjectivePoint::identity()`, the size that `to_bytes()` will attempt to convert with `clone_from_slice` will only be 1, but it expects 33.

daxpedda:
This could maybe alternatively be fixed "upstream" in `EncodedPoint::len()`?
Fixes #442.

### DIFF
--- a/p256/src/arithmetic/projective.rs
+++ b/p256/src/arithmetic/projective.rs
@@ -75,7 +75,11 @@ impl GroupEncoding for ProjectivePoint {
     }
 
     fn to_bytes(&self) -> Self::Repr {
-        CompressedPoint::clone_from_slice(self.to_affine().to_encoded_point(true).as_bytes())
+        let bytes = self.to_affine().to_encoded_point(true);
+        let bytes = bytes.as_bytes();
+        let mut result = CompressedPoint::default();
+        result[..bytes.len()].copy_from_slice(bytes);
+        result
     }
 }
 
@@ -521,7 +525,7 @@ impl<'a> Neg for &'a ProjectivePoint {
 mod tests {
     use super::{AffinePoint, ProjectivePoint, Scalar};
     use crate::test_vectors::group::{ADD_TEST_VECTORS, MUL_TEST_VECTORS};
-    use elliptic_curve::group::{ff::PrimeField, prime::PrimeCurveAffine};
+    use elliptic_curve::group::{ff::PrimeField, prime::PrimeCurveAffine, GroupEncoding};
 
     #[test]
     fn affine_to_projective() {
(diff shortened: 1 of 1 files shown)


tarcieri:
This is an unfortunate mismatch between the `GroupEncoding` API and SEC1 encoding.

> This could maybe alternatively be fixed "upstream" in EncodedPoint::len()?

`EncodedPoint::len` is doing the right thing: the SEC1 encoding for the identity point is a 1-byte `0` tag.

Technically this is an invalid SEC1 encoding, but there's nothing else that can be done with a fixed-sized `GroupEncoding::Repr`, and it's certainly preferable to panicking.

p256/src/arithmetic/projective.rs
```diff
@@ -665,4 +669,9 @@ mod tests {
             assert_eq!(res.y.to_bytes(), coords.1.into());
         }
     }
+
+    #[test]
+    fn projective_to_bytes() {
+        assert_eq!([0; 33], ProjectivePoint::identity().to_bytes().as_slice());
+    }
```
tarcieri, line 676: It would be good to make a note that this test is specifically for the identity point, and that the result is technically/deliberately an invalid SEC1 encoding.

daxpedda:
> Technically this is an invalid SEC1 encoding, but there's nothing else that can be done with a fixed-sized `GroupEncoding::Repr`, and it's certainly preferable to panicking.

We could change the interface to return an `Option`?

tarcieri:
`GroupEncoding` is defined in the `group` crate, so the change would have to be made there.

Really it's an eccentricity of the SEC1 encoding, and I doubt they would want to make the API fallible just to support that.

daxpedda:
> `GroupEncoding` is defined in the `group` crate, so the change would have to be made there.

Ah, I wasn't aware this isn't owned by RustCrypto.

p256/src/arithmetic/projective.rs
daxpedda: I hope this is enough:
https://github.com/RustCrypto/elliptic-curves/blob/0ba5440254df0928c69b80cf6e74182931182141/p256/src/arithmetic/projective.rs#L674-L675

tarcieri:
It'd also be good to make sure that `GroupEncoding::from_bytes` can decode this encoding, and also that the problem is also solved for the `GroupEncoding` impl on `AffinePoint`.

tarcieri:
This is fine as a start. I'll take care of the other comments I raised.

PR #444, opened 2021-10-01
Handle identity point in `GroupEncoding`

daxpedda:
This enabled `GroupEncoding::to_bytes()` and `GroupEncoding::from_bytes()` for `AffinePoint` and `ProjectivePoint` for p256 and k256 to handle the identity point.

See #443.

### DIFF
--- a/k256/src/arithmetic/affine.rs
+++ b/k256/src/arithmetic/affine.rs
@@ -171,6 +171,12 @@ impl GroupEncoding for AffinePoint {
         let is_compressed_point = y_is_odd | tag.ct_eq(&sec1::Tag::CompressedEvenY.into());
         Self::decompress(FieldBytes::from_slice(&bytes[1..]), y_is_odd)
             .and_then(|point| CtOption::new(point, is_compressed_point))
+            .or_else(|| {
+                CtOption::new(
+                    AffinePoint::identity(),
+                    tag.ct_eq(&sec1::Tag::Identity.into()),
+                )
+            })
     }
 
     fn from_bytes_unchecked(bytes: &Self::Repr) -> CtOption<Self> {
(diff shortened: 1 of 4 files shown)


k256/src/arithmetic/affine.rs
```diff
@@ -171,6 +171,12 @@ impl GroupEncoding for AffinePoint {
         let is_compressed_point = y_is_odd | tag.ct_eq(&sec1::Tag::CompressedEvenY.into());
         Self::decompress(FieldBytes::from_slice(&bytes[1..]), y_is_odd)
             .and_then(|point| CtOption::new(point, is_compressed_point))
+            .or_else(|| {
+                CtOption::new(
+                    AffinePoint::identity(),
+                    tag.ct_eq(&sec1::Tag::Identity.into()),
+                )
+            })
```
daxpedda, line 179: This seemed to me to be the cleanest solution, but not entirely correct because it doesn't check if the rest of the bytes are also `0`s.

I think the best approach if this is desired is to change `DecompressPoint::decompress()` to handle the identity point, which would require a breaking change.

tarcieri: This would accept a `0`-tagged value with garbage bytes. Not sure that's a good idea. It should probably check all of them.

This is a lot of logic to repeat on a per-crate basis. It seems like it might be worth hoisting into the `elliptic-curve` crate in some form. It could probably be composed in terms of `FromDecodedPoint`.

daxpedda: I'm unsure what you mean, `FromDecodedPoint::from_encoded_point()` requires an `EncodedPoint` and doesn't do the check we want. Do you mean adding a new method to `FromDecodedPoint`?

tarcieri: No, I mean something which uses `FromDecodedPoint` to do the decoding, possibly after changing its result to a `CtOption`, and offloading the SEC1 tag handling there

daxpedda: I just realized, I thought you meant `FromEncodedPoint`, I can't find `FromDecodedPoint`.

tarcieri: Err yes sorry (responding on my phone here)

tarcieri:
I've opened a PR with a slight variation on this:

#446

It's able to leverage a now `CtOption` return value from the `FromEncodedPoint` trait introduced in #445.

tarcieri:
Closing in favor of #446

PR #446, opened 2021-10-02
k256+p256: handle identity point in `GroupEncoding`

tarcieri:
`GroupEncoding` uses a fixed-width `Repr` whereas SEC1 is variable-width with respect to compressed points versus the identity.

This change allows 33-bytes of zeroes to be used as the identity. Technically that's not a valid SEC1 encoding: the SEC1 encoding for the identity is 1-byte: 0x00. However, this is the best we can do with a fixed-width encoding.

See also: #443 and #444

### DIFF
--- a/k256/src/arithmetic/affine.rs
+++ b/k256/src/arithmetic/affine.rs
@@ -132,11 +132,15 @@ impl GroupEncoding for AffinePoint {
     type Repr = CompressedPoint;
 
     fn from_bytes(bytes: &Self::Repr) -> CtOption<Self> {
-        let tag = bytes[0];
-        let y_is_odd = tag.ct_eq(&sec1::Tag::CompressedOddY.into());
-        let is_compressed_point = y_is_odd | tag.ct_eq(&sec1::Tag::CompressedEvenY.into());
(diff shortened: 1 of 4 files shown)


k256/src/arithmetic/affine.rs
```diff
@@ -132,11 +132,15 @@ impl GroupEncoding for AffinePoint {
...
-        Self::decompress(FieldBytes::from_slice(&bytes[1..]), y_is_odd)
-            .and_then(|point| CtOption::new(point, is_compressed_point))
+        EncodedPoint::from_bytes(bytes)
+            .map(|point| CtOption::new(point, Choice::from(1)))
+            .unwrap_or_else(|_| {
+                // SEC1 identity encoding is technically 1-byte 0x00, but the
+                // `GroupEncoding` API requires a fixed-width `Repr`
+                let is_identity = bytes.ct_eq(&Self::Repr::default());
+                CtOption::new(EncodedPoint::identity(), is_identity)
+            })
```
tarcieri, line 142: Technically this is a branch, but it's one which only occurs with respect to the SEC1 tag.

The important part to ensure is actually constant-time is the point decoding, which happens regardless of which branch is taken through the initial SEC1 decoding.

PR #449, opened 2021-10-09
k256: impl `Drop` for `ecdsa::SigningKey`

tarcieri:
Adds a `Drop` handler which calls `zeroize()` on the `NonZeroScalar` which the `SigningKey` newtype wraps.


tarcieri:
It's to avoid people accidentally using the `SigningKey` after it's been zeroized.

There's a bit of a tricky tradeoff here which we're aware causes some ergonomic problems. See https://github.com/iqlusioninc/crates/issues/757

It would probably be good if the custom derive support could allow skipping of individual fields which are known to be self-zeroizing. Something like `#[zeroize(skip)]`

PR #482, opened 2021-12-13
[WIP] k256: Taproot Schnorr (BIP 340)

tarcieri:
Implements Taproot-flavored Schnorr signatures as described in:
https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki

Work-In-Progress implementation: BROKEN, DO NOT USE

### DIFF
--- a/k256/src/arithmetic/affine.rs
+++ b/k256/src/arithmetic/affine.rs
@@ -9,7 +9,7 @@ use elliptic_curve::{
     sec1::{self, FromEncodedPoint, ToEncodedPoint},
     subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption},
     zeroize::DefaultIsZeroes,
-    AffineArithmetic, AffineXCoordinate, DecompressPoint, Error, Result,
+    AffineArithmetic, AffineXCoordinate, DecompactPoint, DecompressPoint, Error, Result,
 };
 
 impl AffineArithmetic for Secp256k1 {
@@ -41,6 +41,17 @@ pub struct AffinePoint {
     pub(super) infinity: Choice,
 }
 
+impl AffinePoint {
+    /// Create a new [`AffinePoint`] with the given coordinates.
+    pub(crate) fn new(x: FieldElement, y: FieldElement) -> Self {
+        Self {
+            x,
+            y,
+            infinity: Choice::from(0),
+        }
+    }
+}
+
 impl PrimeCurveAffine for AffinePoint {
     type Scalar = Scalar;
     type Curve = ProjectivePoint;
@@ -59,21 +70,20 @@ impl PrimeCurveAffine for AffinePoint {
         // SECP256k1 basepoint in affine coordinates:
         // x = 79be667e f9dcbbac 55a06295 ce870b07 029bfcdb 2dce28d9 59f2815b 16f81798
         // y = 483ada77 26a3c465 5da4fbfc 0e1108a8 fd17b448 a6855419 9c47d08f fb10d4b8
-        AffinePoint {
-            x: FieldElement::from_bytes(&arr![u8;
+        Self::new(
+            FieldElement::from_bytes(&arr![u8;
                 0x79, 0xbe, 0x66, 0x7e, 0xf9, 0xdc, 0xbb, 0xac, 0x55, 0xa0, 0x62, 0x95, 0xce, 0x87,
                 0x0b, 0x07, 0x02, 0x9b, 0xfc, 0xdb, 0x2d, 0xce, 0x28, 0xd9, 0x59, 0xf2, 0x81, 0x5b,
                 0x16, 0xf8, 0x17, 0x98
             ])
             .unwrap(),
-            y: FieldElement::from_bytes(&arr![u8;
+            FieldElement::from_bytes(&arr![u8;
                 0x48, 0x3a, 0xda, 0x77, 0x26, 0xa3, 0xc4, 0x65, 0x5d, 0xa4, 0xfb, 0xfc, 0x0e, 0x11,
                 0x08, 0xa8, 0xfd, 0x17, 0xb4, 0x48, 0xa6, 0x85, 0x54, 0x19, 0x9c, 0x47, 0xd0, 0x8f,
                 0xfb, 0x10, 0xd4, 0xb8
             ])
             .unwrap(),
-            infinity: Choice::from(0),
-        }
+        )
     }
 
     /// Is this point the identity point?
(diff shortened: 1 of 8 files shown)


k256/src/arithmetic/affine.rs
```diff
@@ -140,16 +150,22 @@ impl DecompressPoint<Secp256k1> for AffinePoint {
...
         })
     }
 }
 
+/// Decompaction using Taproot conventions as described in [BIP 340].
+///
+/// [BIP 340]: https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki
+impl DecompactPoint<Secp256k1> for AffinePoint {
+    fn decompact(x_bytes: &FieldBytes) -> CtOption<Self> {
+        // TODO(tarcieri): implement full `lift_x` algorithm as described in BIP 340
```
tarcieri, line 164: BIP 340 defines the `lift_x` function as follows:

> The function ''lift_x(x)'', where ''x'' is an integer in range ''0..p-1'', returns the point ''P'' for which ''x(P) = x''<ref>
>    Given a candidate X coordinate ''x'' in the range ''0..p-1'', there exist either exactly two or exactly zero valid Y coordinates. If no valid Y coordinate exists, then ''x'' is not a valid X coordinate either, i.e., no point ''P'' exists for which ''x(P) = x''. The valid Y coordinates for a given candidate ''x'' are the square roots of ''c = x<sup>3</sup> + 7 mod p'' and they can be computed as ''y = &plusmn;c<sup>(p+1)/4</sup> mod p'' (see [https://en.wikipedia.org/wiki/Quadratic_residue#Prime_or_prime_power_modulus Quadratic residue]) if they exist, which can be checked by squaring and comparing with ''c''.</ref> and ''has_even_y(P)'', or fails if no such point exists. The function ''lift_x(x)'' is equivalent to the following pseudocode:
> * Let c = x<sup>3</sup> + 7 mod p.
> * Let y = c<sup>(p+1)/4</sup> mod p.
> * Fail if c &ne; y<sup>2</sup> mod p.
> * Return the unique point P such that x(P) = x and y(P) = y if y mod 2 = 0 or y(P) = p-y otherwise.

Unfortunately I'm unable to compute *y = c<sup>(p+1)/4</sup> mod p* as the only support we have for raising a `FieldElement` to a power is `FieldElement::pow2k` which requires the exponent be a power of 2.

cc @fjarri

k256/src/schnorr.rs
```diff
@@ -0,0 +1,318 @@
...
+    /// Parse signing key from big endian-encoded bytes.
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self> {
+        let secret_key = NonZeroScalar::try_from(bytes).map_err(|_| Error::new())?;
+        let public_key =
+            PublicKey::from_affine((ProjectivePoint::generator() * *secret_key).to_affine())
+                .map_err(|_| Error::new())?;
+
+        Ok(Self {
+            secret_key,
+            verifying_key: VerifyingKey { inner: public_key },
```
tarcieri, line 77: I forgot to add a TODO in the code, but this should be constructed using `DecompactPoint::decompact` once it properly implements the `lift_x` function.

k256/src/schnorr.rs
```diff
@@ -0,0 +1,318 @@
...
+        // SignVector {
+        //     secret_key: hex!("0B432B2677937381AEF05BB02A66ECD012773062CF3FA2549E44F58ED2401710"),
+        //     public_key: hex!("25D1DFF95105F5253C4022F628A996AD3A0D95FBF21D468A1B33F8C160D8F517"),
+        //     aux_rand: hex!("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"),
+        //     message: hex!("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"),
+        //     signature: hex!(
+        //         "7EB0509757E246F19449885651611CB965ECC1A187DD51B64FDA1EDC9637D5EC
+        //          97582B9CB13DB3933705B32BA982AF5AF25FD78881EBB32771FC5922EFC66EA3"
+        //     ),
+        // },
```
tarcieri, line 302: Though the other test vectors work, this one computes an incorrect signature:

```
thread 'schnorr::tests::bip340_sign_vectors' panicked at 'assertion failed: `(left == right)`
  left: `[126, 176, 80, 151, 87, 226, 70, 241, 148, 73, 136, 86, 81, 97, 28, 185, 101, 236, 193, 161, 135, 221, 81, 182, 79, 218, 30, 220, 150, 55, 213, 236, 151, 88, 43, 156, 177, 61, 179, 147, 55, 5, 179, 43, 169, 130, 175, 90, 242, 95, 215, 136, 129, 235, 179, 39, 113, 252, 89, 34, 239, 198, 110, 163]`,
 right: `[179, 77, 104, 140, 113, 27, 241, 243, 3, 27, 237, 173, 214, 221, 46, 61, 98, 13, 145, 199, 80, 58, 160, 96, 2, 171, 117, 128, 88, 175, 141, 194, 239, 225, 35, 117, 214, 130, 65, 221, 88, 38, 146, 1, 167, 11, 172, 54, 106, 122, 83, 94, 88, 208, 142, 255, 85, 40, 67, 58, 64, 34, 36, 59]`', k256/src/schnorr.rs:315:13
```

I'm uncertain why. As the comment notes, reducing the message mod n/p would cause a failure, however the current implementation only performs a reduction of the output of `tagged_hash` which seems orthogonal to what that's referring to.

tarcieri: The notable property of this test vector per the code that generates it is that the secret key needs to be negated before signing:

https://github.com/bitcoin/bips/blob/274fa400d630ba757bec0c03b35ebe2345197108/bip-0340/test-vectors.py#L73

k256/src/schnorr.rs
```diff
@@ -0,0 +1,319 @@
...
+        aux_rand: &[u8; 32],
+    ) -> Result<Signature> {
+        let d = if self.verifying_key.is_y_even() {
+            self.secret_key
+        } else {
+            -self.secret_key
+        };
+
+        let t = xor(
+            &self.secret_key.to_bytes(),
```
nickray, line 101: This should be `&d.to_bytes()`, not `&self.secret_key.to_bytes()`, then case 3 generates the right signature.

nickray:
Rebased onto v0.10.4, adding missing verify test vectors: https://github.com/RustCrypto/elliptic-curves/compare/k256/v0.10.4...nickray:bip340-ideas-v0.10.4

tarcieri:
@nickray nice! Perhaps I could merge the incomplete implementation with some TODOs and you could open a followup PR to that?

tarcieri:
Merged in #554

PR #490, opened 2021-12-18
p384: base field arithmetic

tarcieri:
Vendors formally verified secp384r1 base field arithmetic generated by the fiat-crypto project:

https://github.com/mit-plv/fiat-crypto/tree/master/fiat-rust

The generated code is triple licensed Apache 2.0, MIT, and BSD.

Vendored code comes from the following commit:

https://github.com/mit-plv/fiat-crypto/commit/c92bed39e597f875536c4c70a37ae09909acabae

Commit: c92bed39e597f875536c4c70a37ae09909acabae (tag: v0.0.10)
Author: brycx <brycx@protonmail.com>
Date:   Mon Dec 13 09:37:54 2021 +0100
Title:  Update generated Rust files


tarcieri:
I was able to import the fiat-crypto field backends in such a way that they can be used verbatim from upstream: same filenames, contents unmodified.

That should simplify auditing and upgrades.

Issue #439, opened 2021-09-28
Exposing FieldElement?


randombit:
We're interested in supporting hash2curve for k256 and p256 but these require arithmetic on field elements (*not* scalars), and the field elements are not exposed.

tarcieri:
Thus far we've deliberately avoided exposing field elements in order to provide a misuse resistant API.

For hash2curve specifically my suggestion would be [defining traits for hash2curve in the `elliptic-curve` crate](https://github.com/RustCrypto/traits/issues/481) and then implementing them directly in `k256`/`p256`.

str4d:
I agree with @tarcieri. I would definitely like to see a hash-to-curve trait (either in `group` or `elliptic-curve`) that can provide this functionality, and if the hash-to-field component of the ID proves to be more widely-useful, then a trait to that effect could end up in `ff`.

randombit:
Wouldn't this effectively require implementing SSWU and related logic twice, once for each curve?

tarcieri:
You should be able to define it as e.g. a generic function somewhere like `elliptic-curve`

tarcieri:
Another approach that might work for abstracting SSWU would be to define a trait (e.g. `Sswu`) intended to be impl'd on a curve-specific `FieldElement` type, which has a `map_to_curve_simple_swu` function with a default impl and an associated curve-specific `const Z`.

Then all curve-specific impls of that trait would need to do is define `Z`.

tarcieri:
Generic optimized SWU has been added upstream to the `elliptic-curve` crate: https://github.com/RustCrypto/traits/pull/854

PR #495, opened 2021-12-21
feat: hash to curve

mikelodder7:
Adds hash to curve according to [draft v13](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/).

For now is enabled with the `hashing` feature.


k256/Cargo.toml
```diff
@@ -24,6 +24,7 @@ sec1 = { version = "0.2", default-features = false }
 
 # optional dependencies
 ecdsa-core = { version = "0.13", package = "ecdsa", optional = true, default-features = false, features = ["der"] }
+hash2field = { version = "0.4", optional = true }
```
newpavlov, line 27: @tarcieri Maybe it's worth to incorporate this crate into `elliptic-curve`?

tarcieri: @mikelodder7 would you be interested in upstreaming `hash2field` somewhere, either `elliptic-curve` or possibly `ff`?

mikelodder7: Sure thing. Elliptic-curve probably makes more sense but I can be convinced otherwise

k256/src/arithmetic/field.rs
```diff
@@ -225,6 +228,252 @@ impl FieldElement {
...
+            }
+        }
+        res
+    }
+
+    /// Optimized Simplified WWU map
+    #[cfg(feature = "hashing")]
+    pub fn oswwu(&self) -> (Self, Self) {
+        // See section 8.7 in
+        // <https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/>
```
tarcieri, line 285: Should this have been Section 8.2? https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#name-suites-for-nist-p-256

mikelodder7: Both p256 and k256 are implemented.
P256 is 8.2
K256 is 8.7

k256/src/arithmetic/field.rs
```diff
@@ -225,6 +228,252 @@ impl FieldElement {
...
+
+                res.conditional_assign(&(res * self), Choice::from(((*e >> i) & 1) as u8));
+            }
+        }
+        res
+    }
+
+    /// Optimized Simplified WWU map
+    #[cfg(feature = "hashing")]
+    pub fn oswwu(&self) -> (Self, Self) {
```
tarcieri, line 283: What is WWU exactly?

Isn't this Simplified SWU as in the Shallue-van de Woestijne-Ulas method?

k256/src/arithmetic/field.rs
tarcieri: Oh whoops, lost track of what file I was looking at 😅

k256/src/arithmetic/field.rs
mikelodder7: Yeah typo. I just updated it.

k256/Cargo.toml
tarcieri: Sounds good!

If you're up for it, I think it might also be possible to extract a generic implementation of SSWU, although I'm fine with duplication for now and I can try to extract it as part of a follow-up:

https://github.com/RustCrypto/elliptic-curves/issues/439#issuecomment-929500577

You can use e.g. `ff::{Field, PrimeField}` as the bounds, and I can look into ensuring they're impl'd for the respective `FieldElement` types in `k256` and `p256`.

mikelodder7: Yeah if `ff::Field` is impl'd then the dedup is easy from there. `osswu` will just have to take a set of params but otherwise, they'll be the same. Should `osswu` go upstream as well in `elliptic-curve`?

tarcieri: I think it'd be great if you can add a generic implementation, and in parallel I can work on trying to get the `ff` traits impl'd for the `FieldElement` types for `k256` and `p256`.

mikelodder7: Cool. I'd extract the `osswu` method and `hash2field` crate to `elliptic-curve`.

tarcieri: Awesome!

mikelodder7:
OSSWU upstreamed in [PR 854](https://github.com/RustCrypto/traits/pull/854)

daxpedda:
I tested this as part of opaque-ke and it works great so far, following requests:
- It would be amazing if the new `hash_from_bytes` and `encode_from_bytes` can be part of a trait specified in elliptic-curve.
- Currently everything is already in place to support `hash_to_field()`, which is the new `FieldElement::encode()` function. It would be great if this too can be exposed for `Scalar`, preferably through a trait specified in elliptic-curve too.

tarcieri:
It seems like with #503 merged this can be closed?

PR #498, opened 2021-12-23
k256+p256: impl `ff::Field` trait for `FieldElement` types

tarcieri:
Allows for writing code which is generic over field elements, such as the optimized SWU implementation added in RustCrypto/traits#854

### DIFF
--- a/k256/src/arithmetic/field.rs
+++ b/k256/src/arithmetic/field.rs
@@ -1,5 +1,7 @@
 //! Field arithmetic modulo p = 2^256 - 2^32 - 2^9 - 2^8 - 2^7 - 2^6 - 2^4 - 1
 
+#![allow(clippy::assign_op_pattern, clippy::op_ref)]
+
 use cfg_if::cfg_if;
 
 cfg_if! {
@@ -30,10 +32,12 @@ cfg_if! {
 }
 
 use crate::FieldBytes;
-use core::ops::{Add, AddAssign, Mul, MulAssign};
+use core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
 use elliptic_curve::{
+    ff::Field,
+    rand_core::RngCore,
     subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption},
-    zeroize::Zeroize,
+    zeroize::DefaultIsZeroes,
 };
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


tarcieri:
One thing I'm a bit worried about in regard to generic field arithmetic is lazy normalization used by `k256`. It feels like a potentially leaky abstraction.

k256/src/arithmetic/field.rs
```diff
@@ -1,5 +1,7 @@
 //! Field arithmetic modulo p = 2^256 - 2^32 - 2^9 - 2^8 - 2^7 - 2^6 - 2^4 - 1
 
+#![allow(clippy::assign_op_pattern, clippy::op_ref)]
```
tarcieri, line 3: Left all of the current arithmetic borrowing as-is in order to not affect performance as part of this PR, but it's probably worth benchmarking removing these in a followup and seeing if it affects performance

p256/src/arithmetic/affine.rs
```diff
@@ -1,5 +1,7 @@
 //! Affine points
 
+#![allow(clippy::op_ref)]
```
mikelodder7, line 3: Could probably fix this in a subsequent PR

k256/src/arithmetic/field.rs
mikelodder7: Yeah I can fix in a follow up PR. I did this crypto-bigint

mikelodder7:
> One thing I'm a bit worried about in regard to generic field arithmetic is lazy normalization used by `k256`. It feels like a potentially leaky abstraction.

How much performance do we loose if normalization is applied with each call? We could add a benchmark for it

k256/src/arithmetic/field.rs
tarcieri: It's really easy to fix automatically with `cargo clippy --fix`, but I just want to make sure it doesn't cause a performance regression

tarcieri:
> How much performance do we loose if normalization is applied with each call? We could add a benchmark for it

I think it could potentially be a lot for the `core::ops::*` traits.

mikelodder7:
> > How much performance do we loose if normalization is applied with each call? We could add a benchmark for it
> 
> I think it could potentially be a lot for the `core::ops::*` traits.

Hmm well not sure how to avoid the leaky then

mikelodder7 (approved):

tarcieri:
Given that `FieldElement` is an internal type, I think this is less worrisome than leaking these details outside the crate, but I would expect cases where a generic implementation may not do what you expect with `k256`'s `FieldElement` type.

Going to go ahead and merge this, but I'd love to get some additional feedback if anyone has anything to add in perpetuity.

PR #501, opened 2021-12-30
k256: add impl of `ReduceNonZero<U256>` for `Scalar`

fjarri:
- add impl of `ReduceNonZero<U256>` for `Scalar`
- fix a bug in `WideScalar::reduce_nonzero()` (a zero value returned when reducing `MODULUS-1`)

Very embarrassing for me, but at least chances of someone hitting it were astronomically small, and if they did, it most probably resulted in a panic during inversion or something of the sort.

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -591,8 +591,8 @@ impl Reduce<U512> for Scalar {
 
 impl ReduceNonZero<U256> for Scalar {
     fn from_uint_reduced_nonzero(w: U256) -> Self {
-        let order_minus_one = ORDER.wrapping_sub(&U256::ONE);
-        let (r, underflow) = w.sbb(&order_minus_one, Limb::ZERO);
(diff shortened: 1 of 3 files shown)


tarcieri:
> there may be a small performance penalty for calculating order_minus_one in the method itself

It's `const fn` so I'd hope not, although binding it to a function-scoped constant would be a way to make sure.

> should I add tests?

That'd be nice 😉

fjarri:
> It's const fn so I'd hope not, although binding it to a function-scoped constant would be a way to make sure.

Fixed.

Also it's a good thing you asked me to add tests...

tarcieri:
Thank you!

fjarri:
Btw, are you planning on making a release of `k256` any time soon? My other project would benefit from having this available.

tarcieri:
Sure, I can cut a release

tarcieri:
In case you didn't see, I cut `k256` v0.10.1 yesterday.

Another `elliptic-curve` crate release might take a bit longer as I'd like it to include hash-to-curve work, which is currently half-finished.

PR #503, opened 2022-01-07
Implement `GroupDigest` for `NistP256` and `Secp256k1`

daxpedda:
See RustCrypto/traits#865.
This replaces #495 where I copied some code from @mikelodder7. The reason I did that is to test changes, for example, for RustCrypto/traits#871.

Currently builds on top of RustCrypto/traits#876.


k256/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,396 @@
...
+                0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xfe,
+                0xff, 0xff, 0xfc, 0x24,
+            ]);
+
+            let tv1 = u.square(); // u^2
+            let tv3 = Z * tv1; // Z * u^2
+            let mut tv2 = tv3.square(); // tv3^2
+            let mut xd = tv2 + tv3; // tv3^2 + tv3
+            let x1n = MAP_B * (xd + FieldElement::one()); // B * (xd + 1)
+            xd = (xd * MAP_A.negate(1)).normalize(); // -A * xd
```
daxpedda, line 135: I couldn't use `OsswuMap` from elliptic-curve because it's missing `normalize` in a couple of places, which it fail without. @mikelodder7 if you could make a PR for that would be amazing!

k256/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    fn into_subgroup(self) -> CtOption<Self::Subgroup> {
+        CtOption::new(self, 1.into())
+    }
+
+    fn is_torsion_free(&self) -> Choice {
+        1.into()
+    }
+}
```
daxpedda, line 288: I don't know what I'm doing here, any feedback would be appreciated.

p256/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    fn into_subgroup(self) -> CtOption<Self::Subgroup> {
+        CtOption::new(self, 1.into())
+    }
+
+    fn is_torsion_free(&self) -> Choice {
+        1.into()
+    }
+}
```
daxpedda, line 119: As above.

k256/src/arithmetic/hash2curve.rs
mikelodder7: The problem only happens when using `ct_eq` and `is_zero`. I’ve noticed that sometimes these return false without calling normalize and true when normalize is called. This sounds like a bug to me. As I’ve tested this in the `k256` crate when I add normalize calls to those two functions then everything works as expected and you can use osswu. Do you foresee a problem with this change? I can’t imagine it being one

k256/src/arithmetic/hash2curve.rs
mikelodder7: Yeah for k256 cofactor is 1 so every valid point is also in the subgroup

k256/src/arithmetic/hash2curve.rs
daxpedda: I'm pretty much out of my depth to answer this, @tarcieri if you could chime in here please.

mikelodder7:
I’d like to not have to copy osswu into k256. I’ve got it working locally with only those two changes

daxpedda:
I agree, we will need to fix this in elliptic-curve.

k256/src/arithmetic/hash2curve.rs
tarcieri: [I was worried about that before, as it were](https://github.com/RustCrypto/elliptic-curves/pull/498#issuecomment-1000376582).

You might need to use a copy-paste implementation for `k256` in order to accommodate lazy normalization. 

FWIW we do the same thing for ECDSA for similar reasons.

`p256` should be able to use the generic implementation though (of both OSSWU and ECDSA, as it were).

daxpedda: That's too bad, if we don't come up with a better idea then I will "clean" this up, putting it into a separate function at least should be better.

tarcieri: Unfortunately we haven’t found a good abstraction to encapsulate lazy normalization. It’s a tricky problem.

The performance gains definitely make it worth it though.

daxpedda: I decided that the easiest way to clean this up was just to override the `OsswuMap` implementation.

p256/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,283 @@
...
+    fn from_okm(data: &GenericArray<u8, Self::Length>) -> Self {
+        let n = NonZero::new(U384::from_be_hex(
+            "00000000000000000000000000000000ffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551",
+        ))
+        .unwrap();
+
+        let bytes = U256::from_be_slice(&(U384::from_be_slice(data) % n).to_be_byte_array()[16..]);
+        Scalar(bytes)
+    }
+}
```
daxpedda, line 133: I can't do the magic that @mikelodder7 did, so this is my amateur solution. It works.

p256/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,283 @@
...
+            dst: b"HashToScalar-VOPRF08-\x00\x00\x03",
+            seed: &hex!("a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3"),
+            sk_sm: &hex!("c15d9e9ab36d495d9d62954db6aafe06d3edabf41600d58f9be0737af2719e97"),
+        },
+        TestVector {
+            dst: b"HashToScalar-VOPRF08-\x01\x00\x03",
+            seed: &hex!("a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3"),
+            sk_sm: &hex!("7f62054fcd598b5e023c08ef0f04e05e26867438d5e355e846c9d8788d5c7a12"),
+        },
+    ];
```
daxpedda, line 272: I copied these test vectors from the VOPRF spec, I don't know any other spec providing test vectors for `hash_to_scalar`. The VOPRF spec also doesn't define any for k256.

p256/src/arithmetic/hash2curve.rs
daxpedda: I was taught some black magic by @mikelodder7 in the meantime.

mikelodder7:
I'm happy to move my other PR which implements all the hash2curve to something similar to this unless you are planning to turn this into a regular PR. Thoughts?

daxpedda:
The plan is to turn this into a regular PR actually, I just can't until a new version of elliptic-curve is released. So as soon as that's done I'm going to update this.

daxpedda:
I can't make sense of this error. `infinity` should be a `Choice`, not a `u8`.
```
    Checking k256 v0.10.1 (/home/runner/work/elliptic-curves/elliptic-curves/k256)
error[E0308]: mismatched types
   --> k256/src/arithmetic/hash2curve.rs:144:23
    |
144 |             infinity: Choice::from(0u8),
    |                       ^^^^^^^^^^^^^^^^^ expected `u8`, found struct `elliptic_curve::subtle::Choice`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `k256` due to previous error
```

tarcieri:
@daxpedda that was a recent change (#511) which was needed to allow an inherent constants for `IDENTITY` and `GENERATOR`, since `Choice` can't be used in a `const` context

daxpedda:
Thanks, this should be good to go now.

p256/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,324 @@
...
+            0,
+        ])
+        .to_montgomery();
+        d0 * F_2_192 + d1
+    }
+}
+
+impl Sgn0 for FieldElement {
+    fn sgn0(&self) -> Choice {
+        FieldElement::sgn0(self)
```
tarcieri, line 51: Seems like you could just call `self.is_odd()` here rather than adding a separate inherent method in addition to the trait method

k256/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,370 @@
...
+
+        // if e2 , x = x1, else x = x2
+        let mut x = Self::conditional_select(&x2n, &x1n, e2);
+        // xn / xd
+        x *= xd.invert().unwrap();
+
+        // if e2, y = y1, else y = y2
+        let mut y = Self::conditional_select(&y2, &y1, e2);
+
+        y.conditional_assign(&-y, self.normalize().sgn0() ^ y.normalize().sgn0());
```
daxpedda, line 129: I noticed when removing the `sgn0` method and internally using `is_odd` instead, it complained that it wasn't normalized. Does that look alright @mikelodder7?

p256/src/arithmetic/hash2curve.rs
daxpedda: Done.

k256/src/arithmetic/hash2curve.rs
daxpedda: https://github.com/RustCrypto/elliptic-curves/blob/8a44a3f8ba2844dc30643c63b4c6c5cb8a4fb24b/k256/src/arithmetic/field/field_impl.rs#L91-L94

tarcieri: @fjarri do you think `is_odd` should call `normalize` first?

mikelodder7: Anytime comparisons are used for osswu it has to be normalized otherwise it gets wrong results

fjarri: Yes, I would assume you need to normalize, since the modulus is odd, so `x` and `x+modulus` (possible even if you have magnitude=1, but not normalized) will have different parity.

daxpedda: Alright, will change `is_odd` to normalize first then.

daxpedda: Done.

daxpedda:
Re-based because of #506.

PR #506, opened 2022-01-09
Implement `VoprfParameters` for `NistP256` and `NistP384`

daxpedda:
See RustCrypto/traits#878.
Reference: https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-08.html#section-4.


daxpedda:
Good to go!

PR #507, opened 2022-01-09
Implement `ReduceNonZero<U256>` for `p256::Scalar`

daxpedda:
IANA cryptographer, I basically copied this from k256.


tarcieri:
This is a correct implementation for `p256` as well. Thank you.

PR #530, opened 2022-03-14
k256: Do not normalize the argument in `FieldElementImpl::is_odd()`

### LINKED ISSUE #529: k256 Point deserialization+serialization round trips incorrectly
TBH I'm not sure if this is a bug in `k256` or a Rust miscompilation bug.

We've found that in certain cases `k256` compressed point serialization and deserialization do not round trip correctly. In particular the sign of `y` will for certain rare points, be flipped.

This is using `k256` 0.10.2, `rustc` 1.58.1

```
use k256::elliptic_curve::group::GroupEncoding;
use k256::elliptic_curve::sec1::FromEncodedPoint;

pub fn deserialize(bytes: &[u8]) -> Option<k256::ProjectivePoint> {
    match k256::EncodedPoint::from_bytes(bytes) {
        Ok(ept) => {
            let apt = k256::AffinePoint::from_encoded_point(&ept);

            if bool::from(apt.is_some()) {
                Some(k256::ProjectivePoint::from(apt.unwrap()))
            } else {
                None
            }
        }
        Err(_) => None,
    }
}

pub fn serialize(pt: k256::ProjectivePoint) -> Vec<u8> {
    pt.to_affine().to_bytes().to_vec()
}

fn main() {
    let bits =
        hex::decode("024b395881d9965c4621459ad2ec12716fa7f669b6108ad3b8b82b91644fb44808").unwrap();

    let pt = deserialize(&bits).unwrap();

    let pt_bytes = serialize(pt);

    assert_eq!(bits, pt_bytes);
}
```

In release mode (only), this fails:

```
$ cargo run --release
   Compiling k256-bug v0.1.0 (/home/jack/sw/k256-bug)
    Finished release [optimized] target(s) in 0.58s
     Running `target/release/k256-bug`
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[2, 75, 57, 88, 129, 217, 150, 92, 70, 33, 69, 154, 210, 236, 18, 113, 111, 167, 246, 105, 182, 16, 138, 211, 184, 184, 43, 145, 100, 79, 180, 72, 8]`,
 right: `[3, 75, 57, 88, 129, 217, 150, 92, 70, 33, 69, 154, 210, 236, 18, 113, 111, 167, 246, 105, 182, 16, 138, 211, 184, 184, 43, 145, 100, 79, 180, 72, 8]`', src/main.rs:31:5
```

So far we have not been able to reproduce this in either debug mode or with coverage guided fuzzing enabled.

fjarri:
Fixes #529

`FieldElementImpl::is_odd()` (which was only compiled in debug version) normalized its argument, wherein in release mode it was bypassed in favor of the specific 32- or 64-bit implementation. This cause an inconsistency between the release and debug versions, and lead to uncaught issues where an un-normalized field element could be checked for `is_odd()` (the exact situation `FieldElementImpl` was designed to detect). 

In this PR:
- `FieldElementImpl::is_odd()` now checks for normalization instead of performing it
- `AffinePoint::decompress()` normalizes `y` coordinate (`beta` in the code) before checking if it is odd
- `Scalar::try_sign_prehashed` normalizes `R.y` before checking if it is odd

I added a regression test for the first change; the other two are already tested by `arithmetic::affine::tests::compressed_round_trip`, `arithmetic::affine::tests::compressed_to_uncompressed` and `ecdsa::recoverable::tests::public_key_recovery` - those tests only passed because they were always run in debug mode where force normalization was on.


k256/src/arithmetic/field.rs
```diff
@@ -542,6 +544,28 @@ mod tests {
...
+        let x = FieldElement::from_bytes_unchecked(&[
+            61, 128, 156, 189, 241, 12, 174, 4, 80, 52, 238, 78, 188, 251, 9, 188, 95, 115, 38, 6,
+            212, 168, 175, 174, 211, 232, 208, 14, 182, 45, 59, 122,
+        ]);
+        // Produces an unnormalized FieldElement with magnitude 1
+        // (we cannot create one directly).
+        let y = x.sqrt().unwrap();
+        // Panics since `y` is not normalized.
+        let result = std::panic::catch_unwind(|| y.is_odd());
+        assert!(result.is_err());
```
tarcieri, line 564: `std` isn't always going to be available in unit tests, since this is a `no_std` crate.

There are a couple options here to keep the tests `no_std`-friendly:

1. Write a `#[should_panic]` unit test: https://doc.rust-lang.org/reference/attributes/testing.html#the-should_panic-attribute
2. Use an integration test in `tests/` where `std` is linked unconditionally

I would personally suggest the first option

fjarri: Yeah, I just noticed that fails too. On it.

newpavlov: IIRC it's not trivial to run tests in a `no_std` environment, since the `test` crates currently depends on `std`.

A bigger issue is that tests may be compiled with `panic="abort"`, with which `catch_unwind` obviously will not work.

I agree with going with the first option.

fjarri: I guess those fails are with `panic="abort"` where the test doesn't panic. Any ideas how to gate it?

k256/src/arithmetic/field.rs
```diff
@@ -542,6 +542,29 @@ mod tests {
         assert_eq!(four.sqrt().unwrap().normalize(), two.normalize());
     }
 
+    #[test]
+    #[should_panic(expected = "assertion failed: self.normalized")]
```
tarcieri, line 546: Another potential issue is that test will only succeed (i.e. `is_odd` will only panic) when `debug_assertions` are enabled.

Perhaps the test, or at least the `should_panic`, should be gated on that?

```suggestion
    #[cfg_attr(debug_assertions, should_panic(expected = "assertion failed: self.normalized"))]
```

tarcieri:
FYI: I opened #531 with a proposal to encapsulate concerns around normalizing field elements using type safety

Issue #531, opened 2022-03-14
k256: type safety for non-normalized field elements


tarcieri:
The `k256` crate uses lazy normalization of field elements. While not a user-facing concern as we deliberately encapsulate `FieldElement`, there is a potential for bugs in code in `k256` itself in the event one "forgets" to normalize a field element prior to returning it as the result of some computation. See #530 as an example (although there have been others).

One potential way to eliminate this class of bugs is to use separate types for normalized vs non-normalized field elements. For example, we could introduce something like `LazyFieldElement` on which all of the arithmetic operations are defined, while retaining `FieldElement` for the normalized form and defining all serialization operations on that.

`LazyFieldElement::normalize` could return `FieldElement`, and we could additionally have bidirectional `From` conversions between the two.

Things would be a bit tricky in regard to the the `ff::Field` traits. I imagine for compatibility reasons they would need to be impl'd on `FieldElement`, converting to a `LazyFieldElement` to perform the arithmetic, and then calling `normalize()` to get a `FieldElement` again as a result, which would mean that the performance benefits of lazy normalization wouldn't be available under such an API. However, that is the only safe usage pattern since the traits are designed to abstract over different fields/field implementations, and not all of them have lazy normalization.

cc @fjarri

fjarri:
I think it's a great idea for preventing errors, but yes, trait stuff might get difficult.

kayabaNerve:
I'd consider a `Lazy<F: LazyField>` where the `LazyField` is a trait defining the amount of additions which can be performed before overflow.

My immediately prior work was reviewing a 255-bit field where 3 * q > 2**256. This allowed it to perform a single addition without overflow, but of course, reducing it would still require two reduction steps (due to needing to detect if one or two steps is necessary, with the fastest way to detect if a step is necessary is to attempt it and see if it underflows). Accordingly, the only actual performance benefit would be when *then used in a product statement* which produces a 512-bit output, which would already be fully reduced even if the top two bits were unset (as they're within the limb boundary).

This, to me, implies the most basic API as a `Lazy<F>` which upon one addition, is explicitly reduced or used in a product statement.

For compatibility with non-amenable field (implementations), a `NotLazy<F>` wrapper type can be provided which defers lazy_add to regular addition, allowing libraries to require specification of *some* lazy curve, and the user having the option of whether or not to actually provide one.

The challenge from there, for our friends with even shorter elliptic curves, is integrating the `Lazy` type with a solution such as `typenum`.

[...]

PR #533, opened 2022-03-15
k256: normalize before calling is_odd() in sng0()

fjarri:
An `is_odd()` call I forgot to address in #530

While this fixes the problem at hand, there are a couple of considerations:
- To prevent future problems like this, we do need either #531, or just auto-normalizing in `FieldElement::is_odd()` (but not in `FieldElementImpl`!). Originally I didn't do that because I wanted to leave a possibility for a slightly more efficient code (where the caller would know whether the value has been previously normalized or not), but perhaps it is not worth it, and it's better to prevent errors instead.
- This was not caught in tests in #530 because in CI we run tests with `--release`, and apparently this bug does not affect `hash2curve` test, which is the only one touching `sgn0()`. @tarcieri , why was that decision made originally?

### DIFF
--- a/k256/src/arithmetic/hash2curve.rs
+++ b/k256/src/arithmetic/hash2curve.rs
@@ -42,7 +42,7 @@ impl FromOkm for FieldElement {
 
 impl Sgn0 for FieldElement {
     fn sgn0(&self) -> Choice {
-        self.is_odd()
+        self.normalize().is_odd()
     }
 }
 


tarcieri:
> in CI we run tests with --release, and apparently this bug does not affect hash2curve test, which is the only one touching sgn0(). @tarcieri , why was that decision made originally?

Are you asking why CI runs the tests with `--release`?

We've had bugs in the past that only affected the release profile. An easy way to screw up is to perform some sort of effectful operation inside of a `debug_assert` which is completely elided in a release build.

Release builds are what people actually use, so we test against them as they're more important.

We can potentially test both debug and release builds if you think that'd be helpful.

fjarri:
> We can potentially test both debug and release builds if you think that'd be helpful.

I actually not sure now how it happened, because `FieldElementImpl` safety harness should have still been active when CI ran in #530 - but I see `hash2curve` test passing there, but if I run `cargo test --all-features` locally it fails. Yesterday I didn't notice it because I only ran `cargo test` with default features locally.

tarcieri (approved):

fjarri:
I hope it's not too much to ask to publish k256 0.10.4 with this? My library is using hash2curve, that's how I noticed this - in my own tests.

tarcieri:
v0.10.4 released: #534

PR #552, opened 2022-05-10
k256: use SHA-256 for computing RFC6979 for `recoverable::Signature`

tarcieri:
For whatever reason when the Ethereum developers decided to use Keccak256 as the digest function for ECDSA itself, they still used SHA-256 as the digest function for RFC6979, possibly because they didn't realize they needed to change the digest function in two places.

Previously the implementation in `k256` would use the same digest function for both (typically Keccak256) but we've had various complaints that doing so doesn't match the test vectors produced by various other libraries (though the signatures will still verify).

This commit uses HMAC-DRBG instantiated with SHA-256 for computing RFC6979 for `ecdsa::recoverable::Signature`, regardless of what digest function is used to hash the message. In theory this should match the Ethereum test vectors.

### DIFF
--- a/k256/Cargo.toml
+++ b/k256/Cargo.toml
@@ -39,12 +39,12 @@ proptest = "1.0"
 rand_core = { version = "0.6", features = ["getrandom"] }
 
 [features]
-default = ["arithmetic", "ecdsa", "pkcs8", "sha256", "std"]
+default = ["arithmetic", "ecdsa", "pkcs8", "std"]
 arithmetic = ["elliptic-curve/arithmetic"]
 bits = ["arithmetic", "elliptic-curve/bits"]
 digest = ["ecdsa-core/digest", "ecdsa-core/hazmat"]
 ecdh = ["arithmetic", "elliptic-curve/ecdh"]
-ecdsa = ["arithmetic", "digest", "ecdsa-core/sign", "ecdsa-core/verify"]
+ecdsa = ["arithmetic", "ecdsa-core/sign", "ecdsa-core/verify", "sha256"]
 expose-field = ["arithmetic"]
 hash2curve = ["arithmetic", "elliptic-curve/hash2curve"]
 jwk = ["elliptic-curve/jwk"]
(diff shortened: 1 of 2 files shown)


nickray:
Would you happen to know of any "canonical" test vectors for this?

tarcieri:
I believe there are test vectors like that in ethers-rs somewhere /cc @gakonst @agostbiro 

https://github.com/gakonst/ethers-rs

agostbiro:
@nickray no canonical test vectors afaik, but I'm using the one below in a soon-to-be open source project to check that I'm producing identical signatures to `ethers-rs`. The values are based on [this](https://github.com/gakonst/ethers-rs/blob/ba00f549dd05e72fed7f00703f29752324645172/ethers-signers/src/wallet/private_key.rs#L197) unit test in `ethers-rs`.


```rust
const SECRET_KEY: &str = "4c0883a69102937d6231471b5dbb6204fe5129617082792ae468d01a3f362318";

const TX_BYTES: &str = "e9808504e3b29200831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca0080018080";

const ETH_MAINNET_LEGACY_SIGNATURE: &str = "f869808504e3b29200831e848094f0109fc8df283027b6285cc889f5aa624eac1f55843b9aca008025a0c9cf86333bcb065d140032ecaab5d9281bde80f21b9687b3e94161de42d51895a0727a108a0b8d101465414033c3f705a9c7b826e596766046ee1183dbc8aeaa68";
```

Please note that the signature is not directly the `RecoverableSignature` output by this library, but [adjusted](https://github.com/gakonst/ethers-rs/blob/916e9a73341fab3b326be6ddb6fcd61550c8fc7f/ethers-signers/src/lib.rs#L51) as per EIP-155 and [encoded](https://github.com/gakonst/ethers-rs/blob/0abc3ca39ae54fc11a7541da9e92bbebfe51c899/ethers-core/src/types/transaction/request.rs#L184) as RLP in the legacy Ethereum transaction format.

PR #563, opened 2022-05-25
p256: use a 4-bit window for scalar multiplication

jedisct1:
Immediately improves performance even on a variable base:

- 35% faster on x86_64
- 37% faster on M1
- 37% faster on WebAssembly (Wasmtime)
- 39% faster on WebAssembly (Wasmedge)

### DIFF
--- a/p256/src/arithmetic/projective.rs
+++ b/p256/src/arithmetic/projective.rs
@@ -210,9 +210,9 @@ impl ProjectivePoint {
         loop {
             let slot = (k[31 - (pos >> 3) as usize] >> (pos & 7)) & 0xf;
             let mut t = ProjectivePoint::IDENTITY;
-            for i in 1..16 {
+            for (i, pci) in pc.iter().enumerate().skip(1) {
                 t.conditional_assign(
-                    &pc[i],
+                    pci,
                     Choice::from(((slot as usize ^ i).wrapping_sub(1) >> 8) as u8 & 1),
                 );
             }
--- a/p256/src/arithmetic/projective.rs
+++ b/p256/src/arithmetic/projective.rs
@@ -204,8 +204,7 @@ impl ProjectivePoint {
             };
         }
         let mut q = ProjectivePoint::IDENTITY;
-        let k_ = k.to_bytes();
-        let k = k_.as_slice();
+        let k = k.to_bytes();
         let mut pos = 256 - 4;
         loop {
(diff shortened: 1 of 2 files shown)


p256/src/arithmetic/projective.rs
```diff
@@ -194,16 +193,37 @@ impl ProjectivePoint {
...
     /// Returns `[k] self`.
     fn mul(&self, k: &Scalar) -> ProjectivePoint {
-        let mut ret = ProjectivePoint::IDENTITY;
-
-        for limb in k.limbs().iter().rev() {
-            for i in (0..Limb::BIT_SIZE).rev() {
-                ret = ret.double();
-                ret.conditional_assign(&(ret + self), Choice::from(((limb.0 >> i) & 1) as u8));
+        let mut pc = [ProjectivePoint::default(); 16];
+        pc[0] = ProjectivePoint::IDENTITY;
```
tarcieri, line 197: FWIW, `ProjectivePoint::default()` returns `ProjectivePoint::IDENTITY`.

I suppose you're initializing it explicitly because `ProjectivePoint::default()` is effectively a placeholder for an "uninitialized" element of the array which will be overwritten?

p256/src/arithmetic/projective.rs
```diff
@@ -194,16 +193,37 @@ impl ProjectivePoint {
...
+            } else {
+                pc[i - 1].add(self)
+            };
+        }
+        let mut q = ProjectivePoint::IDENTITY;
+        let k_ = k.to_bytes();
+        let k = k_.as_slice();
+        let mut pos = 256 - 4;
+        loop {
+            let slot = (k[31 - (pos >> 3) as usize] >> (pos & 7)) & 0xf;
```
tarcieri, line 211: Was there actually an issue here using the `Index` impl on `GenericArray`? i.e. why doesn't this work:
```suggestion
        let k = k.to_bytes();
        let mut pos = 256 - 4;
        loop {
            let slot = (k[31 - (pos >> 3) as usize] >> (pos & 7)) & 0xf;
```

tarcieri:
Thanks! Impressive performance improvement

p256/src/arithmetic/projective.rs
jedisct1: Sure!

p256/src/arithmetic/projective.rs
jedisct1: Exactly.

 Elements are initialized right after and the compiler doesn't emit any initialization code.

PR #564, opened 2022-05-25
p256 invert(), sqrt(): replace exponentiations with addition chains

jedisct1:
```text
invert() using pow_vartime: 256 squarings, 128 multiplications
         addition chain   : 255 squarings,  12 multiplications
```

```text
sqrt()   using pow_vartime: 256 squarings, 129 multiplications
         addition chain   : 253 squarings,   7 multiplications
```

Ported from Zig's standard library.


tarcieri:
Thank you!

PR #565, opened 2022-05-26
p384: import arithmetic from jedisct1/rust-p384

### LINKED ISSUE #240: p384: arithmetic + ECDSA support
Hi,

Is there a way to compute and verify p384 signatures yet?

`p384` has an `ecdsa` feature, but I couldn't get anything done with it.
`ProjectiveArithmetic` is not implemented for `NistP384`. The `arithmetic` feature of the `elliptic_curve` crate isn't set, so I wasn't even able to manually compute a public key.

Is it possible to use it in a similar way as `p256` and `k256`? Or to use it for ECDSA at all? Or is it still a work in progress?

Thanks for your help :)

tarcieri:
Imports an MVP arithmetic implementation from:

https://github.com/jedisct1/rust-p384

This includes the following:

- Addition formulas from Renes-Costello-Batina 2015, adapted from @str4d's implementation in the `p256` crate
- `FieldElement::{invert, sqrt}` implementations
- `Scalar::{invert, sqrt}` implementations
- Scalar field implementation generated using fiat-crypto
- Scalar multiplication using 4-bit window ala #563
- Point compaction support
- ECDH support using generic implementation from the `elliptic-curve` crate
- ECDSA support using generic implementation from the `ecdsa` crate

Closes #240


p384/src/arithmetic/scalar/p384_scalar.rs
```diff
@@ -0,0 +1,3813 @@
+//! Autogenerated: './word_by_word_montgomery' --lang Rust --inline p384_scalar
+//! 64 39402006196394479212279040100143613805079739270465446667946905279627659399113263569398956308152294913554433653942643
+//! curve description: p384_scalar
```
tarcieri, line 3: Where did you get this? Did you generate it yourself?

I don't see pregenerated versions in the fiat-rust tree: https://github.com/mit-plv/fiat-crypto/tree/master/fiat-rust/src

Anyway, this is great!

jedisct1: Yes I generated it. It's deterministically reproducible using the command above.

We're using Fiat for p256 scalars in Zig, Go is looking at doing the same thing, so maybe the scalar fields can be added to what gets pregenerated with fiat builds.
But even though some optimizations have been, and keep being made, Fiat CI tasks require a significant amount of time and memory to complete, and I'm not sure that adding these new fields wouldn't hit the CI limits. I'm going to give it a shot.

p384/src/arithmetic/projective.rs
```diff
@@ -0,0 +1,596 @@
...
+                pc[i / 2].double()
+            } else {
+                pc[i - 1].add(self)
+            };
+        }
+        let mut q = ProjectivePoint::IDENTITY;
+        let k = k.to_bytes();
+        let mut pos = 384 - 4;
+        loop {
+            let slot = (k[(pos >> 3) as usize] >> (pos & 7)) & 0xf;
```
brycx, line 212: `(k[(pos >> 3) as usize] >> (pos & 7)) & 0xf;` -> `(k[47 - (pos >> 3) as usize] >> (pos & 7)) & 0xf;`

Just found this PR before submitting this one at @jedisct1's repository: https://github.com/jedisct1/rust-p384/pull/2

brycx (changes requested):

p384/src/arithmetic/projective.rs
jedisct1: You should import the recent commits from `jedisct1/rust-p384` instead.

tarcieri:
Imported the latest changes from @jedisct1's repo, got all of the tests passing, and wrote a real commit message.

Removed WIP/draft. PTAL.

tarcieri:
Sidebar: an important next step to this PR is getting the arithmetic working on 32-bit backends, including generating a 32-bit backend for the scalar field

jedisct1:
https://github.com/mit-plv/fiat-crypto/pull/1259 includes the 32-bit backend for the scalar field.

tarcieri:
See #566 for next steps

PR #567, opened 2022-05-27
p384: initial test vectors for group operations

tarcieri:
Adds basic test vectors for repeated additions of the generator.

Sourced from:

http://point-at-infinity.org/ecc/nisttv


p384/src/arithmetic/projective.rs
```diff
@@ -594,3 +594,120 @@ impl TryFrom<&ProjectivePoint> for PublicKey {
...
+            ProjectivePoint::from(basepoint_affine),
+            basepoint_projective,
+        );
+        assert_eq!(basepoint_projective.to_affine(), basepoint_affine);
+        assert!(!bool::from(basepoint_projective.to_affine().is_identity()));
+
+        // TODO(tarcieri): BROKEN!
+        // assert!(bool::from(
+        //     ProjectivePoint::IDENTITY.to_affine().is_identity()
+        // ));
```
tarcieri, line 619: @jedisct1 @brycx FYI, this test is failing

brycx: I think I found the issue: #568

PR #568, opened 2022-05-27
p384: `FieldElement::invert()` should check if self was non-zero

brycx:
Fix for broken test https://github.com/RustCrypto/elliptic-curves/pull/567#discussion_r883209549

Edit: The same check has been changed in `Scalar::invert()` but there was no existing test to detect this.


tarcieri:
Thank you!

PR #573, opened 2022-05-27
p384: initial 32-bit support

tarcieri:
Imports the 32-bit fiat-crypto scalar field implementation generated by @jedisct1 from this PR:

https://github.com/mit-plv/fiat-crypto/pull/1259

Gates all field backends and precomputed constants on `target_pointer_width`, substituting 12-limb 32-bit versions for 32-bit targets where necessary.

Inversions are implemented using 64-bit arithmetic, so for now they are marked as `todo!()`.


tarcieri:
@brycx @jedisct1  field and scalar inversions will need a 32-bit version for this to work properly

tarcieri:
I'm at a complete loss why it won't build with just the `arithmetic` feature on ostensibly any 32-bit target:

https://github.com/RustCrypto/elliptic-curves/runs/6630521978?check_suite_focus=true

It seems like enabling practically any other feature, like `digest`, `ecdh`, or `ecdsa-core` fixes the problem, but I have no idea why.

I'm trying to enable `elliptic-curve/digest` as a workaround. That seems to do the trick, but I'm confused why.

PR #591, opened 2022-06-03
Add RFC 6979 test vectors to `p256` and `p384` ECDSA

brycx:
One additional test for `p256` and added matching to `p384`.


tarcieri:
Thank you!

PR #599, opened 2022-06-11
Update P-256 VOPRF test vectors

daxpedda:
Adds a bit more code in the tests, which I hope is okay.


tarcieri:
What's the provenance of these test vectors and why did they change? (It would probably be good to include a link to the source)

Is this from an evolving I-D?

PR #600, opened 2022-06-11
Implement `HashToScalar` for P-384

daxpedda:
Depends on https://github.com/RustCrypto/traits/pull/1024.

### DIFF
--- a/p384/src/arithmetic/hash2curve.rs
+++ b/p384/src/arithmetic/hash2curve.rs
@@ -1,0 +1,326 @@
+use super::FieldElement;
+use crate::{AffinePoint, FieldBytes, NistP384, ProjectivePoint, Scalar};
+use elliptic_curve::{
+    bigint::{ArrayEncoding, U384},
+    consts::U72,
+    generic_array::GenericArray,
+    hash2curve::{FromOkm, GroupDigest, MapToCurve, OsswuMap, OsswuMapParams, Sgn0},
+    ops::Reduce,
+    point::DecompressPoint,
+    subtle::Choice,
+};
+
+impl GroupDigest for NistP384 {
+    type FieldElement = FieldElement;
+}
+
+impl FromOkm for FieldElement {
+    type Length = U72;
+
+    fn from_okm(data: &GenericArray<u8, Self::Length>) -> Self {
+        const F_2_288: FieldElement = FieldElement::from_be_hex(
+            "000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000",
+        );
+
+        let mut d0 = FieldBytes::default();
+        d0[12..].copy_from_slice(&data[0..36]);
+        let d0 = FieldElement::from_uint_unchecked(U384::from_be_byte_array(d0));
+
+        let mut d1 = FieldBytes::default();
+        d1[12..].copy_from_slice(&data[36..]);
+        let d1 = FieldElement::from_uint_unchecked(U384::from_be_byte_array(d1));
+
+        d0 * F_2_288 + d1
+    }
+}
+
+impl Sgn0 for FieldElement {
+    fn sgn0(&self) -> Choice {
+        self.is_odd()
+    }
+}
+
+impl OsswuMap for FieldElement {
+    const PARAMS: OsswuMapParams<Self> = OsswuMapParams {
+        c1: &[
+            0x0000_0000_3fff_ffff,
+            0xbfff_ffff_c000_0000,
+            0xffff_ffff_ffff_ffff,
+            0xffff_ffff_ffff_ffff,
+            0xffff_ffff_ffff_ffff,
+            0x3fff_ffff_ffff_ffff,
+        ],
+        c2: FieldElement::from_be_hex(
+            "019877cc1041b7555743c0ae2e3a3e61fb2aaa2e0e87ea557a563d8b598a0940d0a697a9e0b9e92cfaa314f583c9d066",
+        ),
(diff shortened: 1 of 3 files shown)


p384/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,148 @@
...
+            dst: b"HashToScalar-VOPRF08-\x00\x00\x04",
+            seed: &hex!("a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3"),
+            sk_sm: &hex!("b9ff42e68ef6f8eaa3b4d15d15ceb6f3f36b9dc332a3473d64840fc7b44626c6e70336bdecbe01d9c512b7e7d7e6af21"),
+        },
+        TestVector {
+            dst: b"HashToScalar-VOPRF08-\x01\x00\x04",
+            seed: &hex!("a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3a3"),
+            sk_sm: &hex!("42c4d1c15d27be015844404088967afe48c8ae96d4f00ce48e4d38ecabfb8feb5b748de625cdf81ab076745d6211be95"),
+        },
+    ];
```
daxpedda, line 91: I'm using test Voprf version 8 test vectors to keep it simple, as version 9 requires more code.

p384/src/arithmetic/hash2curve.rs
```diff
@@ -0,0 +1,148 @@
...
+        let mut out = [0; 72];
+
+        for (src, dst) in scalar
+            .into_limbs()
+            .into_iter()
+            .rev()
+            .zip(out.chunks_exact_mut(Limb::BYTE_SIZE))
+        {
+            dst.copy_from_slice(&src.to_be_bytes());
+        }
```
daxpedda, line 125: This is just `ArrayEncoding::to_be_byte_array`, which isn't available for `U576`.

p384/src/arithmetic/scalar.rs
```diff
@@ -35,7 +35,7 @@ impl ScalarArithmetic for NistP384 {
 /// Scalars are elements in the finite field modulo n.
 #[derive(Clone, Copy, Debug)]
 #[cfg_attr(docsrs, doc(cfg(feature = "arithmetic")))]
-pub struct Scalar(U384);
+pub struct Scalar(pub(crate) U384);
```
daxpedda, line 38: I hope this is okay.

p384/src/arithmetic/hash2curve.rs
tarcieri: I can potentially add `U576`

p384/src/arithmetic/scalar.rs
tarcieri: That's fine, although potentially avoidable if all of the constants are given in Montgomery form

p384/src/arithmetic/hash2curve.rs
tarcieri: As it turns out, `U576` is useful for `p521` as well, so I went ahead and added it:

https://github.com/RustCrypto/crypto-bigint/pull/94

tarcieri: Released in `crypto-bigint` v0.4.5

p384/src/arithmetic/hash2curve.rs
daxpedda: Now that everything's working, I updated the test vectors.

p384/src/arithmetic/hash2curve.rs
daxpedda: Amazing, that helped out a lot. How do I pull in 0.4.5 though?
- Update `elliptic-curve` to require `0.4.5`.
- Add a direct dependency to `0.4.5`.
- Only update `Cargo.lock`. That wouldn't work for downstream users, right?

p384/src/arithmetic/scalar.rs
daxpedda: Didn't need it anymore, so I reverted it.

daxpedda:
So this has to build on https://github.com/RustCrypto/traits/pull/1024 now. Considering this will be a breaking change, what's the solution here?
I could just copy-paste the OSSWU code here, like we did in [`k256`](https://github.com/RustCrypto/elliptic-curves/blob/0dd58a00593cc77952ab7101cce284c4751a78f6/k256/src/arithmetic/hash2curve.rs#L85-L127). Or should we just delay this PR? Or release an `elliptic-curve` 0.13-pre?

tarcieri:
I think it'd probably make sense to just delay the PR until the next round of breaking changes to `elliptic-curve`, and in the meantime perhaps we could work out some better solutions for https://github.com/RustCrypto/traits/pull/1024 regarding the relationships between `FieldElement`s and their inner `UInt`s.

daxpedda:
Alright, good to know!
I'm in no rush :).

daxpedda:
Updated to https://github.com/RustCrypto/traits/commit/28ec035531ed028ba010a652980f78007c8f49ff.

daxpedda:
@tarcieri this is ready now.

tarcieri (approved):

PR #611, opened 2022-06-22
Remove mention of `None` in `to_affine` docs

conradoplg:
Since `to_affine` doesn't return an `Option`, the mention of `None` in the documentation doesn't seem to make sense anymore.


tarcieri:
Thanks! We added a representation of the identity for `AffinePoint`, so indeed that comment no longer makes sense.

PR #621, opened 2022-07-07
p256: Allow converting affine points to coordinates and back

codahale:
This adds a trait `AffineCoordinates` with two methods (`from_coordinates` and `to_coordinates`) to convert field element coordinates to points and back, checking that the coordinates are on the curve. Internal conversions now use this trait, and the trait is exposed publicly if the `expose-field` flag is set.

Also re-exports the two curve constants if the `expose-field` flag is set.

These two changes would allow for a full Elligator Squared implementation (#540) entirely outside of the `p256` crate.

### DIFF
--- a/p256/src/arithmetic.rs
+++ b/p256/src/arithmetic.rs
@@ -14,13 +14,13 @@ use projective::ProjectivePoint;
 use scalar::Scalar;
 
 /// a = -3
-const CURVE_EQUATION_A: FieldElement = FieldElement::ZERO
+pub const CURVE_EQUATION_A: FieldElement = FieldElement::ZERO
     .subtract(&FieldElement::ONE)
     .subtract(&FieldElement::ONE)
     .subtract(&FieldElement::ONE);
 
 /// b = 0x5AC635D8AA3A93E7B3EBBD55769886BC651D06B0CC53B0F63BCE3C3E27D2604B
-const CURVE_EQUATION_B: FieldElement = FieldElement([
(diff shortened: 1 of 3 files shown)


str4d:
I strongly disagree with exposing `from_coordinates_unchecked` on principle; it's dangerous for security. If unchecked construction is to be permitted, I think it should only be from byte encodings (which are less likely to be accidentally manipulated).

str4d:
_Checked_ construction is a more reasonable thing to expose in a public API. I still dislike it, and would prefer it be segmented in some way so downstream users only reach for it if they really need it (as cryptographic protocols that depend on the internal coordinate details of elliptic curves are devilishly tricky to make safe in generic contexts, and really should just be implemented concretely by the concrete curve implementations, or have a different safer interface).

codahale:
I just pushed a change to check the coordinates for validity. Conditionally exposing this functionality requires extracting it as a trait, so it’s a little more verbose.

tarcieri:
@codahale this issue for adding coordinate access APIs to the `group` crate is potentially relevant here: https://github.com/zkcrypto/group/issues/30

triska:
Regarding the statement "this functionality is entirely possible with the existing public API": It seems that `FieldElement` is nowhere documented in the crate's documentation, not even in "List of all items":

https://docs.rs/k256/latest/k256/all.html

Is this really a public API?

tarcieri:
I think we'll go with whatever prospective solution there is for https://github.com/zkcrypto/group/issues/30

PR #630, opened 2022-08-02
k256: implement `ZeroizeOnDrop` for `ecdsa::SigningKey` and `schnor::SigningKey`

fjarri:
- `ecdsa::SigningKey` already has a `Drop` implemented, so just added the `ZeroizeOnDrop` marker
- For `schnor::SigningKey` implemented a zeroizing `Drop`, and added the `ZeroizeOnDrop` marker

`k256` 0.11 depends on `elliptic-curve` 0.12, which depends on `zeroize` 1.5, so this trait should be safe to use.


tarcieri:
Thank you!

PR #631, opened 2022-08-02
weierstrass: initial crate

tarcieri:
Implements `AffinePoint` and `ProjectivePoint` generically and uses it for both the `p256` and `p384` crates, with the potential to use it for `bp256`/`bp384`/`p521` as well.

## TODO
- [x] `serde` support
- [x] `PrimeField` constants
- [x] Compaction support

cc @jedisct1 @brycx


weierstrass/src/projective.rs
```diff
@@ -0,0 +1,695 @@
...
+            } else {
+                pc[i - 1].add(self)
+            };
+        }
+
+        let mut q = Self::IDENTITY;
+        let mut pos = C::UInt::BIT_SIZE - 4;
+
+        loop {
+            let slot = (k[(pos >> 3) as usize] >> (pos & 7)) & 0xf;
```
tarcieri, line 198: This logic is slightly different between `p256` and `p384`, and I'm not quite clear how to make it generic.

The version above is from `p384`. The `p256` version looks like this:

```rust
let slot = (k[31 - (pos >> 3) as usize] >> (pos & 7)) & 0xf;
```

tarcieri: Aah, it was the endianness of the serialized scalar. I've went with little endian, which makes things slightly simpler.

p256/src/arithmetic/field.rs
```diff
@@ -452,6 +445,34 @@ impl Field for FieldElement {
...
+    }
+
+    fn multiplicative_generator() -> Self {
+        todo!()
+    }
+
+    fn root_of_unity() -> Self {
+        todo!()
+    }
+}
```
tarcieri, line 474: Would be good to get the correct constant definitions here

p256/src/arithmetic/affine.rs
```diff
@@ -1,500 +0,0 @@
...
-                let recovered_y = if borrow == 0 { y } else { p_y };
-                AffinePoint {
-                    x,
-                    y: recovered_y.to_montgomery(),
-                    infinity: 0,
-                }
-            })
-        })
-    }
-}
```
tarcieri, line 242: One regression: I didn't port this across yet. It seems like it could be impl'd generically, however

p384/src/arithmetic/field.rs
```diff
@@ -122,6 +123,40 @@ impl FieldElement {
...
+    }
+
+    fn multiplicative_generator() -> Self {
+        todo!()
+    }
+
+    fn root_of_unity() -> Self {
+        todo!()
+    }
+}
```
tarcieri, line 158: Would be good to get the correct constant definitions here

p384/tests/projective.rs
```diff
@@ -0,0 +1,142 @@
...
+        let point = $actual.to_affine().to_encoded_point(false);
+        let (actual_x, actual_y) = match point.coordinates() {
+            sec1::Coordinates::Uncompressed { x, y } => (x, y),
+            _ => unreachable!(),
+        };
+
+        assert_eq!(&expected_x, actual_x.as_slice());
+        assert_eq!(&expected_y, actual_y.as_slice());
+    };
+}
```
tarcieri, line 29: @str4d workaround I ended up with for comparing curve points to test vectors.

I suppose I can just change the structure of the test vectors to include SEC1-encoded points.

Issue #642, opened 2022-08-21
NIST P-256 (secp256r1) - Public key recovery from signature


RomarQ:
Are there any plans to also add `recover_ecdsa` on [p256](https://docs.rs/p256/latest/p256) crate?

PR #643, opened 2022-08-28
feat(ecdsa/secp256r1): Support public key recovery from signature

RomarQ:
This merge request adds support for recovering the public key from a signature. Most of the new code is identical to the `secp256k1` implementation. Only the `try_sign_prehashed` method of `SignPrimitive<NistP256>` trait was slightly modified to produce valid Nist P-256 (`s`, `v`) values.

Also, it adds a public method `is_y_odd` to AffinePoint type in `weierstrass` crate to avoid making the `y` coordinate public.

Issue: #642

### DIFF
--- a/p256/src/ecdsa/recoverable.rs
+++ b/p256/src/ecdsa/recoverable.rs
@@ -1,0 +1,382 @@
+//! Ethereum-style "recoverable signatures".
+//!
+//! These signatures include an additional [`Id`] field which allows for
+//! recovery of the [`VerifyingKey`] which can be used to verify them.
+//!
+//! This is helpful in cases where a hash/fingerprint of a [`VerifyingKey`]
+//! for a given signature in known in advance.
+//!
+//! ## Signing/Recovery Example
+//!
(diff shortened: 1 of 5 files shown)


p256/src/ecdsa/recoverable.rs
```diff
@@ -0,0 +1,382 @@
+//! Ethereum-style "recoverable signatures".
```
tarcieri, line 1: This is an Ethereum-specific signature format, and one which I would not encourage using in other applications.

tarcieri:
A few general thoughts:

- This is probably best implemented generically in the `ecdsa` crate, rather than copying and pasting the implementation into the crate for a concrete curve
- It would probably make sense to add static methods to `VerifyingKey`, e.g. `VerifyingKey::recover_from_signature`
- Instead of using Ethereum's signature format, it can accept an `ecdsa::Signature` as well as an `ecdsa::RecoveryId`

tarcieri:
I opened a tracking issue for a generic implementation: https://github.com/RustCrypto/signatures/issues/525

PR #650, opened 2022-09-13
[k256] Make Proj point equality and is_identity faster

LLFourn:
1. `ProjectivePoint::is_identity` was slow because it did an equality check with identity rather than just checking z coordinate == 0
2. Equality check was much slower than necessary because it converted to affine coords first. This algorithm does the check avoiding a modular inversion at the cost of a few multiplications. Around ~70 times faster.

I believe these changes are correct but please review carefully.

### DIFF
--- a/k256/src/arithmetic/projective.rs
+++ b/k256/src/arithmetic/projective.rs
@@ -279,7 +279,28 @@ impl ConditionallySelectable for ProjectivePoint {
 
 impl ConstantTimeEq for ProjectivePoint {
     fn ct_eq(&self, other: &Self) -> Choice {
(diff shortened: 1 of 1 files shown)


LLFourn:
I added another commit which adds a an fast equality check between projective and affine under the same reasoning.

tarcieri:
Thank you! I'll try to review this soon.

tarcieri:
@fjarri might also be able to provide some input here

fjarri:
Looks good to me, but hopefully someone else can review too. Thanks for looking into this.

fjarri (approved):

PR #662, opened 2022-09-24
resolve #661

### LINKED ISSUE #661: k256: How to get Uncompressed size of VerifiyingKey (PublicKey)
How to get the Uncompressed size of VerifyingKey(PublicKey) from the given SigningKey(PrivateKey), The default size of the VerifiyingKey is `33 byets`, but I need the uncompressed one e.g 64 or 65 bytes of the VerifiyingKey.

curiousdev04:
Adding `to_bytes_uncompressed` on `k256::ecdsa::VerifyingKey`

### DIFF
--- a/k256/Cargo.toml
+++ b/k256/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "k256"
-version = "0.11.5"
+version = "0.11.6"
 description = """
 secp256k1 elliptic curve library written in pure Rust with support for ECDSA
 signing/verification (including Ethereum-style signatures with public-key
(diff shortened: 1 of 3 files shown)


tarcieri:
As I mentioned on #661, the [`ToEncodedPoint`](https://docs.rs/k256/latest/k256/ecdsa/struct.VerifyingKey.html#impl-ToEncodedPoint%3CSecp256k1%3E-for-VerifyingKey) trait can be used for this.

In the next breaking release I'd like to unify the types in `k256::ecdsa` with the ones in the `ecdsa` crate, so we can't add additional methods like this unless the corresponding ones are added upstream.

PR #675, opened 2022-11-06
k256: remove `ecdsa::recoverable` module; add replacement docs

tarcieri:
Support for recovering `VerifyingKey`s from signatures has been upstreamed into the `ecdsa` crate in RustCrypto/signatures#576.

This commit removes the `ecdsa::recoverable` module and the custom impls of `SigningKey`/`VerifyingKey` needed to support it.

In its places are docs about how to use the generic APIs to accomplish equivalent functionality.


tarcieri:
Version bump of what?

The new API will ship in `k256` v0.12 which will use `ecdsa` v0.15

PR #690, opened 2022-11-20
k256: add `schnorr::SigningKey::as_nonzero_scalar`

tarcieri:
Adds an accessor for the inner `NonZeroScalar` value that represents a Schnorr secret key.


mikedilger:
Thanks!

PR #697, opened 2022-12-30
Scalar multiplication by the generator

fjarri:
This PR introduces a way to multiply a scalar by the generator by keeping precomputed lookup tables. It is exported as `mul_by_generator()` (and used in the signing code). It takes 40us compared to 86us for regular multiplication; signing takes 70us compared to 116us.

The precomputed table contains values for every second radix and takes ~30kb; if we calculated it for every radix it would take 60kb with the performance improvement of only ~3%, so I don't think it is worth it (the same choice was taken by https://github.com/dalek-cryptography/curve25519-dalek).

I am not sure how to speed up the verification, which relies on calculating `k1 * G + k2 * x`. Pre-calculating lookup tables for `G` in `lincomb` only speeds it up from 129us to 122us on my machine. Calculating `k1 * G` via `mul_by_generator()` results in the timing of 126us. Not great, any ideas are welcome.

The precomputed tables use `once_cell`. Note that the feature `critical-section` of `once_cell` (which we need to use to support `no_std`) has the MSRV of 1.63. Proper gating is up to @tarcieri

### DIFF
--- a/k256/Cargo.toml
+++ b/k256/Cargo.toml
@@ -20,6 +20,7 @@ rust-version = "1.60"
 [dependencies]
 cfg-if = "1.0"
 elliptic-curve = { version = "0.12.3", default-features = false, features = ["hazmat", "sec1"] }
+once_cell = { version = "1.16", default-features = false, features = ["critical-section"] }
 
 # optional dependencies
 ecdsa-core = { version = "=0.15.0-rc.1", package = "ecdsa", optional = true, default-features = false, features = ["der"] }
--- a/k256/src/arithmetic/mul.rs
+++ b/k256/src/arithmetic/mul.rs
@@ -44,6 +44,7 @@ use elliptic_curve::{
     subtle::{Choice, ConditionallySelectable, ConstantTimeEq},
     IsHigh,
 };
+use once_cell::sync::Lazy;
 
 /// Lookup table containing precomputed values `[p, 2p, 3p, ..., 8p]`
 #[derive(Copy, Clone, Default)]
@@ -61,7 +62,7 @@ impl From<&ProjectivePoint> for LookupTable {
 
 impl LookupTable {
     /// Given -8 <= x <= 8, returns x * p in constant time.
-    pub fn select(&self, x: i8) -> ProjectivePoint {
+    fn select(&self, x: i8) -> ProjectivePoint {
         debug_assert!(x >= -8);
         debug_assert!(x <= 8);
 
@@ -230,30 +231,30 @@ fn decompose_scalar(k: &Scalar) -> (Scalar, Scalar) {
 // (required because it's used in static_map later)
 // Otherwise we could just have a function returning an array.
 #[derive(Copy, Clone)]
-struct Radix16Decomposition([i8; 33]);
+struct Radix16Decomposition<const D: usize>([i8; D]);
 
-impl Radix16Decomposition {
+impl<const D: usize> Radix16Decomposition<D> {
     /// Returns an object containing a decomposition
-    /// `[a_0, ..., a_32]` such that `sum(a_j * 2^(j * 4)) == x`,
+    /// `[a_0, ..., a_D]` such that `sum(a_j * 2^(j * 4)) == x`,
     /// and `-8 <= a_j <= 7`.
-    /// Assumes `x < 2^128`.
+    /// Assumes `x < 2^(4*(D-1))`.
     fn new(x: &Scalar) -> Self {
-        debug_assert!((x >> 128).is_zero().unwrap_u8() == 1);
+        debug_assert!((x >> (4 * (D - 1))).is_zero().unwrap_u8() == 1);
 
         // The resulting decomposition can be negative, so, despite the limit on `x`,
-        // it can have up to 256 bits, and we need an additional byte to store the carry.
-        let mut output = [0i8; 33];
+        // we need an additional byte to store the carry.
+        let mut output = [0i8; D];
 
         // Step 1: change radix.
         // Convert from radix 256 (bytes) to radix 16 (nibbles)
         let bytes = x.to_bytes();
-        for i in 0..16 {
+        for i in 0..(D - 1) / 2 {
             output[2 * i] = (bytes[31 - i] & 0xf) as i8;
             output[2 * i + 1] = ((bytes[31 - i] >> 4) & 0xf) as i8;
         }
 
         // Step 2: recenter coefficients from [0,16) to [-8,8)
-        for i in 0..32 {
+        for i in 0..(D - 1) {
             let carry = (output[i] + 8) >> 4;
             output[i] -= carry << 4;
             output[i + 1] += carry;
@@ -263,9 +264,9 @@ impl Radix16Decomposition {
     }
 }
 
-impl Default for Radix16Decomposition {
+impl<const D: usize> Default for Radix16Decomposition<D> {
     fn default() -> Self {
-        Self([0i8; 33])
+        Self([0i8; D])
     }
 }
 
@@ -342,14 +343,14 @@ fn lincomb_generic<const N: usize>(xs: &[ProjectivePoint; N], ks: &[Scalar; N]) 
     );
 
     let digits1 = static_map(
-        |r| Radix16Decomposition::new(&r),
+        |r| Radix16Decomposition::<33>::new(&r),
         &r1s_c,
-        Radix16Decomposition::default(),
+        Radix16Decomposition::<33>::default(),
     );
     let digits2 = static_map(
-        |r| Radix16Decomposition::new(&r),
+        |r| Radix16Decomposition::<33>::new(&r),
         &r2s_c,
-        Radix16Decomposition::default(),
+        Radix16Decomposition::<33>::default(),
     );
 
     let mut acc = ProjectivePoint::IDENTITY;
@@ -369,6 +370,41 @@ fn lincomb_generic<const N: usize>(xs: &[ProjectivePoint; N], ks: &[Scalar; N]) 
         }
     }
     acc
+}
+
+static GEN_LOOKUP_TABLE: Lazy<[LookupTable; 33]> = Lazy::new(precompute_gen_lookup_table);
+
+fn precompute_gen_lookup_table() -> [LookupTable; 33] {
+    let mut gen = ProjectivePoint::GENERATOR;
+    let mut res = [LookupTable::default(); 33];
+    #[allow(clippy::needless_range_loop)]
+    for i in 0..33 {
+        res[i] = LookupTable::from(&gen);
+        // We are storing tables spaced by two radix steps,
+        // to decrease the size of the precomputed data.
+        for _ in 0..8 {
+            gen = gen.double();
+        }
+    }
+    res
+}
+
+/// Calculages `k * G`, where `G` is the generator.
+pub fn mul_by_generator(k: &Scalar) -> ProjectivePoint {
+    let digits = Radix16Decomposition::<65>::new(k);
+    let table = *GEN_LOOKUP_TABLE;
+    let mut acc = table[32].select(digits.0[64]);
+    let mut acc2 = ProjectivePoint::IDENTITY;
+    for i in (0..32).rev() {
+        acc2 += &table[i].select(digits.0[i * 2 + 1]);
+        acc += &table[i].select(digits.0[i * 2]);
+    }
+    // This is the price of halving the precomputed table size (from 60kb to 30kb)
+    // The performance hit is minor, about 3%.
+    for _ in 0..4 {
+        acc2 = acc2.double();
+    }
+    acc + acc2
 }
 
 #[inline(always)]
@@ -425,6 +461,7 @@ impl MulAssign<&Scalar> for ProjectivePoint {
 
 #[cfg(test)]
 mod tests {
+    use super::mul_by_generator;
     use crate::arithmetic::{ProjectivePoint, Scalar};
     use elliptic_curve::{ops::LinearCombination, rand_core::OsRng, Field, Group};
 
@@ -437,6 +474,14 @@ mod tests {
 
         let reference = &x * &k + &y * &l;
         let test = ProjectivePoint::lincomb(&x, &k, &y, &l);
+        assert_eq!(reference, test);
+    }
+
+    #[test]
+    fn test_mul_by_generator() {
+        let k = Scalar::random(&mut OsRng);
+        let reference = &ProjectivePoint::GENERATOR * &k;
+        let test = mul_by_generator(&k);
         assert_eq!(reference, test);
     }
 }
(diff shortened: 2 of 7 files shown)


tarcieri:
Thanks for contributing this! I'm working on a somewhat analogous problem over here:

https://github.com/dalek-cryptography/ed25519-dalek/pull/251

I agree it would be good to have a trait for fixed-base scalar multiplication / multiply-by-generator, particularly so basepoint tables can be enabled conditionally as a performance-vs-code-size tradeoff while encapsulating the details of whether or not basepoint tables are used.

fjarri:
It is more of a startup time than code size, since I'm using `lazy_static`.

But that actually gives me an idea, I can also put the `g * 2^(i * window)` into a table so that I can get rid of the doubling.

str4d:
You may also want to look at the types in the `group` crate under `group::wnaf`, which also provide APIs for precomputing scalar multiplication tables.

k256/Cargo.toml
```diff
@@ -20,6 +20,7 @@ rust-version = "1.57"
 [dependencies]
 cfg-if = "1.0"
 elliptic-curve = { version = "0.12.3", default-features = false, features = ["hazmat", "sec1"] }
+lazy_static = { version = "1.4", features = ["spin_no_std"] }
```
tarcieri, line 23: I generally prefer `once_cell` these days, especially as it's [slated for eventual inclusion in `core`/`std`](https://github.com/rust-lang/rust/issues/74465).

The `spin_no_std` "feature" of `lazy_static` was always weirdly non-additive too. Enabling it enables it for all crates that use `lazy_static`.

tarcieri:
@str4d speaking of `group`, it seems like `mul_by_generator` might be a useful addition there? https://github.com/zkcrypto/group/pull/44

k256/Cargo.toml
fjarri: Changed to `once_cell`. I am not 100% sure I'm using it correctly, so if you have experience with it, please check.

fjarri:
- replaced `lazy_static` with `once_cell`
- separated `mul_by_generator` into its own module, since it doesn't need to use endomorphism
- added more precalculated tables

So now `mul_by_generator()` takes 40us compared to 86us for regular multiplication; signing takes 70us compared to 116us. Verification would be improved too but less so, since we will lose the advantage of calculating linear combination.

Also I cannot change the verification code in this PR since it's `signatures/ecdsa` and not in `k256`.

I guess this PR is blocked by https://github.com/zkcrypto/group/pull/44; after that's merged and published, this can be updated too, and then `signatures/ecdsa` to use it for verification.

fjarri:
As for using wNAF, I am not certain, won't it make the multiplication variable-time?

k256/Cargo.toml
fjarri: It seems that to use `once_cell` in `no_std` environment we need the `critical_section` feature, but when such a library is used with `std` enabled, there is a linker error that can be prevented by specifying `critical-section = { version = "1.1", features = ["std"] }` explicitly. This is kind of a bummer. Is there a better way?

fjarri: Also `once_cell` has the MSRV 1.63, so this crate's MSRV will have to be updated.

tarcieri: Huh? `once_cell` should be MSRV 1.56

> This is kind of a bummer. Is there a better way?

`no_std` users who want the basepoint tables can link `once_cell` and enable the `critical_section` feature themselves.

But I suspect many embedded will not basepoint tables at all due to limitations on the size of the loaded program. In that regard it would probably be good to feature gate them ala https://github.com/dalek-cryptography/curve25519-dalek/pull/489

Since you have a `mul_by_generator` function it can fallback to a slower implementation when the basepoint tables aren't available, but still provide a common API for either case.

fjarri: > Huh? once_cell should be MSRV 1.56

With no features, yes, but `critical-section` requires 1.63.

> Due to limitations on the size of the loaded program

It is a lazily evaluated value, so the limitation will be on the memory/cache usage. Although adding the evaluated lookup tables in the code itself (like `ed25519` does it) is another possibility, but it will still consume cache on usage.

> Since you have a mul_by_generator function it can fallback to a slower implementation when the basepoint tables aren't available, but still provide a common API for either case.

I thought the common API would be provided by the trait, which could use `mul_by_generator()` or `mul()` (in the default implementation)

tarcieri:
@fjarri I wouldn't say this PR is blocked on https://github.com/zkcrypto/group/pull/44

You can make `mul_by_generator` an inherent method for now, and then we can potentially circle back on getting it added to the trait.

k256/Cargo.toml
fjarri: > no_std users who want the basepoint tables can link once_cell and enable the critical_section feature themselves.

Not sure I would like that as a user. Seems pretty brittle.

tarcieri: > It is a lazily evaluated value, so the limitation will be on the memory/cache usage. 

Well yes, but that's what I mean by "size of the loaded program". It may need to fit in e.g. SRAM, on a device with no DRAM.

Lazy or not, at the end of the day it's a static table that needs to fit in whatever RAM the device has available, and many embedded devices are highly constrained.

As a data point, the `ed25519-dalek` basepoint tables were too big to fit into some ARM Cortex M0/M4 devices people were experimenting with at OxidizeConf when I attended it a few years ago.

> I thought the common API would be provided by the trait, which could use mul_by_generator() or mul() (in the default implementation)

We can use the trait when it's available, which provides a default `mul`-based implementation, and would let you conditionally override it in the impl block to use the basepoint tables.

But absent that upstream feature in `group`, you can do something like this with an inherent `mul_by_generator()` method:

https://github.com/dalek-cryptography/curve25519-dalek/pull/489/files#diff-0dbe5da3c59af6d6e20fbb1c92b248a7bcce0a5e15f0e5de6d16c9a60abe3ee3R705-R722

fjarri:
I guess I'll focus on providing the optimized logic in this PR, and leave the issues of gating and `once_cell` usage to you :)

k256/Cargo.toml
tarcieri: > Not sure I would like that as a user. Seems pretty brittle.

Hopefully the ergonomic concerns here go away when it's merged upstream into `core`/`std`. Really that's the main reason to prefer it: it will soon hopefully be part of the standard library.

The `spin_no_std` feature of `lazy_static` is not a great solution here either. If one dependency in your `std`-dependent program activates it, suddenly every lazy static becomes spinlocked. And for that reason I think it's bad for dependencies to enable it rather than the toplevel `[bin]` crate.

tarcieri:
Sure, that's fine. Happy to do a followup which adds a `basepoint-tables` feature along the lines of https://github.com/dalek-cryptography/curve25519-dalek/pull/489

tarcieri:
> Also I cannot change the verification code in this PR since it's in signatures/ecdsa and not in k256.

You can still change the verification implementation if you like: https://github.com/RustCrypto/elliptic-curves/blob/8ffa72c/k256/src/ecdsa.rs#L241

It's currently using the provided implementation of `verify_prehashed`:

https://docs.rs/ecdsa/0.15.0-pre.1/ecdsa/hazmat/trait.VerifyPrimitive.html#method.verify_prehashed

However if you copy/paste that implementation into the `VerifyPrimitive` impl you can tweak it however you'd like.

fjarri:
It's actually not necessary anymore, since I currently don't know how to speed up verification (see the updated top message).

tarcieri:
Yeah, seems like the two approaches are mutually incompatible.

It might be worth checking if the speedup provided by precomputed basepoint tables outperforms using linear combinations, though that said, being able to use the provided implementation of ECDSA from the `ecdsa` crate is also nice.

fjarri:
> It might be worth checking if the speedup provided by precomputed basepoint tables outperforms using linear combinations

I did check it (if I understand you correctly), see the timings in the first post

tarcieri:
I meant specifically in regard to verification, where you said this:

> Verification would be improved too but less so, since we will lose the advantage of calculating linear combination. 

...which implies it would still benefit from basepoint tables over linear combinations, but I wasn't clear how much.

fjarri:
Yes, at the moment I wrote this I was only guessing, but then I actually measured different approaches and updated the top post. Sorry for a bit of a non-linear post history.

tarcieri:
@fjarri here's some semi-related code from curve25519-dalek, if it helps: https://github.com/dalek-cryptography/curve25519-dalek/blob/e01bb1b/src/backend/serial/scalar_mul/vartime_double_base.rs#L22-L64

fjarri:
That's basically the approach I tried in here:
> Pre-calculating lookup tables for `G` in `lincomb` only speeds it up from 129us to 122us on my machine.

Of course, that was still constant-time, using regular windows. They also use wNAF, since the verification does not have to be constant-time, which is an orthogonal change. It may give more speed-up. I don't think it should be done in this PR though.

tarcieri:
That's cool.

Per @str4d's comment, I've been curious if the wNAF functionality in `group::wnaf` could assist here.

fjarri:
@tarcieri , to use `group::Wnaf` I'll need `group = 0.13`, which requires `elliptic-curve = 0.13` to be released and its version bumped here.

tarcieri:
Duly noted. I'll try to wrap up another release cycle soon so we can switch `master` over to `elliptic-curve` v0.13 prereleases

tarcieri (approved):
Speaking of which, it would be nice to get this PR into that set of releases.

@fjarri it seems like this should be good to go other than feature-gating it? (which I can do in a followup)

fjarri:
> Speaking of which, it would be nice to get this PR into that set of releases.

I was thinking the same, but it seems that I will need those version bumps first. The `ecdsa` dep will have to be bumped too since ecdsa=0.15.0-rc.1 depends on elliptic-curve=0.12 . 

> @fjarri it seems like this should be good to go other than feature-gating it? (which I can do in a followup)

Yes, should be fine.

tarcieri:
> I was thinking the same, but it seems that I will need those version bumps first.

What do you need bumped specifically?

> The ecdsa dep will have to be bumped too since ecdsa=0.15.0-rc.1 depends on elliptic-curve=0.12 .

...which is the version currently in use? Or are you talking about for adding wNAF?

fjarri:
> Or are you talking about for adding wNAF?

Yes, for wNAF we need the new `group` which means the new `ecdsa` and `elliptic-curve`. For this PR no bumps are necessary.

k256/src/arithmetic/mul.rs
```diff
@@ -371,6 +372,41 @@ fn lincomb_generic<const N: usize>(xs: &[ProjectivePoint; N], ks: &[Scalar; N])
...
+        acc2 += &table[i].select(digits.0[i * 2 + 1]);
+        acc += &table[i].select(digits.0[i * 2]);
+    }
+    // This is the price of halving the precomputed table size (from 60kb to 30kb)
+    // The performance hit is minor, about 3%.
+    for _ in 0..4 {
+        acc2 = acc2.double();
+    }
+    acc + acc2
+}
```
tarcieri, line 408: @fjarri one thing I noticed too late refactoring... I think this would make sense as an inherent method of `ProjectivePoint` so it can be called as `ProjectivePoint::mul_by_generator(&scalar)`?

That way you don't have to import it separately.

fjarri: Yes, I think I put it into `mul.rs` to make use of `Radix16Decomposition` and `LookupTable`, but it can be definitely defined in an `impl` block there.

tarcieri: Cool, I think I'll go ahead and refactor it into an inherent method in #705 then

k256/Cargo.toml
```diff
@@ -20,6 +20,7 @@ rust-version = "1.60"
 [dependencies]
 cfg-if = "1.0"
 elliptic-curve = { version = "0.12.3", default-features = false, features = ["hazmat", "sec1"] }
+once_cell = { version = "1.16", default-features = false, features = ["critical-section"] }
```
matklad, line 23: :thinking: not sure that enabling critical-section _here_ is the right call. I'd imagine it should be up to the leaf crate with Cargo.lock to enable this, such that pulling this crate doesn't pull cs unconditionally. 

One imporovement here would be to add `std = "once-cell/std"`, so that once_cell doesn't use critical section (Cargo would still download and compile this crate, it'll just be unused). 

But I think what you should do is:

```
[dependencies]
once_cell = { version = "1.16", default-features = false }

[features]
default = ["arithmetic", "ecdsa", "pkcs8", "schnorr", "std"]
alloc = ["ecdsa-core?/alloc", "elliptic-curve/alloc"]

# Either std or critical-section should be enabled
std = ["once-cell/std", "alloc", "ecdsa-core?/std", "elliptic-curve/std"]
critical-section = ["once-cell/critical-section"]
```

That way:

* people with `std` don't get `critical_section` at all
* people with `no_std` would have to explicitly opt-into critical section
  * that's a bit bad because it doesn't "just work"
  * OTOH, silently opting the users in critical-section behavior feels not super friendly. It's a big hammer, better to be explicitly aware about it!

tarcieri: Sounds good to me. Want to make a PR?

FWIW, the situation is probably not as bad as it seems, because I expect most embedded users will want to disable `precomputed-tables` anyway to reduce the size of the library in memory

matklad: > Want to make a PR?

Not really :) 

>FWIW, the situation is probably not as bad as it seems

This is a bit ambiguous, but, with the current code, my main worry is std people who get opted into critical-section. I am not sure how exactly critical-section behaves on std, but at minimum that hits much less tested and much slower path of once_cell, _for all once cells in the final binary_, and, at maximum, that actually serializes access to all once cells everywhere :fearful:

tarcieri: Heh ok, I'll make a PR

fjarri: >  for all once cells in the final binary

I guess the `once_cell` API has the same problems as `lazy_static`. Big yikes. Hopefully it won't be the case when it's a part of the language proper.

matklad: std won't have this problem by virtue of not supporting no_std at all. If you want to configure synchronisation for a _specific_ once_cell, rather than for all of them globally, https://crates.io/crates/generic_once_cell is the right crate.

fjarri: Ah, so it's going to be std-only? That's a bummer.

tarcieri: @fjarri with an `alloc` dependency it would be possible to use `race::OnceBox`

k256/Cargo.toml
tarcieri: @fjarri another option is to transform the precomputed tables into Rust literal syntax, and use that to define a static. That's what the dalek crates do:

https://github.com/dalek-cryptography/curve25519-dalek/blob/3effd73307a606e44469a425974f0b7b0eb85899/src/backend/serial/u64/constants.rs#L328

You can test if it diverges by running the precomputation again and seeing if it equals the static value.

fjarri: That will lead to increased code size which may be less preferable than some start-up time, e.g. in WASM context.

PR #705, opened 2023-01-09
k256: add `basepoint-tables` feature

tarcieri:
This allows developers concerned with the resident size of the program to avoid the large `GEN_LOOKUP_TABLE` static.

### DIFF
--- a/k256/Cargo.toml
+++ b/k256/Cargo.toml
@@ -20,9 +20,9 @@ rust-version = "1.60"
 [dependencies]
 cfg-if = "1.0"
 elliptic-curve = { version = "0.12.3", default-features = false, features = ["hazmat", "sec1"] }
-once_cell = { version = "1.16", default-features = false, features = ["critical-section"] }
 
 # optional dependencies
+once_cell = { version = "1.16", optional = true, default-features = false, features = ["critical-section"] }
 ecdsa-core = { version = "=0.15.0-rc.1", package = "ecdsa", optional = true, default-features = false, features = ["der"] }
 hex-literal = { version = "0.3", optional = true }
 serdect = { version = "0.1", optional = true, default-features = false }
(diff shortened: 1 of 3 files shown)


fjarri:
I just have a minor comment: do you think the usage of both "basepoint" and "generator" can be confusing? Should we pick one?

k256/Cargo.toml
```diff
@@ -41,11 +41,12 @@ rand_core = { version = "0.6", features = ["getrandom"] }
 sha3 = { version = "0.10", default-features = false }
 
 [features]
-default = ["arithmetic", "ecdsa", "pkcs8", "schnorr", "std"]
+default = ["arithmetic", "basepoint-tables", "ecdsa", "once_cell", "pkcs8", "schnorr", "std"]
```
fjarri, line 44: "once_cell" would be already enabled by "basepoint-tables", right?

tarcieri: Good call!

tarcieri:
@fjarri yeah it's tricky... I was trying to shoot for some consistency with `curve25519-dalek` where the feature is called that, but I would say with the zkcrypto/group and RustCrypto/elliptic-curves ecosystems it's "generator".

That said, I've always heard "basepoint tables" and never "generator point tables" or "generator tables".

tarcieri:
Going to merge this to open a PR which refactors `mul_by_generator`, but feel free to suggest some alternative feature names and I can potentially follow up on a rename.

fjarri:
I would personally go with "generator" everywhere since it's the mathematical term, and also the terminology that `group::Group` uses. "Basepoint" seems like a term specific to ECC, and I don't think I've seen it used much outside of ed25519 context. So in order of preference:
- "generator" everywhere
- "basepoint" everywhere
- both

But it's up to you.

tarcieri:
I'm fine with `generator` everywhere, so I guess call this feature `generator-table`(s)?

fjarri:
Yes, or possibly `precomputed-tables`? Unless you want to be able to have fine-grained control on which tables are used.

tarcieri:
Ooh, that's a great name!

PR #708, opened 2023-01-09
Vartime point-scalar multiplication 

fjarri:
@tarcieri - the first commit here contains compatibility fixes for group==0.13 - you can use it if you're going to bump the versions (won't build in CI until `elliptic-curve` and `ecdsa` are updated)


tarcieri:
@fjarri seems like this can be rebased now?

fjarri:
This seems obsolete, the support for `ff=0.13` is already in. Closing.

PR #709, opened 2023-01-12
Constructors for scalars from u128

nickray:
These are helpful when using the scalars in Shamir secret sharing with random (128 bit) participants.

I didn't touch the field.rs files, as then automatic method selection in the `multiplicative_generator` implementations fails - so there is potentially a danger of backwards incompat in p256+p384 which don't have `From<u32>`.

Personally, I'm only interested in `From<u128>` for `k256`.


tarcieri:
It seems this breaks inference for integer conversions where the type isn't annotated, so you'll need to annotate the types where they weren't annotated before

nickray:
Right, I missed the 32-bit build test.

tarcieri:
Thanks!

PR #710, opened 2023-01-13
k256: don't eagerly enable `once_cell`'s `critical-section` feature

tarcieri:
Instead this uses weak feature activation to enable `once_cell/std` when the `std` feature is enabled, which it is by default.

The `critical-section` feature can be used in cases where `std` is unavailable but `precomputed-tables` are desired.

### DIFF
--- a/k256/src/arithmetic/mul.rs
+++ b/k256/src/arithmetic/mul.rs
@@ -34,6 +34,12 @@
 //! (Note that 'd' is also equal to the curve order here because `[a1,b1]` and `[a2,b2]` are found
 //! as outputs of the Extended Euclidean Algorithm on inputs 'order' and 'lambda').
 
+#[cfg(all(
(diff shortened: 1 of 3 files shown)


k256/src/arithmetic/mul.rs
```diff
@@ -34,6 +34,12 @@
 //! (Note that 'd' is also equal to the curve order here because `[a1,b1]` and `[a2,b2]` are found
 //! as outputs of the Extended Euclidean Algorithm on inputs 'order' and 'lambda').
 
+#[cfg(all(
+    feature = "precomputed-tables",
+    not(any(feature = "critical-section", feature = "std"))
+))]
+compile_error!("`precomputed-tables` feature requires either `critical-section` or `std`");
```
tarcieri, line 41: This will hopefully make the `critical-section` feature discoverable

PR #729, opened 2023-01-27
primeorder: support curves with any `a`

### LINKED ISSUE #726: primeorder: supports only curves with `a = -3`
`primeorder` docs say that crate provides generic implementation of elliptic curve arithmetic for curves defined as

```
y² = x³ + ax + b
```

However, I can see that [`ProjectivePoint::double`](https://github.com/RustCrypto/elliptic-curves/blob/f6ffb4363932a17e05c86678b672ceb54d14bc9c/primeorder/src/projective.rs#L140) is implementation of [`Renes-Costello-Batina 2015 (Algorithm 6)`](https://eprint.iacr.org/2015/1060.pdf). Paper says that Algorithm 6 is exception-free point doubling for prime order short Weierstrass curves $E/\mathbb{F}_q : y^2 = x^3 + a x + b$ **_with_** $a = −3$.

Do you consider eventually switching to more generic algorithms, like algorithm 3 from the same paper which doesn't require $a$ to be $-3$?  In any case, worth mentioning in the docs that atm only $a=-3$ curves are supported to avoid confusion.

survived:
Closes #726


tarcieri:
@survived can you rebase now that I landed #728?

I'd suggest getting rid of the inherent methods for `ProjectivePoint::add` and use overlapping `Add` impls for the different `EquationAProperties` ZSTs.

survived:
Sure! I just wanted to finish algorithms implementation first and see if it works fine. I'll mark PR as ready for review when I'm done :)

tarcieri:
FWIW you should be able to use the implementations @newpavlov mentioned in #726, e.g. https://github.com/RustCrypto/elliptic-curves/pull/218/files#diff-cf2092dccb9b015c6defd5f649e6449e52e4cef9b10c77e2a96d6201033c8232R170

survived:
If I'm not missing anything, all algos are already implemented, so I only left to rebase it now and use trait approach. But I'll keep in mind that there's another implementation, thanks!

tarcieri:
@survived yeah, rebase on `master` and copy-and-paste the implementations into `impl`s for the respective traits (`Add` and `Double`)

Should be fairly straightforward

tarcieri:
Aah, unfortunate.

Another option is moving the implementation to methods on the ZSTs and having e.g. `Double` call that.

survived:
Indeed that should work! Let me try it
    	
    
  

 From: Tony Arcieri ***@***.***>Sent: Monday, January 30, 2023 5:04 PMTo: RustCrypto/elliptic-curves ***@***.***>Cc: Denis Varlakov ***@***.***>; Mention ***@***.***>Subject: Re: [RustCrypto/elliptic-curves] primeorder: support curves with any `a` (PR #729) 
Aah, unfortunate.
Another option is moving the implementation to methods on the ZSTs and having e.g. Double call that

survived:
@tarcieri I came up with similar ZST-based approach, [check it out](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8bdb6402ac2a8b50034d06f2c743e5c8). I added `type PointArithmetic: PointArithmetic<C>` associated type to the `PrimeCurveParams` trait, and new `PointArithmetic<C>` trait which basically defines `double`, `add`, and `add_mixed` functions. Then we don't actually need to change interface of `ProjectivePoint`, `ProjectivePoint::{double, add, add_mixed}` would just use `PointArithmetic::{double, add, add_mixed}`. What do you think?

tarcieri:
Looks good overall! That's more or less what I had in mind.

survived:
Do you want to keep `Double` trait, or shall I revert this change?

tarcieri:
I already added it to `elliptic-curve`, so seems OK to keep. I'll switch to that trait in #730

survived:
@tarcieri PR is ready for review!

primeorder/src/point_arithmetic.rs
```diff
@@ -0,0 +1,327 @@
...
+/// Provides implementation of point arithmetic (point addition, point doubling) which
+/// might be optimized for the curve.
+pub trait PointArithmetic<C: PrimeCurveParams>: sealed::Sealed {
+    /// Returns `point + point`
+    fn double(point: &ProjectivePoint<C>) -> ProjectivePoint<C>;
+    /// Returns `lhs + rhs`
+    fn add(lhs: &ProjectivePoint<C>, rhs: &ProjectivePoint<C>) -> ProjectivePoint<C>;
+    /// Returns `lhs + rhs`
+    fn add_mixed(lhs: &ProjectivePoint<C>, rhs: &AffinePoint<C>) -> ProjectivePoint<C>;
+}
```
tarcieri, line 35: I'm not sure I see the purpose of having two traits, one sealed (with a nondescript name) and one not, when `PointArithmetic` could just be sealed.

If it were sealed, it couldn't be used in bounds, but I'm not sure where that would be necessary.

survived: There's indeed no benefit in exposing `PointArithmetic` trait as `ProjectivePoint` is the only one who's supposed to use it. I'll hide entire trait then.

survived: Done!

tarcieri (approved):

PR #735, opened 2023-02-01
Update hash2curve implementations to new API

daxpedda:
Follow up to https://github.com/RustCrypto/traits/pull/1238.

Cc @tarcieri.


tarcieri:
Thanks!

PR #737, opened 2023-02-01
k256+p256+p384: add `PrimeField` constant tests

tarcieri:
Adds tests for the `TWO_INV`, `ROOT_OF_UNITY`, and `ROOT_OF_UNITY_INV` constants, based on the tests generated via `ff_derive`.

Some of them are currently failing and have been annotated with `#[ignore]` and a TODO comment.

### DIFF
--- a/k256/src/arithmetic/field.rs
+++ b/k256/src/arithmetic/field.rs
@@ -518,10 +518,9 @@ mod tests {
     }
 
     #[test]
-    #[ignore] // TODO(tarcieri): debug failure
     fn two_inv_constant() {
         assert_eq!(
-            FieldElement::from(2u64) * FieldElement::TWO_INV,
+            (FieldElement::from(2u64) * FieldElement::TWO_INV).normalize(),
             FieldElement::ONE
         );
     }
@@ -531,16 +530,17 @@ mod tests {
     fn root_of_unity_constant() {
         // ROOT_OF_UNITY^{2^s} mod m == 1
         assert_eq!(
-            FieldElement::ROOT_OF_UNITY.pow_vartime(&[1u64 << FieldElement::S, 0, 0, 0]),
+            FieldElement::ROOT_OF_UNITY
+                .pow_vartime(&[1u64 << FieldElement::S, 0, 0, 0])
+                .normalize(),
             FieldElement::ONE
         );
     }
 
     #[test]
-    #[ignore] // TODO(tarcieri): debug failure
     fn root_of_unity_inv_constant() {
         assert_eq!(
-            FieldElement::ROOT_OF_UNITY * FieldElement::ROOT_OF_UNITY_INV,
+            (FieldElement::ROOT_OF_UNITY * FieldElement::ROOT_OF_UNITY_INV).normalize(),
             FieldElement::ONE
         );
     }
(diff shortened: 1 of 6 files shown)


k256/src/arithmetic/field.rs
```diff
@@ -516,6 +517,34 @@ mod tests {
...
+    }
+
+    #[test]
+    #[ignore] // TODO(tarcieri): debug failure
+    fn root_of_unity_inv_constant() {
+        assert_eq!(
+            FieldElement::ROOT_OF_UNITY * FieldElement::ROOT_OF_UNITY_INV,
+            FieldElement::ONE
+        );
+    }
```
tarcieri, line 546: @fjarri guess you were right about a cross-check because all of these are failing.

I'll try recalculating them.

fjarri: Huh, I was pretty sure I did it correctly. That's weird.

tarcieri:
Note: they're copy-pasted for now, but once I get these all working (including `DELTA`) I can extract them into `elliptic-curve::dev` as a macro or something

k256/src/arithmetic/field.rs
fjarri: Are you sure there's no normalization missing? I looked in the code and it seems `FieldElement` does not automatically normalize.

fjarri:
I think there are a few tests missing here:
- `MULTIPLICATIVE_GENERATOR ^ ((modulus - 1) >> S) == ROOT_OF_UNITY`
- `MULTIPLICATIVE_GENERATOR ^ (2^S) == DELTA`

tarcieri:
Yeah as noted in the commit message the `DELTA` constants aren't defined yet

k256/src/arithmetic/field.rs
tarcieri: I tried adding normalization in [84adaa4](https://github.com/RustCrypto/elliptic-curves/pull/737/commits/84adaa4e2d7869a309a7860e239c8a4f65fd8605). It fixed the tests for the constant inversions, but the root of unity constant test is still failing (possibly a normalization-related issue related to `Field::pow_vartime`?)

tarcieri: Went through all of the constants and found the problem: `S` was `0` where it should be 1. `m - 1` is the following in binary:

```
1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111011111111111111111111110000101110
```

`S` is the number of trailing zeroes (which is somewhat confusingly described as "leading zero bits in the little-endian bit representation" in the `ff` docs /cc @str4d

So in this case, it's 1. With that changed, the `k256` tests pass.

tarcieri:
Whoa, just hit an ICE on stable Rust (looks like while compiling the dependencies):

https://github.com/RustCrypto/elliptic-curves/actions/runs/4078778175/jobs/7029397528

k256/src/arithmetic/field.rs
fjarri: It seems that all the values I calculated were correct then, that's a relief.

tarcieri:
Okay, after an ICE and failures possibly related to [GitHub rewriting OS package tarballs](https://github.com/orgs/community/discussions/45830) I'm going to land this while it's green.

Will backfill a few additional tests and try to consolidate the definitions of the test into a reusable macro.

PR #740, opened 2023-02-03
Bump `elliptic-curve` to v0.13.0-pre.4

daxpedda:
Builds on top of https://github.com/RustCrypto/signatures/pull/652.

Would it be possible to make a `p256`/`p384` v0.13.0-pre.0 release?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,7 +331,7 @@ dependencies = [
 [[package]]
 name = "ecdsa"
 version = "0.16.0-pre.0"
-source = "git+https://github.com/RustCrypto/signatures.git#4e2b0b2cfcfaef61a1e8851080cf640313a6e516"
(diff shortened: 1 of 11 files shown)


daxpedda:
Just a small note so it doesn't get overlooked:
I removed the `invert_vartime()` method on `p256::Scalar` because it's now part of the `Invert` trait, leaving it there would lead to conflicting method calls for downstream users.

tarcieri:
I'm going to need to cut another release of `elliptic-curve` for this: https://github.com/RustCrypto/traits/pull/1242

Also I intend to fill in that implementation of `invert_vartime`.

(and this is why I kick myself for releasing crates before I get all of the repos in sync)

daxpedda:
That's alright, thank you for your incredibly hard and timely work :+1:!

PR #741, opened 2023-02-03
k256+p256+p384: impl `Invert` trait for `Scalar` types

tarcieri:
The `Invert` trait now provides an `invert_vartime` method which can use a faster implementation than the constant time version.

Leveraging this requires an explicit impl of the trait however, whereas before it was provided by a blanket impl for types which impl `ff::Field`.


p256/src/arithmetic/scalar.rs
```diff
@@ -348,6 +342,72 @@ impl FromUintUnchecked for Scalar {
...
+    ///
+    /// <https://link.springer.com/article/10.1007/s13389-016-0135-4>
+    ///
+    /// ⚠️ WARNING!
+    ///
+    /// This method should not be used with (unblinded) secret scalars, as its
+    /// variable-time operation can potentially leak secrets through
+    /// sidechannels.
+    #[allow(non_snake_case)]
+    fn invert_vartime(&self) -> CtOption<Self> {
```
tarcieri, line 364: @fjarri FYI, there's now a dedicated method for variable-time inversions which is used by the generic ECDSA implementation.

This can probably be adapted to work with `k256`?  I had previously tried to make it fully generic, but due to the use of `Self(MODULUS)` below it's relying on support in the underlying field implementation.

I tried porting it to `k256` and it seemed to work, although ECDSA verification performance was only up by ~3%.

PR #743, opened 2023-02-04
k256: fast `invert_vartime` using Stein's algorithm

tarcieri:
Adapts the implementation originally contributed to the `p256` crate by @nickray to the `k256` crate.

Implementation is checked against the constant-time `Scalar::invert` using proptests (where `Scalar::invert` is in turn proptested against `num-bigint`).

It results in a ~9% ECDSA verification performance improvement according to our criterion benchmarks:

```
ecdsa/verify_prehashed  time:   [67.681 µs 67.734 µs 67.796 µs]
                        change: [-9.4013% -9.2055% -9.0011%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/k256/src/arithmetic/scalar.rs
+++ b/k256/src/arithmetic/scalar.rs
@@ -418,6 +418,64 @@ impl Invert for Scalar {
     fn invert(&self) -> CtOption<Self> {
         self.invert()
     }
+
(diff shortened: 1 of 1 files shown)


nickray:
Nice! LGTM

PR #748, opened 2023-02-06
K256 schnorr serde

### LINKED ISSUE #745: Serde support for k256::schnorr::VerifyingKey
Whilst there is serde support for the ecdsa verifying key, and for the underlying curve point & scalar - the schnorr structs do not implement it. They do implement a `to_bytes()` but I believe it is helpful to keep a unified interface.

ycscaly:
resolves #745

### DIFF
--- a/k256/src/schnorr/signing.rs
+++ b/k256/src/schnorr/signing.rs
@@ -18,6 +18,9 @@ use signature::{
     DigestSigner, Error, KeypairRef, RandomizedDigestSigner, RandomizedSigner, Result, Signer,
 };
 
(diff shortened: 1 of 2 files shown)


tarcieri:
@ycscaly needs rustfmt

ycscaly:
@tarcieri formatted, should be ready for PR unless you want me to implement Into<FieldBytes<C>> here as well (or should it be a separate PR)?

tarcieri:
This is fine for now.

You can open a separate PR for `FieldBytes` conversions if they're missing (although they should really be `From<&SigningKey> for FieldBytes`.

PR #772, opened 2023-03-02
fix compactabtility check

lthiery:
The function `to_compact_encoded_point` should return a CtOption where is_some set to 0 if the AffinePoint is non-compactable.

The current implementation checks if the EncodedPoint from bytes decodes properly to determine compactability. However, the EncodedPoint will always decode just fine, even when a non-compactable point has been compacted. 

Instead, what needs to be checked is if the compacted point and the original affine points Y coordinates match.

### DIFF
--- a/p256/tests/affine.rs
+++ b/p256/tests/affine.rs
@@ -125,7 +125,7 @@ fn noncompatible_is_none() {
         175, 232, 180, 255, 91, 106, 124, 191, 224, 31, 177, 208, 236, 127, 191, 169, 201, 217,
         75, 141, 184, 175, 120, 85, 171, 8, 54, 57, 33, 177, 83, 211,
     ];
-    let secret = p256::SecretKey::from_be_bytes(&noncompactable_secret).unwrap();
+    let secret = p256::SecretKey::from_bytes(&noncompactable_secret).unwrap();
     let is_compactable = secret.public_key().as_affine().to_compact_encoded_point().is_some().unwrap_u8();
     assert_eq!(is_compactable, 0);
 }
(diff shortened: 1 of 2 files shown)


p256/tests/affine.rs
```diff
@@ -118,3 +118,14 @@ fn identity_encoding() {
...
     ))
 }
+
+#[test]
+fn noncompatible_is_none() {
+    let noncompactable_secret = [
+        175, 232, 180, 255, 91, 106, 124, 191, 224, 31, 177, 208, 236, 127, 191, 169, 201, 217,
+        75, 141, 184, 175, 120, 85, 171, 8, 54, 57, 33, 177, 83, 211,
+    ];
+    let secret = p256::SecretKey::from_be_bytes(&noncompactable_secret).unwrap();
```
tarcieri, line 128: The latest release of `elliptic-curve` defines a specific endianness per-curve:

```suggestion
    let secret = p256::SecretKey::from_bytes(&noncompactable_secret).unwrap();
```

tarcieri:
Needs rustfmt, otherwise looks good

lthiery:
Thanks for the quick review. Let me know if it's good now.

lthiery:
Working on the test still.

lthiery:
Okay, I think it's ready unless you spot anything else.

Is it possible for a point-release with this? We have [a downstream crate](https://github.com/helium/helium-crypto-rs) that we'd like to publish with this dependency updated.

tarcieri:
You caught me just as I was about to cut a new set of breaking crate releases after merging #770.

Backports are a bit messy.

lthiery:
Perfect! Thanks for everything.

PR #794, opened 2023-03-18
Update field.rs

smoelius:
This clarifies test `unnormalized_is_odd` slightly, since the panic is in `is_odd`, not `unwrap_u8`.

Feel free to close this if you think this is unnecessary.


tarcieri:
Thanks!

PR #796, opened 2023-03-22
bign-curve256v1

makavity:
Hello!
Just implemented bign curve (STB 34.101.45-2013).
Need a review.
Thanks!


bign256/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## 0.1.0 (2023-03-22)
+- Initial release
```
tarcieri, line 8: I'd suggest removing this for now and it can be added at the time of release:

```suggestion
```

bign256/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+name = "bign256"
+version = "0.1.0"
```
tarcieri, line 3: This can be bumped at the time of release as well.

```suggestion
version = "0.0.0"
```

bign256/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+description = """
+Pure Rust implementation of the Bign P-256 (a.k.a. bign-curve256v1)
+elliptic curve as defined in STB 34.101.45-2013, with
+general purpose curve arithmetic
+"""
+authors = ["RustCrypto Developers"]
+license = "Apache-2.0 OR MIT"
+documentation = "https://docs.rs/bign-curve256v1"
+repository = "https://github.com/RustCrypto/elliptic-curves/tree/master/bign-curve256v1"
```
tarcieri, line 12:
```suggestion
repository = "https://github.com/RustCrypto/elliptic-curves/tree/master/bign256"
```

bign256/benches/field.rs
```diff
@@ -0,0 +1,55 @@
+//! secp256r1 field element benchmarks
```
tarcieri, line 1:
```suggestion
//! bign-curve256v1 field element benchmarks
```

bign256/benches/scalar.rs
```diff
@@ -0,0 +1,75 @@
+//! secp256r1 scalar arithmetic benchmarks
```
tarcieri, line 1:
```suggestion
//! bign-curve256v1 scalar arithmetic benchmarks
```

bign256/src/arithmetic/field.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// Returns self^(2^n) mod p.
+    const fn sqn(&self, n: usize) -> Self {
+        let mut x = *self;
+        let mut i = 0;
+        while i < n {
+            x = x.square();
+            i += 1;
+        }
+        x
+    }
```
tarcieri, line 116: If this is dead code you can just delete it

bign256/README.md
```diff
@@ -0,0 +1,88 @@
...
+operations are performed in constant time (using the `subtle` crate and
+constant-time formulas). However, it has not been thoroughly assessed to ensure
+that generated assembly is constant time on common CPU architectures.
+
+USE AT YOUR OWN RISK!
+
+## Supported Algorithms
+
+- [Elliptic Curve Digital Signature Algorithm (ECDSA)][ECDSA]: gated under the
+  `ecdsa` feature.
```
tarcieri, line 32: It doesn't look like you've wired up an ECDSA implementation yet (e.g. by adding an `ecdsa` module which defines curve-specific type aliases)

tarcieri (commented):
It'd be good to add some test vectors. Ones like these are ideal if you can find them:

https://github.com/RustCrypto/elliptic-curves/blob/master/p256/src/test_vectors/group.rs

Otherwise (or in addition to the above) ECDSA test vectors would also work.

tarcieri:
Another way to obtain test vectors is to generate them from a known working implementation

bign256/README.md
makavity: It is necessary to add ECDSA, or I can add it later?

makavity:
> It'd be good to add some test vectors. Ones like these are ideal if you can find them:
> 
> https://github.com/RustCrypto/elliptic-curves/blob/master/p256/src/test_vectors/group.rs
> 
> Otherwise (or in addition to the above) ECDSA test vectors would also work.

I will generate it by sagemath, I suppose.

bign256/README.md
tarcieri: You can add it later, but in the meantime it would probably be good to remove that from the documentation until it's implemented

makavity:
```python
def itoh(value):
    return '{:X}'.format(int(str(value)))

P = 2**256 - 189
F = GF(P)
A = 2**256 - 192
B = 0x77CE6C1515F3A8EDD2C13AABE4D8FBBE4CF55069978B9253B22E7D6BD69C03F1
EC = EllipticCurve(F, [A, B])

Gy = 0x6BF7FC3CFB16D69F5CE4C9A351D6835D78913966C408F6521E29CF1804516A93
Gx = 0

G = EC(Gx, Gy)

q = G

k = [
    112233445566778899,
112233445566778899112233445566778899,
29852220098221261079183923314599206100666902414330245206392788703677545185283,
57896042899961394862005778464643882389978449576758748073725983489954366354431,
1766845392945710151501889105729049882997660004824848915955419660366636031,
28948025760307534517734791687894775804466072615242963443097661355606862201087,
113078210460870548944811695960290644973229224625838436424477095834645696384,
12078056106883488161242983286051341125085761470677906721917479268909056,
57782969857385448082319957860328652998540760998293976083718804450708503920639,
57896017119460046759583662757090100341435943767777707906455551163257755533312,
452312848374287284681282171017647412726433684238464212999305864837160993279,
904571339174065134293634407946054000774746055866917729876676367558469746684,
115792089210356248762697446949407573529996955224135760342422259061068512044349,
115792089210356248762697446949407573529996955224135760342422259061068512044350,
115792089210356248762697446949407573529996955224135760342422259061068512044351,
115792089210356248762697446949407573529996955224135760342422259061068512044352,
115792089210356248762697446949407573529996955224135760342422259061068512044353,
115792089210356248762697446949407573529996955224135760342422259061068512044354,
115792089210356248762697446949407573529996955224135760342422259061068512044355,
115792089210356248762697446949407573529996955224135760342422259061068512044356,
115792089210356248762697446949407573529996955224135760342422259061068512044357,
115792089210356248762697446949407573529996955224135760342422259061068512044358,
115792089210356248762697446949407573529996955224135760342422259061068512044359,
115792089210356248762697446949407573529996955224135760342422259061068512044360,
115792089210356248762697446949407573529996955224135760342422259061068512044361,
115792089210356248762697446949407573529996955224135760342422259061068512044362,
115792089210356248762697446949407573529996955224135760342422259061068512044363,
115792089210356248762697446949407573529996955224135760342422259061068512044364,
115792089210356248762697446949407573529996955224135760342422259061068512044365,
115792089210356248762697446949407573529996955224135760342422259061068512044366,
115792089210356248762697446949407573529996955224135760342422259061068512044367,
115792089210356248762697446949407573529996955224135760342422259061068512044368,
]

for i in k:
    q = i * G
    x, y, z = itoh(q[0]), itoh(q[1]), itoh(q[2])
    print(f"(\n\thex!(\"{x}\"),\n\thex!(\"{y}\"),\n\thex!(\"{z}\"),\n),")
    
q = G
for i in range(1, 21):
    q = q + G
    x, y, z = itoh(q[0]), itoh(q[1]), itoh(q[2])
    print(f"(\n\thex!(\"{x}\"),\n\thex!(\"{y}\"),\n),")
    
```

tarcieri (approved):
Looks good, thank you!

makavity:
@tarcieri looks like it is merged with version 0.0.0, that’s fine? :)

tarcieri:
@makavity we do separate release PRs that include a version bump and CHANGELOG.md entry. I can do one if you'd like.

makavity:
I will make some changes in Monday and will commit it. After, this would be nice. Thank you.

newpavlov:
@makavity 
BTW why have you decided to name the crate as `bign256` instead of the full name `bign-curve256v1`? I don't think the curve is addressed like this anywhere else (same for "BIGN P-256"), at the very least, after the cursory search all results point to us.

makavity:
> @makavity
> BTW why have you decided to name the crate as `bign256` instead of the full name `bign-curve256v1`? I don't think the curve is addressed like this anywhere else (same for "BIGN P-256"), at the very least, after the cursory search all results point to us.

Because `bign-curve256v1` - is too long.
But, I suppose, we can rename it. There is no problems.

newpavlov:
@tarcieri 
WDYT? Should we keep using the current name? I guess we have a similar precedent with the Brainpool crates.

tarcieri:
I guess I will say I was a little surprised by the crate name when I read the paper recently.

That said, all of the crates use abbreviated names, so it's at least not inconsistent, but whether or not the abbreviated names are a good thing is debatable.

newpavlov:
Yeah, discoverability is my main concern. I even thought that we don't have an implementation of the STB curves after a cursory look at the repository.

For now I registered `bign-curve256v1` and referenced `bign256` in its description.

tarcieri:
@makavity how about `bignp256`?

makavity:
@tarcieri sounds fine and does not contradict the standard

PR #804, opened 2023-04-02
feat: implement AssociatedAlgorithmIdentifier support

lumag:


p256/src/lib.rs
```diff
@@ -130,6 +133,14 @@ impl pkcs8::AssociatedOid for NistP256 {
     const OID: pkcs8::ObjectIdentifier = pkcs8::ObjectIdentifier::new_unwrap("1.2.840.10045.3.1.7");
 }
 
+#[cfg(feature = "algorithm-identifier")]
+impl AssociatedAlgorithmIdentifier for NistP256 {
+    const ALGORITHM_IDENTIFIER: AlgorithmIdentifierRef<'static> = AlgorithmIdentifierRef {
+        oid: const_oid::ObjectIdentifier::new_unwrap("1.2.840.10045.4.3.2"),
```
tarcieri, line 139: This is weird to me as these identify specific flavors of ECDSA, which is a concern quite beyond a specific elliptic curve.

There is an `AlgorithmIdentifier` that should be associated with each curve, but it's the one consisting of two OIDs: `id-ecPublicKey` with params of `C::OID` ("1.2.840.10045.3.1.7" or prime256v1 in this case).

tarcieri: I was trying to handle it within the `ecdsa` crate in this PR: https://github.com/RustCrypto/signatures/pull/686/files#diff-e83fd22999ab8d7360e0664cabe31e0ed25aca5a41ce1c40be42c5ca3a941cf3R518-R524

Though seems I need to pull that out into a `const fn` so it can be used to define constants.

lumag: Hmm, true. It seems I got that mixed up at some point.

lumag: ... or not. I remember why I did it that way. Adding it to the `SigningKey` directly is impossible, so I followed what was implemeted for `AssociatedOid`. Thus I ended with this kind of a hack.

tarcieri: `SigningKey` and `VerifyingKey` should probably both return the `id-ecPublicKey` AlgorithmIdentifier, since it identifies the elliptic curve used.

I changed https://github.com/RustCrypto/signatures/pull/686 to put the signature-specific one on the `Signature<C>` type.

lumag: @tarcieri thank you, the merged implementation is suitable for my needs.

PR #824, opened 2023-04-10
sm2: initial crate

tarcieri:
Initial crate boilerplate with a `Sm2` type, `Curve` impl, and type aliases for the generic types which can work with that.


uudiin:
This is awesome, thanks for your contribution.

PR #839, opened 2023-04-13
Add feature "serde" to primeorder to filter down from curves (P256, at least)

GTC6244:
### WHAT

Per the documentation here: https://docs.rs/crate/p256/latest/features , it would appear that `serde` should be a feature flag within `P256` on the `primeorder` crate, in order properly de/ser various primitives.


tarcieri:
Yes, something like this is needed, but this makes the `serde` feature of `primeorder` active even if `p256/serde` is not.

Instead, you can change the `serde` feature in `p256` to activate `primeorder/serde`.

GTC6244:
Hey @tarcieri  - sorry, late night coding patch for a different project.  Will tidy it up!

PR #840, opened 2023-04-13
Add feature "serde" to primeorder when flagged in P256

GTC6244:


p256/Cargo.toml
```diff
@@ -50,7 +50,7 @@ hash2curve = ["arithmetic", "elliptic-curve/hash2curve"]
 jwk = ["elliptic-curve/jwk"]
 pem = ["elliptic-curve/pem", "ecdsa-core/pem", "pkcs8"]
 pkcs8 = ["ecdsa-core/pkcs8", "elliptic-curve/pkcs8"]
-serde = ["ecdsa-core/serde", "elliptic-curve/serde", "serdect"]
+serde = ["ecdsa-core/serde", "elliptic-curve/serde", "dep:primeorder", "serdect"]
```
tarcieri, line 53:
```suggestion
serde = ["ecdsa-core/serde", "elliptic-curve/serde", "primeorder/serde", "serdect"]
```

GTC6244: @tarcieri  - I'm not sure that works; primeorder is optional itself (I tried your solution first, prior to the dep: change).

GTC6244: ie, to be clear - your change won't compile for me ;)

tarcieri:
I opened #851 to fix this up across all of the crates. Thanks for getting the ball rolling, though!

PR #850, opened 2023-04-14
p192: add `ecdsa` feature

halvors:
Add ECDSA support for P-192, tests not implemented yet due to lacking signer support in P-192 and also requirement to use `PrehashSigner::sign_prehash()` since `ecdsa` crate currently only supports sha2 hashing algorithms for signing

### DIFF
--- a/p192/src/ecdsa.rs
+++ b/p192/src/ecdsa.rs
@@ -39,10 +39,7 @@
 
 pub use ecdsa_core::signature::{self, Error};
 #[cfg(feature = "ecdsa")]
-use {
-    crate::{AffinePoint, Scalar},
-    ecdsa_core::hazmat::{SignPrimitive, VerifyPrimitive},
-};
+use {crate::AffinePoint, ecdsa_core::hazmat::VerifyPrimitive};
 
 use super::NistP192;
 
@@ -59,9 +56,6 @@ pub type SigningKey = ecdsa_core::SigningKey<NistP192>;
 /// ECDSA/P-192 verification key (i.e. public key)
 #[cfg(feature = "ecdsa")]
 pub type VerifyingKey = ecdsa_core::VerifyingKey<NistP192>;
-
-#[cfg(feature = "ecdsa")]
-impl SignPrimitive<NistP192> for Scalar {}
 
 #[cfg(feature = "ecdsa")]
 impl VerifyPrimitive<NistP192> for AffinePoint {}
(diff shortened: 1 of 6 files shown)


p192/src/ecdsa.rs
```diff
@@ -0,0 +1,116 @@
...
+/// ECDSA/P-192 signing key
+#[cfg(feature = "ecdsa")]
+pub type SigningKey = ecdsa_core::SigningKey<NistP192>;
+
+/// ECDSA/P-192 verification key (i.e. public key)
+#[cfg(feature = "ecdsa")]
+pub type VerifyingKey = ecdsa_core::VerifyingKey<NistP192>;
+
+#[cfg(feature = "ecdsa")]
+impl SignPrimitive<NistP192> for Scalar {}
```
tarcieri, line 64: I'd prefer to leave the signing components out for now, per NIST's guidance

halvors: Done

halvors:
@tarcieri I have added test vectors as well test pass (only verifying), i am a little bit unsure about where to obtain `wycheproof` but it seem to require signing, is only verifying test with the test vectors sufficient for P-192?

p192/src/ecdsa.rs
```diff
@@ -13,29 +13,6 @@
 //! - `ecdsa`: provides `ecdsa-core` features plus the [`SigningKey`] and
 //!   [`VerifyingKey`] types which natively implement ECDSA/P-192 signing and
 //!   verification.
-//!
```
halvors: What to do about this, impossible to obtain the signing key without signing support, should we remove this or use a static set example key?

p192/src/ecdsa.rs
```diff
@@ -0,0 +1,87 @@
...
+
+/// ECDSA/P-192 signature (fixed-size)
+pub type Signature = ecdsa_core::Signature<NistP192>;
+
+/// ECDSA/P-192 signature (ASN.1 DER encoded)
+pub type DerSignature = ecdsa_core::der::Signature<NistP192>;
+
+/// ECDSA/P-192 signing key
+#[cfg(feature = "ecdsa")]
+pub type SigningKey = ecdsa_core::SigningKey<NistP192>;
```
tarcieri, line 31: Can you remove this (and other mentions of `SigningKey`) as well?

It won't work without `SignPrimitive`.

p192/src/ecdsa.rs
tarcieri: Remove it from the docs

tarcieri:
Correct you can't run the Wycheproof tests without a `SigningKey`, so just include the NIST CAVS vectors

p192/src/ecdsa.rs
halvors: Sure, done

halvors:
> Needs rustfmt but otherwise looks good.
> 
> Edit: I'll take care of rustfmt.

Thanks, I am sorry that I forgot

PR #860, opened 2023-04-15
Update docs for fiat-crypto generated code

tarcieri:
Some of the field implementations came from the git repo, others were synthesized by running fiat-crypto locally.

This changes the boilerplate to be vague enough to cover both cases, as opposed to what it was before where it claimed all of them were copied from the git repo.


tarcieri:
Separately, I'm not sure fiat-crypto output is actually under copyright, since it's more like compiler output (which isn't necessarily copyright by the compiler author).

But, probably safer to treat it as if it is.

Issue #882, opened 2023-04-17
p256 point multiplication got ~25% slower in 0.12


randombit:
We just noticed in some internal benchmarks that `p256` performance for point multiplication and point addition both became notably slower in the 0.12 release. This is replicated by the benchmarks in `p256` as well, at least for point multiplication (seems point addition is not benchmarked):

0.11.1
```
point operations/point-scalar mul
                        time:   [104.03 us 104.21 us 104.42 us]
```

Current `master`
```
point operations/point-scalar mul
                        time:   [134.47 µs 134.63 µs 134.79 µs]
```

Both run on the same x86-64 Linux laptop, otherwise idle, Rust 1.65.

We looked at the release notes but don't see any reference to a change that suggests a performance slowdown, I also reviewed the diff and don't see any obvious culprits. Wanted to open an issue to check if this performance change was expected, and if so if there is any way to improve the situation to reach parity with 0.11 performance.

tarcieri:
The biggest change was probably switching to the generic curve arithmetic implementation in the `primeorder` crate. Perhaps that impacted inlining?

randombit:
I didn't have much luck bisecting so I wrote a script that just benchmarks every commit

```
#!/usr/bin/python

import subprocess
import re
import sys

def run_command(cmdline, cwd=None):
    proc = subprocess.run(cmdline, cwd=cwd, capture_output=True)
    return proc.stdout.decode('utf8')

time = re.compile(' +time: +\[(.*)\]')

for commit in open('p256_commits'):
    commit = commit.strip()
    run_command(['git', 'checkout', commit])
    at_commit = run_command(['git', 'rev-parse', 'HEAD']).strip()

    assert(commit == at_commit)

    output = run_command(['cargo', 'bench', 'point-'], cwd='p256')

    for line in output.split('\n'):
        m = time.match(line)
        if m:
            print(commit, m.group(1))
            break

    sys.stdout.flush()
```

where `p256_commits` contains the ID for every commit between 0.11.1 and `HEAD` that touched `p256`:

`$ git log  p256/v0.11.1.. p256 | grep ^commit | cut -c 8- > p256_commits`

My results from this seem to implicate especially cea8f60ff76ba4f447a5169b9fb44f788b1217c4, across two runs there is clear jump there

```
cea8f60ff76ba4f447a5169b9fb44f788b1217c4 149.26 µs 149.72 µs 150.16 µs
[...]

PR #885, opened 2023-04-24
p256: revert `primeorder` field impls

tarcieri:
Partially reverts cea8f60 which caused a performance degradation.

This aims to address #882.


tarcieri:
@randombit can you benchmark this change and see if it improves performance the way you're expecting?

I get ~6% improvement testing on a M2 Max versus `master`, and the scalar mult performance looks the same as it was in f1878f9, i.e. the parent of cea8f60 (while performance on the other benchmarks improved across the board)

randombit:
@tarcieri Thanks for looking into this. I'm away from my machines right now but will evaluate this against our internal benchmarks tomorrow.

tarcieri:
@randombit ping!

altkdf:
I get the following run times on my mac (Apple M1):
```
v0.11.1
Benchmarking point operations/point-scalar mul:
                        time:  [97.383 us 97.537 us 97.784 us]
```
```
v0.13.2
Benchmarking point operations/point-scalar mul:
                        time:   [120.76 µs 120.86 µs 120.97 µs]
```
```
p256/revert-primeorder-field-impls
Benchmarking point operations/point-scalar mul:
                        time:   [111.75 µs 111.89 µs 112.11 µs]
```

tarcieri:
I'll go ahead and merge this as it seems to improve performance, but I don't think it's the complete story quite yet.

randombit:
Thanks for following up on this @altkdf

PR #891, opened 2023-05-17
primeorder: improve initialization from uncompressed point

wangwillian0:
If you already have the uncompressed point, it's faster to just check the x and y against the curve's equation than solving the equation and calculating the modular square root of y².

In my very rough benchmarks, doing 200k ECDH operations with the same P224 uncompressed key, the time went from ~76s to ~38s, a 2x speedup!


tarcieri:
Looks good, thanks!

PR #892, opened 2023-05-22
bign256: Cleanup

makavity:
* Gated features
* Removed UInt alias for U256
* Doc link fixed in Cargo.toml

Also, @tarcieri I got question about dsa impl and reduce.
1. Should I implement DSA like sm2 crate? If yes, what structure for signature should be? s: NonZeroScalar?
2. For PAKe algorithm (BPACE STB 34.101.69), I need to reduce 64 bytes to FieldElement size, by modulus. Which is most correct way to do it? Should I implement some additional algorithm right in bign256 crate?
Thank you!


tarcieri:
> Should I implement DSA like sm2 crate? If yes, what structure for signature should be? s: NonZeroScalar?

I'm not really familiar with this curve or the DSA algorithm it supports. Do you have a link to a spec I can look at?

If it's some bespoke ECDSA-alike similar to SM2DSA then yes you'll need to implement it yourself in this crate.

>  I need to reduce 64 bytes to FieldElement size, by modulus. Which is most correct way to do it? Should I implement some additional algorithm right in bign256 crate?

[`elliptic_curve::ops::Reduce`](https://docs.rs/elliptic-curve/latest/elliptic_curve/ops/trait.Reduce.html) is our modular reduction trait. It's generic around an integer representing the input, so you can define it for `U512`.

A wide reduction like that won't be provided by fiat-crypto, so yes you'll need to implement it yourself. You can take a look here (and in `wide32.rs`/`wide64.rs`) for an example: https://github.com/RustCrypto/elliptic-curves/blob/b2e281855cb40f5c9b0903f2d04e73a899f4044e/k256/src/arithmetic/scalar.rs#L676-L686

makavity:
> I'm not really familiar with this curve or the DSA algorithm it supports. Do you have a link to a spec I can look at?
> 
> If it's some bespoke ECDSA-alike similar to SM2DSA then yes you'll need to implement it yourself in this crate.

image
https://apmi.bsu.by/assets/files/std/bign-spec294.pdf

tarcieri:
That looks fairly close to a Schnorr signature, though I'm not clear what H̅ is in step 5.

makavity:
> though I'm not clear what H̅ is in step 5.

It is word from H. ( an and b points on screenshot. 
IMG_3236

tarcieri (approved):
This otherwise looks good, yes.

Issue #847, opened 2023-04-13
bp384: `Scalar::sqrt` is unimplemented


tarcieri:
The following crates lack a `Scalar::sqrt` implementation and panic instead:

- [x] `p224`
- [ ] `bp384`

Unfortunately we can't simply remove the methods because they're mandatory to implement for important traits like [`ff::Field`](https://docs.rs/ff/latest/ff/trait.Field.html#method.sqrt)

PR #904, opened 2023-07-04
p224 Scalar::sqrt

ashWhiteHat:
Part of https://github.com/RustCrypto/elliptic-curves/issues/847

I implemented p224 sqrt.
I would appreciate it if you could confirm.
Thank you!


tarcieri:
Thank you!

Issue #908, opened 2023-07-10
As of 0.12 k256 no longer requires ECDSA signatures have minimum s


randombit:
In 0.11 and earlier versions, `k256` required that ECDSA signatures have minimal `s`, but in 0.12 it appears verification will accept either `s`. I don't see anything mentioned about this in the `k256` changelog; was this change intentional?

PR #911, opened 2023-07-21
k256: use generic signing implementation from `ecdsa`

tarcieri:
Uses `ecdsa::hazmat::sign_prehashed` which provides a generic implementation of ECDSA signing, which was introduced in https://github.com/RustCrypto/signatures/pull/731 and released in `ecdsa` v0.16.8.

The code is largely identical to what was previously used, that code having originally been copied from `p256` and updated iteratively to match the generic implementation in the `ecdsa` crate as it evolved.

The only reason why we can't use the default implementation of `SignPrimitive::try_sign_prehashed` is to handle low-S normalization, which is unique to the secp256k1 ecosystem but nearly ubiquitous and very much expected of ECDSA/secp256k1 implementations.

In the next release this can hopefully be upstreamed to the `ecdsa` crate as well, which would eliminate the need for the `SignPrimitive` and `VerifyPrimitive` traits (which pretty much exist exclusively so `k256` can have custom logic to handle low-S normalization).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -350,9 +350,9 @@ dependencies = [
 
 [[package]]
 name = "ecdsa"
-version = "0.16.7"
+version = "0.16.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0997c976637b606099b9985693efa3581e84e41f5c11ba5255f88711058ad428"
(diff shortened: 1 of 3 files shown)


tarcieri:
Note: #908 still needs to be addressed, but is a similar change

PR #917, opened 2023-08-02
k256: simplify internal helper functions in the scalar arithmetic

fjarri:
- use `overflowing_add()` instead of the explicit `ct_less()`. Seems to compile to the same assembly code. 
- use checked addition (the default for arithmetic operations) instead of an explicit `debug_assert()`

I am 99% sure that `overflowing_add()` will compile to the same (or safer) assembly... On the weird architectures where there's no `adc` or an equivalent (are there any even?), `<` may very easily not be constant-time too.

### DIFF
--- a/k256/src/arithmetic/scalar/wide32.rs
+++ b/k256/src/arithmetic/scalar/wide32.rs
@@ -417,28 +417,18 @@ impl WideScalar {
     }
 }
 
-/// Constant-time comparison.
-#[inline(always)]
-fn ct_less(a: u32, b: u32) -> u32 {
-    // Do not convert to Choice since it is only used internally,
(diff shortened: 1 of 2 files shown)


tarcieri (approved):

tarcieri:
> On the weird architectures where there's no adc or an equivalent (are there any even?)

RISC-V is one of the strangest here, I believe due to a requirement that instructions not output to multiple registers. Instead it uses [`add` followed by `sltu` to obtain the carryout](https://gmplib.org/list-archives/gmp-devel/2021-September/006013.html). Somewhat annoyingly these have data dependencies on each other and can't be executed in parallel.

PR #918, opened 2023-08-02
bp384: impl `Scalar::sqrt`

ashWhiteHat:
Close https://github.com/RustCrypto/elliptic-curves/issues/847

I implemented bp384 sqrt.
I would appreciate it if you could confirm.
Thank you!

### DIFF
--- a/bp384/src/arithmetic/scalar.rs
+++ b/bp384/src/arithmetic/scalar.rs
@@ -212,19 +212,26 @@ impl Scalar {
         res
     }
 
-    /// Returns the square root of self mod n, or `None` if no square root
-    /// exists.
+    /// Atkin algorithm for q mod 8 = 5
+    /// <https://eips.ethereum.org/assets/eip-3068/2012-685_Square_Root_Even_Ext.pdf>
+    /// (page 10, algorithm 3)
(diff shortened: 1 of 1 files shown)


bp384/src/arithmetic/scalar.rs
```diff
@@ -212,19 +212,26 @@ impl Scalar {
...
 
-    /// Returns the square root of self mod n, or `None` if no square root
-    /// exists.
+    /// Atkin algorithm for q mod 8 = 5
+    /// <https://eips.ethereum.org/assets/eip-3068/2012-685_Square_Root_Even_Ext.pdf>
+    /// (page 10, algorithm 3)
     pub fn sqrt(&self) -> CtOption<Self> {
-        // Because n ≡ 3 mod 4 for brainpoolP384's scalar field modulus, sqrt
-        // can be implemented with only one exponentiation via the computation
-        // of self^((n + 1) // 4) (mod n).
```
tarcieri, line 220: Oops, this should've been `todo!()`. Anyway, thanks for fixing it!

ashWhiteHat: You're welcome 👍

PR #935, opened 2023-10-03
[bign256] Digital signature

makavity:
* Implement tests
* Implement signing
* Implement verifying

Need a review, thanks :)


tarcieri (approved):
FYI, I know little of this algorithm and can't really speak to the correctness of the implementation, but this seems reasonable enough

PR #945, opened 2023-11-01
p521: Use unsaturated limbs

MasterAwesome:
Introduces a function to converts the `U576` into a little endian byte array in constant time and uses it for `from_uint_unchecked`.

Tests:
- Verified that math on generator such as doubling the point etc returns the expected outputs.

### DIFF
--- a/p521/src/arithmetic/field.rs
+++ b/p521/src/arithmetic/field.rs
@@ -42,8 +42,7 @@ use elliptic_curve::{
     Error, FieldBytesEncoding,
 };
 
-#[cfg(target_pointer_width = "32")]
-use super::util::u32x18_to_u64x9;
+use super::util::uint_to_le_bytes_unchecked;
 
 /// Constant representing the modulus serialized as hex.
 /// p = 2^{521} − 1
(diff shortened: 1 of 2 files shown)


p521/src/arithmetic/field.rs
```diff
@@ -101,14 +101,54 @@ impl FieldElement {
         Self::from_uint_unchecked(U576::from_u64(w))
     }
 
+    /// Converts the words into a byte array to be consumed by [`fiat_p521_from_bytes`]
+    const fn words_to_le_bytes_unchecked(words: [u64; 9]) -> [u8; 66] {
```
tarcieri, line 105: This can go in the [`util` module](https://github.com/RustCrypto/elliptic-curves/blob/master/p521/src/arithmetic/util.rs) as a free/static function.

Also, since it's designed to operate on the saturated representation used by `crypto-bigint`, it would probably make sense for it to accept a `&U576` as an argument to avoid confusion with `fiat_p521_tight_field_element`.

MasterAwesome: Done in https://github.com/RustCrypto/elliptic-curves/pull/945/commits/f0299fd0395c1997470d39b8ad1f2eb0c6fecf84

p521/src/arithmetic/util.rs
```diff
@@ -30,3 +32,47 @@ pub(crate) const fn u64x9_to_u32x18(w: &[u64; 9]) -> [u32; 18] {
 
     ret
 }
+
+/// Converts the saturated representation [`U576`] into a 528bit array. Each
+/// word is copied in little-endian.
+pub const fn uint_to_le_bytes_unchecked(w: U576) -> [u8; 66] {
```
tarcieri, line 38: We should probably add some `const fn` serializers upstream in `crypto-bigint` that could handle this sort of thing, but this is fine for now.

tarcieri (approved):

tarcieri:
Thanks! Curious if this might make some of the field arithmetic tests that were failing work.

tarcieri:
Something else still seems to be amiss: https://github.com/RustCrypto/elliptic-curves/pull/946

PR #946, opened 2023-11-01
p521: impl `FieldElement::{sqrt, invert}` and add basic field tests

tarcieri:
See also: #947

Also relevant (sage calculations):

- PR for base field constants: #758
- PR for scalar field constants: #760

### DIFF
--- a/p521/src/arithmetic/field.rs
+++ b/p521/src/arithmetic/field.rs
@@ -199,6 +199,17 @@ impl FieldElement {
         self.relax().square()
     }
 
+    /// Returns self^(2^n) mod p
+    const fn sqn(&self, n: usize) -> Self {
+        let mut x = *self;
+        let mut i = 0;
+        while i < n {
+            x = x.square();
+            i += 1;
+        }
+        x
+    }
+
     /// Returns `self^exp`, where `exp` is a little-endian integer exponent.
     ///
     /// **This operation is variable time with respect to the exponent.**
@@ -227,7 +238,40 @@ impl FieldElement {
 
     /// Compute [`FieldElement`] inversion: `1 / self`.
     pub fn invert(&self) -> CtOption<Self> {
-        todo!("`invert` not yet implemented")
+        CtOption::new(self.invert_unchecked(), !self.is_zero())
+    }
+
+    /// Returns the multiplicative inverse of self.
+    ///
+    /// Does not check that self is non-zero.
+    const fn invert_unchecked(&self) -> Self {
+        // Adapted from addchain: github.com/mmcloughlin/addchain
+        let z = self.square();
+        let z = self.multiply(&z);
+        let t0 = z.sqn(2);
+        let z = z.multiply(&t0);
+        let t0 = z.sqn(4);
+        let z = z.multiply(&t0);
+        let t0 = z.sqn(8);
+        let z = z.multiply(&t0);
+        let t0 = z.sqn(16);
+        let z = z.multiply(&t0);
+        let t0 = z.sqn(32);
+        let z = z.multiply(&t0);
+        let t0 = z.square();
+        let t0 = self.multiply(&t0);
+        let t0 = t0.sqn(64);
+        let z = z.multiply(&t0);
+        let t0 = z.square();
+        let t0 = self.multiply(&t0);
+        let t0 = t0.sqn(129);
+        let z = z.multiply(&t0);
+        let t0 = z.square();
+        let t0 = self.multiply(&t0);
+        let t0 = t0.sqn(259);
+        let z = z.multiply(&t0);
+        let z = z.sqn(2);
+        self.multiply(&z)
     }
 
     /// Returns the square root of self mod p, or `None` if no square root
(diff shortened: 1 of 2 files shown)


MasterAwesome:
The arithmetic itself seems to work, like `mul`, it needs the invert and square function implementation, but for the `two_inverse_const`, I can pass it if I input the correct FieldElement limbs:

```
two = FieldElement([2, 0, 0, 0, 0, 0, 0, 0, 0])
invert = FieldElement([0, 0, 0, 0, 0, 0, 0, 0, 72057594037927936])
mul = FieldElement([1, 0, 0, 0, 0, 0, 0, 0, 0])
```

tarcieri:
Aah yes, `PrimeField::{TWO_INV, ROOT_OF_UNITY_ENV}` are still stubbed out due to a missing inversion implementation.

That doesn't explain why the `root_of_unity_constant` and `delta_constant` tests are failing, though. That would speak to an issue with the constant definitions, `pow_vartime`, or `mul`

tarcieri:
Regarding square root, it should be possible to use the Tonelli–Shanks algorithm (https://eprint.iacr.org/2012/685.pdf) for the case where q ≡ 3 (mod 4).

This is a special case that can be solved using a simpler variant which is just the Shanks algorithm, computed as a single exponentiation, raising the input to (q + 1) / 4 (see p.2).

I'm computing `q` as:

```
0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```

...and `(q + 1) / 4` as:

```
0x4000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

...although if I attempt to implement square root that way it's failing the tests.

It computes `sqrt(4 mod p)` as:

```
FieldElement([0, 0, 0, 0, 536870912, 0, 0, 0, 0])
```

...which, when squared, yields:

```
FieldElement([2, 0, 0, 0, 0, 0, 0, 0, 0])
```

...instead of the expected `4`.

MasterAwesome:
> Regarding square root, it should be possible to use the Tonelli–Shanks algorithm (https://eprint.iacr.org/2012/685.pdf) for the case where `q ≡ 3 (mod 4)`.
> 
> This is a special case that can be solved using a simpler variant which is just the Shanks algorithm, computed as a single exponentiation, raising the input to `(q + 1) / 4` (see p.2).
> 
> I'm computing `q` as:
> 
> ```
> 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
> ```
> 
> ...and `(q + 1) / 4` as:
> 
> ```
> 0x4000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
> ```
> 
> ...although if I attempt to implement square root that way it's failing the tests.
> 
> It computes `sqrt(4 mod p)` as:
> 
> ```
> FieldElement([0, 0, 0, 0, 536870912, 0, 0, 0, 0])
> ```
> 
> ...which, when squared, yields:
> 
> ```
> FieldElement([2, 0, 0, 0, 0, 0, 0, 0, 0])
> ```
> 
> ...instead of the expected `4`.
> 
> Edit: pushed up sqrt experiments as [142d5f5](https://github.com/RustCrypto/elliptic-curves/pull/946/commits/142d5f5314f0f401897cabe25c2dd4482544cb0a)

Interesting, I implemented sqrt by exponentiating self 519 times but realized it only fails for 9,25,36,49. The top word is always input-1. So for 9 the tight_field[0] = 8. Same for 25,36,49. I'm unsure why the other words are incorrect. I'll take a look into it soon.

MasterAwesome:
Ok found something it seems to be that the square implementation is incorrect. When you do a self.relax().square() it produces different results compared to `Self(fiat_p521_carry_square(&self.0))` the latter passes my sqrt implementation

MasterAwesome:
Maybe the strategy is to reimplement PartialEq to compare as bytes instead of words, checking if this works now

p521/src/arithmetic/field.rs
```diff
@@ -233,7 +233,22 @@ impl FieldElement {
...
+        let w = self.pow_vartime(&[
+            0x00000000_00000000,
+            0x00000000_00000000,
+            0x00000000_00000000,
+            0x00000000_00000000,
+            0x00000000_00000000,
+            0x00000000_00000000,
+            0x00000000_00000000,
+            0x00000000_00000000,
+            0x00000000_00000040,
```
MasterAwesome, line 247: Needs to end with `80` because it's 2^519.

tarcieri: Hmm, interesting, substituting:

```suggestion
            0x00000000_00000080,
```

(not sure I follow why, but)

...passes sqrt for 1 and 4 but fails on 9:

```
---- arithmetic::field::tests::sqrt stdout ----
thread 'arithmetic::field::tests::sqrt' panicked at 'assertion failed: `(left == right)`
  left: `FieldElement([8, 0, 288230376151711744, 288230376151711743, 288230376151711743, 288230376151711743, 288230376151711743, 288230376151711743, 144115188075855871])`,
 right: `FieldElement([9, 0, 0, 0, 0, 0, 0, 0, 0])`', p521/src/arithmetic/field.rs:574:5
```

MasterAwesome: 9 fails due to https://github.com/RustCrypto/elliptic-curves/pull/946#issuecomment-1789836157

This will fix it

tarcieri: Well, indeed it all works, the sqrt tests are passing, although I'm still not sure why it's (p+1)/4 instead of (q+1)/4

tarcieri:
Looks like we just need inversions now. I found this paper which specifies an addition chain for P-521 inversions, although I'm not sure it's the most efficient option: https://eprint.iacr.org/2014/852.pdf

tarcieri:
With [82d6608](https://github.com/RustCrypto/elliptic-curves/pull/946/commits/82d6608159b2abbef1c8ea1339e75455fbb46085) added all of the basic field tests are now passing! 🎉

p521/src/arithmetic/field.rs
MasterAwesome: Assuming our number is a quadratic residue => `x^((p-1)/2)) = 1` due to Euler's criterion. Therefore `x^((p+1)/2)) = x`.

There's a simplification that can be made when `p=3mod4` in which case p can be expressed as `4k +3` which guarantees  `(p+1)/4` to be an integer in which case `(x^((p+1)/4))^2 = x`. This means `x^((p+1)/4)` is the square root.

MasterAwesome:
> With [82d6608](https://github.com/RustCrypto/elliptic-curves/pull/946/commits/82d6608159b2abbef1c8ea1339e75455fbb46085) added all of the basic field tests are now passing! 🎉 

You should be able to simplify the math that's happening in `pow` by just squaring 519 times. Also thanks for working with me on this, I appreciate the quick replies and everything else!

tarcieri:
Thanks for helping to get it working!

PR #949, opened 2023-11-02
p521: Improve `Debug` for unsaturated math

MasterAwesome:
When using p521, the limbs can be populated differently which when deriving [`Debug`] shows up as different values for equal FieldElements, this can lead to confusion.

See:
- https://github.com/RustCrypto/elliptic-curves/pull/946#issuecomment-1789749836, equal field elements were considered as unequal.
- https://github.com/RustCrypto/elliptic-curves/pull/946#discussion_r1380388316, assert_eq! wasn't meaningful.

### DIFF
--- a/p521/src/arithmetic/field.rs
+++ b/p521/src/arithmetic/field.rs
@@ -51,8 +51,44 @@ const MODULUS_HEX: &str = "00000000000001fffffffffffffffffffffffffffffffffffffff
 const MODULUS: U576 = U576::from_be_hex(MODULUS_HEX);
 
 /// Element of the secp521r1 base field used for curve coordinates.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy)]
 pub struct FieldElement(fiat_p521_tight_field_element);
+
+impl core::fmt::Debug for FieldElement {
+    /// Formatting machinery for [`FieldElement`]
+    ///
+    /// # Why
+    /// ```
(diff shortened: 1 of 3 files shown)


p521/src/arithmetic/field.rs
```diff
@@ -51,9 +51,45 @@ const MODULUS_HEX: &str = "00000000000001fffffffffffffffffffffffffffffffffffffff
...
+    /// For the above example, deriving [`core::fmt::Debug`] will result in returning 2 different strings,
+    /// which are in reality the same due to p521's unsaturated math, instead print the output as a hex string in
+    /// big-endian
+    ///
+    /// This makes debugging easier.
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let mut bytes = fiat_p521_to_bytes(&self.0);
+        bytes.reverse();
+
+        write!(f, "0x")?;
```
tarcieri, line 85: It'd probably be good to use `f.debug_tuple("FieldElement").field(...)` still.

You can use `base16ct::HexDisplay` to format the field, since it's already a hard transitive dependency by way of the `elliptic-curve` crate, which uses it for similar purposes (but now I see unfortunately does not re-export it, so it would have to be added to Cargo.toml)

MasterAwesome: https://github.com/RustCrypto/elliptic-curves/pull/949/commits/f3847272ef1bcbea430f089e5bd6b5b6f67a1657 adds this dependency

tarcieri (approved):

PR #955, opened 2023-11-08
k256: rename `lincomb_generic` to `lincomb` and expose it.

### LINKED ISSUE #921: Expose `lincomb_generic` (or: Generic Linear Combination)
The `LinearCombination` [trait](https://github.com/RustCrypto/traits/blob/eb8e409d81b7708a9ed18304c9102b5165fe4c7a/elliptic-curve/src/ops.rs#L34) allows for an optimized implementation of the linear combination of two points and scalars.

This is later generalized for `k256` with [`lincomb_generic()`](https://github.com/RustCrypto/elliptic-curves/blob/d8e44b0440dbf61b1c42fe02e4e95b4f0e55ba69/k256/src/arithmetic/mul.rs#L313C4-L313C19) which allows for an optimized implementation of a linear combination between any group of `N` points and scalars, but is unfortunately private.

Here I am stressing that users such as myself may benefit from such a generic optimized linear combination implementation, and wish that we either expose `lincomb_generic()` or, preferably, think of a suitable extension for the `LinearCombination` trait.

@tarcieri

ycscaly:
Resolves #921

### DIFF
--- a/k256/src/lib.rs
+++ b/k256/src/lib.rs
@@ -28,6 +28,8 @@
 
 #[cfg(feature = "arithmetic")]
 mod arithmetic;
+#[cfg(feature = "arithmetic")]
+pub use arithmetic::lincomb;
 
 #[cfg(feature = "ecdh")]
 pub mod ecdh;
--- a/k256/src/lib.rs
+++ b/k256/src/lib.rs
@@ -28,8 +28,6 @@
 
 #[cfg(feature = "arithmetic")]
 mod arithmetic;
-#[cfg(feature = "arithmetic")]
-pub use arithmetic::lincomb;
 
 #[cfg(feature = "ecdh")]
 pub mod ecdh;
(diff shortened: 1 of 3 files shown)


ycscaly:
@tarcieri

k256/src/lib.rs
```diff
@@ -28,6 +28,8 @@
 
 #[cfg(feature = "arithmetic")]
 mod arithmetic;
+#[cfg(feature = "arithmetic")]
+pub use arithmetic::lincomb;
```
tarcieri, line 32: I'd prefer to keep the re-export in just `arithmetic`:

```suggestion
```

ycscaly: But `arithmetic` is a private module, it will be inaccessible.
I did move it to be with the other re-exports from this module

tarcieri: Aah, my bad

PR #964, opened 2023-11-10
p521: implement hash2curve

daxpedda:
This implements the hash-to-curve proposal for P521.

I went ahead and updated the OPRF test vectors as well.

Additionally I added a test that makes sure that the constants we calculated by hand are correct and simplified them, which is now possible because of all the added const methods.

See #503 and #600 for previous implementations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -784,6 +784,7 @@ dependencies = [
  "elliptic-curve",
  "hex-literal",
  "primeorder",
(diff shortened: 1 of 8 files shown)


daxpedda:
Just rebased to get #967.

tarcieri (approved):

tarcieri:
Thanks!

PR #967, opened 2023-11-11
p521: fix panics when decoding `FieldElement`s

### LINKED ISSUE #965: P521 `FieldElement` conversions can panic
Minimal test case:
```rust
let _ = PublicKey::<NistP521>::from_sec1_bytes(&[
    2, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0,
]);
```
... instead of returning an error it panics with:
```
Input bound for the result[65] is [0x0 ~> 0x1]
```

I believe this can be boiled down to:
```rust
FieldElement::from_uint_unchecked(U576::from_be_slice(&[
    0, 0, 0, 0, 0, 2, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
]));
```
... where `FieldElement` is in `p521::arithmetic::field::FieldElement`, but is actually not public.

tarcieri:
Some `debug_assert!`s were getting triggered on inputs that need to be handled without panicking.

Fixes #965

cc @MasterAwesome

### DIFF
--- a/p521/src/arithmetic/field.rs
+++ b/p521/src/arithmetic/field.rs
@@ -628,6 +628,7 @@ impl<'a> Product<&'a FieldElement> for FieldElement {
 mod tests {
     use super::FieldElement;
     use elliptic_curve::ff::PrimeField;
+    use hex_literal::hex;
     use primeorder::{
         impl_field_identity_tests, impl_field_invert_tests, impl_field_sqrt_tests,
         impl_primefield_tests,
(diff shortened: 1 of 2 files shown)


MasterAwesome:
Hmm, so invalid points can still be created through other functions when the name doesn't contain the suffix `_unchecked` like from_hex. And those functions don't return an `Option` or a `Result`. Should the `u576_to_le_bytes` return an Option?

tarcieri:
`from_hex` is documented as not performing the check. Are you seeing anything else which isn't documented as such?

`u576_to_le_bytes` needs to operate in constant time for `from_bytes` to be constant-time.

MasterAwesome:
Oh sorry, I just saw that it's a `pub(crate)` and that it's documented. Looks good to me!

Issue #943, opened 2023-10-26
Batch Conversion of Projective -> Affine Points


ycscaly:
In #937 I explained how expensive the `to_affine()` function is in `k256`, and how it can become a bottleneck in certain situations. My suggestion to expose the projective coordinates got some backfire:

              Depending on the implementation, exposing the projective point can leak information about the discrete logarithm (https://www.iacr.org/archive/eurocrypt2004/30270258/projective.pdf)

Instead, the suggestion was to have a batch conversion:

              It sounds to me like a batched projective->affine conversion would solve your problem without introducing a footgun. Typically a batch conversion of any size is basically as efficient as a single, since you only have to do a single field inversion for the whole batch.

_Originally posted by @randombit in https://github.com/RustCrypto/elliptic-curves/issues/937#issuecomment-1758008352_

I am for this, and wondering:
1. where can I find an algorithm for this method that I could implement @randombit
2. if there is a specific interface you wish for me to expose @tarcieri

randombit:
Regarding the algorithm:

The Jacobian coordinates of an elliptic curve point `(x,y)` are `(X,Y,Z)` where `x = X / Z^2` and `y = Y / Z^3`. To convert from projective to affine coordinates, we must compute the modular inverse of `Z`, which is expensive. Given `Z^-1` we can cheaply compute `Z^-2` and `Z^-3` and so compute the affine coordinates `x` and `y`.

If we wish to convert many points at once, then we have to compute the inverse of `Z_i` for `i \in 0..n`

There is a trick for computing many modular inverses at once, often called "Montgomery's trick", which reduces the problem of computing `n` inverses to computing a single inversion, plus some storage and `O(n)` extra multiplications. There is a description of Montgomery's trick in https://iacr.org/archive/pkc2004/29470042/29470042.pdf section 2.2 that explains this part as well as I could. It is also described in Algorithm 2.26 of "Guide to Elliptic Curve Cryptography".

So the batch conversion works by applying Montgomery's trick to reduce the `n` inversions of the `Z` coordinates to just a single inversion, then using that to extract each individual inverse `Z_i^-1` and computing the affine coordinates.

There is an implementation of this technique in C++ [here](https://github.com/randombit/botan/blob/master/src/lib/pubkey/ec_group/ec_point.cpp#L388) that demonstrates how it all comes together.

ycscaly:
> Regarding the algorithm:
> 
[...]

PR #971, opened 2023-11-12
k256+primeorder: add `BatchInvert` and `BatchNormalize` impls

ycscaly:
Addresses #943


k256/src/arithmetic.rs
```diff
@@ -33,10 +35,77 @@ pub(crate) const CURVE_EQUATION_B: FieldElement = FieldElement::from_bytes_unche
     0, 0, 0, 0, 0, 0, 0, CURVE_EQUATION_B_SINGLE as u8,
 ]);
 
+impl ToAffineBatch for Secp256k1 {
```
ycscaly, line 38: I know this code isn't DRY, but the call to `invert_batch_generic` in one case and `invert_batch` in the other made it hard to DRY-out. Suggestions are welcome

ycscaly:
@tarcieri there are the tests for inversion, and also implemented the projective to affine batch conversion.
However, this is only for `k256` for now. I can do it for other crates, but its a bit of repetitive work - perhaps we leave it so that if any community request for it comes up, they can copy my logic?

k256/src/arithmetic/projective.rs
```diff
@@ -250,6 +253,71 @@ impl From<AffinePoint> for ProjectivePoint {
     }
 }
 
+impl Normalize for ProjectivePoint {
```
ycscaly, line 256: I know this code isn't DRY, but the call to invert_batch_generic in one case and invert_batch in the other made it hard to DRY-out. Suggestions are welcome

tarcieri:
> However, this is only for k256 for now. I can do it for other crates, but its a bit of repetitive work

Once you have `k256` nailed down, the only other thing that needs it is `primeorder`, then all of the other crates will benefit.

k256/src/arithmetic/projective.rs
```diff
@@ -381,12 +448,30 @@ impl GroupEncoding for ProjectivePoint {
...
+}
+#[cfg(feature = "alloc")]
 impl Curve for ProjectivePoint {
     type AffineRepr = AffinePoint;
 
     fn to_affine(&self) -> AffinePoint {
         ProjectivePoint::to_affine(self)
     }
+
+    fn batch_normalize(p: &[Self], q: &mut [Self::AffineRepr]) {
```
tarcieri, line 467:
```suggestion
impl Curve for ProjectivePoint {
    type AffineRepr = AffinePoint;

    fn to_affine(&self) -> AffinePoint {
        ProjectivePoint::to_affine(self)
    }

    #[cfg(feature = "alloc")]
    fn batch_normalize(p: &[Self], q: &mut [Self::AffineRepr]) {
```

ycscaly:
> > However, this is only for k256 for now. I can do it for other crates, but its a bit of repetitive work
> 
> Once you have `k256` nailed down, the only other thing that needs it is `primeorder`, then all of the other crates will benefit.

Ok ok, I did it

k256/src/arithmetic/projective.rs
```diff
@@ -250,6 +253,70 @@ impl From<AffinePoint> for ProjectivePoint {
...
+                // If the `z` coordinate is non-zero, we can use it to invert;
+                // otherwise it defaults to the `IDENTITY` value in initialization.
+                affine_points[i] = points[i].to_affine_internal(zs_inverses[i])
+            }
+        }
+
+        affine_points
+    }
+
+    #[cfg(not(feature = "alloc"))]
```
ycscaly, line 284: @tarcieri I'm not sure how to get rid of this.
And if I change this to return `vec`, it will not compile obviously as we're in `(not(feature = "alloc")`

ycscaly:
@tarcieri should be good now

tarcieri:
Why did you remove all of the `Invert` impls on scalars?

ycscaly:
> Why did you remove all of the `Invert` impls on scalars?

Because I added a macro that implements it, so they were double implementations

tarcieri:
Aah ok, so why didn't you use that macro for the `FieldElement` impls of `Invert`?

primeorder/src/field.rs
```diff
@@ -317,6 +317,14 @@ macro_rules! impl_mont_field_element_arithmetic {
             }
         }
 
+        impl $crate::elliptic_curve::ops::Invert for $fe {
+            type Output = $crate::elliptic_curve::subtle::CtOption<Self>;
+
+            fn invert(&self) -> $crate::elliptic_curve::subtle::CtOption<Self> {
+                self.invert()
+            }
+        }
```
tarcieri, line 326: I think this is going to wind up being a breaking change, since it would conflict with the existing impls of `Invert` in existing crate releases of the downstream elliptic curve implementations.

It's a good idea for the next breaking release, though.

ycscaly: So I should remove the macro and implement it manually for whatever type it is missing?

tarcieri: Yeah, there shouldn't be any missing impls as `Invert` is required in the bounds for `elliptic_curve::Curve::Scalar`

ycscaly: But I need it for FieldElement. There were missing implementations for sure

ycscaly: Ok fixed

tarcieri:
In places where you reference `alloc::vec::Vec`, particularly the `projective.rs` files, I think the code would be a bit less noisy with a conditional import:

```rust
#[cfg(feature = "alloc")]
use alloc::vec::Vec;
```

k256/src/arithmetic/projective.rs
```diff
@@ -250,6 +253,71 @@ impl From<AffinePoint> for ProjectivePoint {
...
+                zs[i] = points[i].z;
+            }
+        }
+
+        // This is safe to unwrap since we assured that all elements are non-zero
+        let zs_inverses: alloc::vec::Vec<_> =
+            <FieldElement as Invert>::batch_invert(zs.as_slice()).unwrap();
+
+        let mut affine_points: alloc::vec::Vec<_> =
+            (0..points.len()).map(|_| AffinePoint::IDENTITY).collect();
```
tarcieri, line 302:
```suggestion
        let mut affine_points = vec![AffinePoint::IDENTITY; points.len()];
```

k256/src/arithmetic/projective.rs
tarcieri: I can think of some possibilities here but they might require some experimentation. This seems fine for now.

tarcieri: This is what I had in mind, FWIW https://github.com/RustCrypto/traits/pull/1376/files#r1393273429

k256/src/arithmetic/projective.rs
```diff
@@ -250,6 +256,72 @@ impl From<AffinePoint> for ProjectivePoint {
...
+            if points[i].z != FieldElement::ZERO {
+                // If the `z` coordinate is non-zero, we can use it to invert;
+                // otherwise it defaults to the `IDENTITY` value in initialization.
+                affine_points[i] = points[i].to_affine_internal(zs_inverses[i])
+            }
+        }
+
+        affine_points.into_iter().collect()
+    }
+}
```
tarcieri, line 323: I would expect this to conflict with the blanket impl

ycscaly: It does not, since the return value of the function is different.
Look at the changes I needed to do to the tests

tarcieri: Oh sorry, the blanket impl was for `BatchInvert`, my bad

Cargo.toml
```diff
@@ -16,3 +16,6 @@ members = [
 
 [profile.dev]
 opt-level = 2
+
+[patch.crates-io]
+elliptic-curve = { git = "https://github.com/ycscaly/traits", branch = "batch_invert" }
```
tarcieri, line 21: Now that the `branch_invert` branch is merged:

```suggestion
elliptic-curve = { git = "https://github.com/RustCrypto/traits" }
```

tarcieri (approved):

tarcieri (approved):

ycscaly:
Thanks!

PR #975, opened 2023-11-15
k256+primeorder: add `batch_normalize_generic`

tarcieri:
Adds a generic implementation of `batch_normalize` which abstracts over backing storages.

This is just implemented in `k256` for now, but could potentially be further abstracted and extracted into `elliptic-curve`.

### DIFF
--- a/k256/src/arithmetic/projective.rs
+++ b/k256/src/arithmetic/projective.rs
@@ -256,69 +256,57 @@ impl From<AffinePoint> for ProjectivePoint {
     }
 }
 
-impl<const N: usize> BatchNormalize<&[ProjectivePoint; N]> for ProjectivePoint {
+impl<const N: usize> BatchNormalize<[ProjectivePoint; N]> for ProjectivePoint {
     type Output = [Self::AffineRepr; N];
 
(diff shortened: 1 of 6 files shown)


k256/src/arithmetic/projective.rs
```diff
@@ -256,69 +256,58 @@ impl From<AffinePoint> for ProjectivePoint {
...
+}
 
-        for i in 0..points.len() {
-            if points[i].z != FieldElement::ZERO {
-                // Even a single zero value will fail inversion for the entire batch.
-                // Put a dummy value (above `FieldElement::ONE`) so inversion succeeds
-                // and treat that case specially later-on.
-                zs[i] = points[i].z;
-            }
+fn batch_normalize_generic<P, Z, ZI, O>(points: &P, zs: &mut Z, out: &mut O)
```
tarcieri, line 282: @ycscaly some gnarly bounds, but generic over the container types at least

ycscaly:
Yeah I see how our new trait design enables this. I've wanted something like this and am happy with your suggestion, so we ended up doing something better than originally intended.

PR #1011, opened 2024-01-10
Bump `elliptic-curve` crate to v0.14.0-pre.2

tarcieri:
Begins the next round of breaking changes by upgrading `elliptic-curve` to the latest prerelease.

Notably this includes a migration from `generic-array` to `hybrid-array`, and also bumps `crypto-bigint` to the latest prerelease.


k256/src/arithmetic/scalar.rs
```diff
@@ -120,8 +120,8 @@ impl Scalar {
     /// Right shifts the scalar.
     ///
     /// Note: not constant-time with respect to the `shift` parameter.
-    pub fn shr_vartime(&self, shift: usize) -> Scalar {
-        Self(self.0.shr_vartime(shift))
+    pub fn shr_vartime(&self, shift: u32) -> Scalar {
+        Self(self.0.wrapping_shr_vartime(shift))
     }
```
tarcieri, line 125: @fjarri could use a little help on this one... this seems broken after updating `crypto-bigint` to v0.6.0-pre.9

tarcieri: See example failures here: https://github.com/RustCrypto/elliptic-curves/actions/runs/7480942676/job/20361467871

tarcieri: This seems to be a legitimate bug. Going to try to fix it upstream.

bign256/src/arithmetic/scalar.rs
```diff
@@ -154,7 +154,7 @@ impl Invert for Scalar {
 
 impl IsHigh for Scalar {
     fn is_high(&self) -> Choice {
-        const MODULUS_SHR1: U256 = BignP256::ORDER.shr_vartime(1);
+        const MODULUS_SHR1: U256 = BignP256::ORDER.wrapping_shr_vartime(1);
```
tarcieri, line 157: I tried using `.shr(1)` and `.shr(1u32)` here (and for the same constant in other curves) but I just can't get it to use the `const fn` over the trait:

```
error[E0015]: cannot call non-const fn `<elliptic_curve::crypto_bigint::Uint<4> as Shr<u32>>::shr` in constants
   --> bign256/src/arithmetic/scalar.rs:157:52
    |
157 |         const MODULUS_SHR1: U256 = BignP256::ORDER.shr(1u32);
    |                                                    ^^^^^^^^^
    |
    = note: calls in constants are limited to constant functions, tuple structs and tuple variants
```

tarcieri: Curiously, it works just fine here: https://github.com/RustCrypto/crypto-bigint/blob/5ff769f/src/lib.rs#L56

k256/src/arithmetic/scalar.rs
tarcieri: Fixed https://github.com/RustCrypto/crypto-bigint/pull/552

bign256/src/arithmetic/scalar.rs
tarcieri: Well, `shl_vartime` works, now that I've added it (back)

PR #1029, opened 2024-02-07
Batch Normalization `normalizes_to_zero()` bug fix

ycscaly:
This PR fixes a bug in batch normalization, where non-normalized identity points would be misidentified, and therefore batch normalization would panic in `unwrap`ing a `None` batch inversion result.

See thread in https://github.com/RustCrypto/elliptic-curves/issues/943#issuecomment-1925992540.

@tarcieri

### DIFF
--- a/primeorder/src/projective.rs
+++ b/primeorder/src/projective.rs
@@ -393,7 +393,7 @@ where
         // Even a single zero value will fail inversion for the entire batch.
         // Put a dummy value (above `FieldElement::ONE`) so inversion succeeds
         // and treat that case specially later-on.
-        zs.as_mut()[i].conditional_assign(&points[i].z, !points[i].z.normalizes_to_zero());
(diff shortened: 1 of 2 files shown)


primeorder/src/projective.rs
```diff
@@ -405,7 +405,7 @@ where
         out[i] = C::AffinePoint::conditional_select(
             &points[i].to_affine_internal(zs_inverses.as_ref()[i]),
             &C::AffinePoint::IDENTITY,
-            points[i].z.ct_eq(&C::FieldElement::ZERO),
+            points[i].z.normalizes_to_zero(),
```
tarcieri, line 408: Only `k256` uses lazy normalization, so this isn't needed here

ycscaly: Fixed

ycscaly:
Can we do a minor release to fix this?

tarcieri:
@ycscaly I’ll have to backport it, but sure

PR #1037, opened 2024-03-23
p256: oid feature

yayanyang:
If Oid support is not added, SigningKey<NistP256> cannot be passed to CertificateBuilder::new as the cert_signer parameter


tarcieri:
Note that this PR is to the development version of `p256`.

In these prereleases we have already enabled the `oid` feature of `sha2` by default, so this feature won’t accomplish anything.

It’s also unclear which “oid” this is referring to, since several things (including the curves themselves) can be identified by OIDs.

Issue #1038, opened 2024-03-31
Bug(deps) p521 crate requires ecdsa 16.8


joshka:
Hey, doing a minimal versions check on another library, I stumbled on p521 not being able to compile against the stated dependencies.

p521 0.13.3 requires ecdsa 0.16.8 , but specifies 0.16.5

<details><summary>Error Details</summary>
<p>

```
error[E0432]: unresolved import `ecdsa_core::hazmat::sign_prehashed`
  --> /Users/joshka/.cargo/registry/src/index.crates.io-6f17d22bba15001f/p521-0.13.3/src/ecdsa.rs:48:30
   |
48 |         hazmat::{bits2field, sign_prehashed, SignPrimitive, VerifyPrimitive},
   |                              ^^^^^^^^^^^^^^ no `sign_prehashed` in `hazmat`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `p521` (lib) due to 1 previous error
```

</p>
</details> 

Change that introduces the sign_prehashed function
https://github.com/RustCrypto/signatures/commit/d7557a62bc1bb43944d89778fcbb7d4c8b3c9118

tarcieri:
Aah, we normally run a `minimal-versions` check but it seems this was missed for `p521`.

A fix would require a backport since `master` currently has the next prerelease versions, but I can get to it at some point.

joshka:
I figured that might be the case seeing the -pre dep in the Cargo.toml. I wasn't sure quite what your release process would be for something like that.

The workaround for any consuming packages in the mean time (assuming they care about such things) is to just add ecdsa 0.16.8 as a dependency, so it's not a huge deal.

> Aah, we normally run a minimal-versions check but it seems this was missed for p521.

Sounds like this might be worth adding to your CI.

PR #1039, opened 2024-03-31
ci: add minimal versions check

joshka:
Related to https://github.com/RustCrypto/elliptic-curves/issues/1038


.github/workflows/check-minimal-versions.yml
```diff
@@ -0,0 +1,28 @@
+# This workflow checks that the project builds and tests successfully with the
+# minimal versions of all dependencies that are compatible with the current
+# Cargo.lock file. This is useful for ensuring that the project is not
+# accidentally relying on newer versions of dependencies than it actually
+# needs.
```
tarcieri, line 5: We already have a standard shared action for checking minimal versions here: https://github.com/RustCrypto/actions/blob/master/.github/workflows/minimal-versions.yml

tarcieri:
Note that as we're working on prereleases we tend to disable these checks as they're not particularly compatible with things like `[patch.crates-io]` directives that pull in crates from `git`

joshka:
> Note that as we're working on prereleases we tend to disable these checks as they're not particularly compatible with things like `[patch.crates-io]` directives that pull in crates from `git`

Makes sense. That said, would you prefer to have a failing check to ignore or a missing check to forget for this?

I've updated the workflow to point at the the standard shared action - feel free to close this though if the idea is incompatible with your general use case. I see that the call to minimal versions is commented out in at least one other repo (with a note to re-enable upon release), perhaps this is an approach that's somewhere in the middle.

tarcieri:
It will definitely require some tuning as `cargo hack` is, in cases like this, incredible overkill and generating nearly 10,000 feature combinations (though it seems it is finding some valid build failures, so it's separately something we should generally add)

joshka:
> It will definitely require some tuning as `cargo hack` is, in cases like this, incredible overkill and generating nearly 10,000 feature combinations (though it seems it is finding some valid build failures, so it's separately something we should generally add)

Adding an initial `cargo check` that runs against all features and fails fast would probably fix that problem, but so too might the `-Zdirect-minimal-versions` option instead of `-Zminimal-versions`. The former fails immediately without even attempting to run cargo check. If there's multiple failures, it's indeterminate which specific dependency will cause the process to halt. This is two runs both from within the k256 folder:

```text
❯ cargo +nightly update -Zdirect-minimal-versions
    Updating crates.io index
error: failed to select a version for `proptest`.
    ... required by package `k256 v0.14.0-pre (/Users/joshka/local/elliptic-curves/k256)`
versions that meet the requirements `^1.4` are: 1.4.0

all possible versions conflict with previously selected packages.

  previously selected package `proptest v1.0.0`
    ... which satisfies dependency `proptest = "^1"` of package `bign256 v0.14.0-pre (/Users/joshka/local/elliptic-curves/bign256)`

failed to select a version for `proptest` which could resolve this conflict
```
```
❯ cargo +nightly update -Zdirect-minimal-versions
    Updating crates.io index
error: failed to select a version for `criterion`.
    ... required by package `p521 v0.14.0-pre (/Users/joshka/local/elliptic-curves/p521)`
versions that meet the requirements `^0.5.1` are: 0.5.1

all possible versions conflict with previously selected packages.

  previously selected package `criterion v0.5.0`
    ... which satisfies dependency `criterion = "^0.5"` of package `bign256 v0.14.0-pre (/Users/joshka/local/elliptic-curves/bign256)`

failed to select a version for `criterion` which could resolve this conflict
```

joshka:
What was your eventual preference on this - going through my old PRs to close out any that don't really have a next step and I wasn't sure about this one. Closing for now. Feel free to reopen if there's more that you want out of it.

tarcieri:
Might be worth adding after stable releases

PR #1041, opened 2024-04-15
Add support for BIP340 signtures where the messages are not exactly 32 bytes

### LINKED ISSUE #1040: BIP340 Schnorr should accept arbitrary length messages
[BIP340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) was modified last year to [allow arbitrary length inputs](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#messages-of-arbitrary-size) instead of just 32 bytes. But `verify_prehash` continues to require the message be exactly 32 bytes. Curiously (IIUC) it's already possible to sign messages longer than 32 bytes, as I don't see any length check in that code. As I understand it all that's required is to remove the line 

```
let prehash: [u8; 32] = prehash.try_into().map_err(|_| Error::new())?;
```

and pass the arbitrarily sized prehash to the inner tagged hash.

randombit:
This was an extension to BIP340 that was introduced in 2022. The official [BIP340 test vectors](https://github.com/bitcoin/bips/blob/master/bip-0340/test-vectors.csv) do contain data for this case (tests 15-18) which are included here.

Fixes #1040

### DIFF
--- a/k256/src/schnorr/signing.rs
+++ b/k256/src/schnorr/signing.rs
@@ -78,7 +78,7 @@ impl SigningKey {
     /// The preferred interfaces are the [`Signer`] or [`RandomizedSigner`] traits.
     pub fn sign_prehash_with_aux_rand(
         &self,
-        msg_digest: &[u8; 32],
+        msg_digest: &[u8],
         aux_rand: &[u8; 32],
     ) -> Result<Signature> {
         let mut t = tagged_hash(AUX_TAG).chain_update(aux_rand).finalize();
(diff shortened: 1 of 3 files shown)


k256/src/schnorr/signing.rs
```diff
@@ -78,7 +78,7 @@ impl SigningKey {
     /// The preferred interfaces are the [`Signer`] or [`RandomizedSigner`] traits.
     pub fn sign_prehash_with_aux_rand(
```
tarcieri, line 79: Perhaps this method should be renamed to reflect the input is no longer necessarily a prehash?

Something like `sign_raw` perhaps?

k256/src/schnorr/verifying.rs
```diff
@@ -64,7 +64,6 @@ impl PrehashVerifier<Signature> for VerifyingKey {
         prehash: &[u8],
```
tarcieri, line 64: Perhaps this should be extracted into something like `verify_raw` and this argument renamed to something which reflects it isn’t necessarily a prehash

k256/src/schnorr/signing.rs
randombit: Yeah makes sense, I renamed to `sign_raw`. I didn't know if you wanted to keep any compatability forwarding (eg a `#[deprecated]` `sign_prehash_with_aux_rand` that forwards to `sign_raw`), so I left it out for now.

k256/src/schnorr/verifying.rs
randombit: I extracted out to a new method on `VerifyingKey` with some scary warnings in the rustdoc, ptal

k256/src/schnorr/signing.rs
tarcieri: That's fine. We can make a note in the changelog.

PR #1055, opened 2024-06-07
cargo: point `repository` metadata to clonable URLs

lucab:
This tweaks the `repository` fields in Cargo metadata in order to use the correct (i.e. git clonable) URL. The existing GitHub webUI URLs for each package have been retained and moved to `homepage` fields.


lucab:
For reference, without this PR the remote fetch would fail with:
```
$ git clone https://github.com/RustCrypto/elliptic-curves/tree/master/bign256

Cloning into 'bign256'...
fatal: repository 'https://github.com/RustCrypto/elliptic-curves/tree/master/bign256/' not found
```

tarcieri (approved):

lucab:
@tarcieri I think I've now completed the tour of all repos under `RustCrypto` and fixed all the misaligned values. Sorry for the sudden spike of PRs and thanks for promptly merging all of those!

tarcieri:
@lucab thanks for getting them corrected! I know people are trying to write tooling to automatically verify this information so it's good to have them valid

Issue #1067, opened 2024-08-06
sm2: Support for SM2PKE


heliannuuthus:

heliannuuthus:
@tarcieri  I can provide an implementation of SM2PKE, but please let me know what the conventions or specifications :)

tarcieri:
@heliannuuthus the spec is linked from here: https://github.com/RustCrypto/elliptic-curves/tree/master/sm2#about-sm2

I'm not terribly familiar with it but can review a PR to add it.

heliannuuthus:
I'll give it a try.

heliannuuthus:
> @heliannuuthus the spec is linked from here: https://github.com/RustCrypto/elliptic-curves/tree/master/sm2#about-sm2
> 
> I'm not terribly familiar with it but can review a PR to add it.

I'll give it a try.

PR #1069, opened 2024-08-12
sm2: add SM2PKE support

heliannuuthus:
1. encrypt
2. decrypt
3. encrypt_der
4. decrypt_der


heliannuuthus:
#1067

tarcieri:
Looks great on a first pass. I'll try to do a more in-depth pass soon.

sm2/src/pke/encrypting.rs
```diff
@@ -191,7 +191,7 @@ fn encrypt(
 fn next_k(bit_length: u32) -> U256 {
     loop {
         let k = U256::random_bits(&mut rand_core::OsRng, bit_length);
-        if k.is_zero().unwrap_u8() == 0 && k <= Sm2::ORDER {
+        if !<elliptic_curve::subtle::Choice as Into<bool>>::into(k.is_zero()) && k < Sm2::ORDER {
```
tarcieri, line 194:
```suggestion
        if !bool::from(k.is_zero()) && k < Sm2::ORDER {
```

heliannuuthus: so sad~, why didn't i think of that :(

sm2/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sm2"
-version = "0.14.0-pre"
+version = "0.14.1-pre"
```
tarcieri, line 3: You don't need to bump this. We've never released `0.14.0-pre`. It's just pre-bumped like this to denote breaking changes.

sm2/Cargo.toml
```diff
@@ -33,13 +33,14 @@ proptest = "1"
...
 [features]
-default = ["arithmetic", "dsa", "pem", "std"]
+default = ["arithmetic", "dsa", "pke", "pem", "std"]
 alloc = ["elliptic-curve/alloc"]
 std = ["alloc", "elliptic-curve/std", "signature?/std"]
 
 arithmetic = ["dep:primeorder", "elliptic-curve/arithmetic"]
 bits = ["arithmetic", "elliptic-curve/bits"]
 dsa = ["arithmetic", "dep:rfc6979", "dep:signature", "dep:sm3"]
+pke = ["arithmetic", "dep:rfc6979", "dep:sm3"]
```
tarcieri, line 43: The `dep:rfc6979` doesn't appear to be necessary here:

```suggestion
pke = ["arithmetic", "dep:sm3"]
```

sm2/src/pke/encrypting.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        // A4: compute point [𝑘]𝑃𝐵 = (𝑥2, 𝑦2)
+        hpb = (s * k).to_affine();
+
+        // A5: compute 𝑡 = 𝐾𝐷𝐹(𝑥2||𝑦2, 𝑘𝑙𝑒𝑛)
+        // A6: compute 𝐶2 = 𝑀 ⊕ t
+        kdf(digest, hpb, &mut c2)?;
+
+        // // If 𝑡 is an all-zero bit string, go to A1.
+        // if all of t are 0, xor(c2) == c2
```
tarcieri, line 168: This comment seems a little odd

sm2/src/pke.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+/// Performs key derivation using a hash function and elliptic curve point.
+fn kdf(hasher: &mut dyn DynDigest, kpb: AffinePoint, c2: &mut [u8]) -> Result<()> {
+    let klen = c2.len();
+    let mut ct: i32 = 0x00000001;
+    let mut offset = 0;
+    let digest_size = hasher.output_size();
+    let mut ha = vec![0u8; digest_size];
```
tarcieri, line 139: This won't work on heapless `no_std` targets, nor is the `alloc` feature being enabled by the `pke` feature.

It seems like you could get rid of these dynamic allocations if you switched from `DynDigest` to a generic `D` parameter like you have for `decrypt_digest` and bound on the [`OutputSizeUser`](https://docs.rs/digest/latest/digest/core_api/trait.OutputSizeUser.html) trait, which will give you access to a compile-time output size which can be used to allocate an appropriately sized array.

sm2/src/pke/decrypting.rs
```diff
@@ -0,0 +1,216 @@
...
+        Mode::C1C2C3 => c.split_at(c.len() - digest_size),
+    };
+
+    // B4: compute 𝑡 = 𝐾𝐷𝐹(𝑥2 ∥ 𝑦2, 𝑘𝑙𝑒𝑛)
+    // B5: get 𝐶2 from 𝐶 and compute 𝑀′ = 𝐶2 ⊕ t
+    let mut c2 = c2.to_owned();
+    kdf(hasher, c1_point, &mut c2)?;
+
+    // compute 𝑢 = 𝐻𝑎𝑠ℎ(𝑥2 ∥ 𝑀′∥ 𝑦2).
+    let mut u = vec![0u8; digest_size];
```
tarcieri, line 195: Similar comment to [this one](https://github.com/RustCrypto/elliptic-curves/pull/1069/files#r1733348316) regarding the dynamic allocations. They're unnecessary.

sm2/src/pke/encrypting.rs
heliannuuthus: I want to explain why we need to determine that xor(c2) == c2，but the description seems odd

tarcieri: At the very least, there's a double `// //`

It seems like `If 𝑡 is an all-zero bit string` and `if all of t are 0` describe the same condition in slightly different ways.

sm2/src/pke.rs
heliannuuthus: > I once saw an [issue](https://github.com/RustCrypto/traits/issues/1574) related to using `OutputSizeUser::USIZE` to create a new array.

I tried to solve this problem with the [`hybird_array`](https://github.com/RustCrypto/hybrid-array/), but I needed a little time to adjust the other places where the `alloc` feature was used

tarcieri (approved):
Will go ahead and approve this and fix up my nits retroactively

PR #1072, opened 2024-08-29
k256: add support for non-32-byte BIP340 signatures (#1041)

### LINKED ISSUE #1040: BIP340 Schnorr should accept arbitrary length messages

randombit:
This is a backport of #1041 onto the 0.13.3 tag, with some small changes to accommodate SemVer.

Fixes #1040

### DIFF
--- a/k256/src/schnorr.rs
+++ b/k256/src/schnorr.rs
@@ -270,13 +270,74 @@ mod tests {
             assert_eq!(sk.verifying_key().to_bytes().as_slice(), &vector.public_key);
 
             let sig = sk
-                .sign_prehash_with_aux_rand(&vector.message, &vector.aux_rand)
+                .sign_raw(&vector.message, &vector.aux_rand)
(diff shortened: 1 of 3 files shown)


tarcieri:
Published in `k256` v0.13.4

PR #1099, opened 2024-11-06
sm2: Fix heap allocation

MaxKingPor:
Fix sm2 heap allocation


tarcieri (approved):

MaxKingPor:
It seems that mine is quite different from the main branch, so I need to make some revisions.

sm2/Cargo.toml
```diff
@@ -18,24 +18,30 @@ edition = "2024"
...
+    "rand_core",
+    "digest",
+] }
 sm3 = { version = "=0.5.0-pre.5", optional = true, default-features = false }
 
 [dev-dependencies]
 hex-literal = "1"
 proptest = "1"
 rand_core = { version = "0.9", features = ["os_rng"] }
+rand = "0.9"
```
tarcieri, line 40: It seems like you're adding this to change the documentation from using `OsRng` to `rand::rng`.

IMO `OsRng` is the safer option for things like key generation. `rand::rng` can use a userspace RNG which can introduce problems like fork safety.

tarcieri: Sidebar: we should also add APIs that encapsulate this so the user doesn't need to pass an explicit RNG at all

sm2/src/pke.rs
```diff
@@ -1,73 +1,52 @@
...
-//! };
-//!
+#![cfg_attr(feature = "alloc", doc = "```")]
+#![cfg_attr(not(feature = "alloc"), doc = "```ignore")]
+//! use sm2::pke::{EcDecrypt, EcEncrypt, Cipher, Mode};
+//! use sm2::SecretKey;
+//! use rand_core::OsRng;
 //! // Encrypting
-//! let secret_key = SecretKey::try_from_rng(&mut OsRng).unwrap(); // serialize with `::to_bytes()`
+//! let secret_key = SecretKey::try_from_rng(&mut OsRng).unwrap();
```
tarcieri, line 10: Would prefer the original comment be retained:

```suggestion
//! let secret_key = SecretKey::try_from_rng(&mut OsRng).unwrap(); // serialize with `::to_bytes()`
```

sm2/src/pke/decrypting.rs
```diff
@@ -1,216 +1,108 @@
...
+    sec1::{ModulusSize, ToEncodedPoint},
 };
-use primeorder::PrimeField;
-
-use sm3::{Digest, Sm3, digest::DynDigest};
-
-use super::{Cipher, Mode, encrypting::EncryptingKey, kdf, vec};
-/// Represents a decryption key used for decrypting messages using elliptic curve cryptography.
-#[derive(Clone)]
-pub struct DecryptingKey {
```
tarcieri, line 24: I'd prefer these names be retained, as they're more consistent with our other crates which identify key types by function

sm2/src/pke/decrypting.rs
```diff
@@ -1,216 +1,108 @@
...
-    mode: Mode,
+use primeorder::PrimeCurveParams;
+
+use signature::digest::{Digest, FixedOutputReset, Output};
+use sm3::Sm3;
+
+use super::{Cipher, kdf};
+
+/// Decrypt messages using elliptic curve cryptography.
+pub trait EcDecrypt<C>
```
tarcieri, line 15: Not sure why a trait is necessary where when it has only one impl. What's the intended use case?

tarcieri:
Why did you close this?

MaxKingPor:
> Why did you close this?

My recent code modifications don’t align with your style guidelines.

tarcieri:
It'd be great if you could add `no_std` support while retaining something closer to the existing library API. Is that something you are still interested in?

MaxKingPor:
> It'd be great if you could add `no_std` support while retaining something closer to the existing library API. Is that something you are still interested in?

Sorry, I haven’t had much time to focus on this lately.

PR #1107, opened 2024-12-11
Fix spelling error

dannbbb1:
Fixed a spelling error where "compactabtility" was used instead of the correct spelling "compatibility.


andrewwhitehead:
"compactability" was used in the PR, which had to do with compacting points.

tarcieri:
Yes, the current spelling is correct

PR #1119, opened 2025-02-07
primeorder: use `PrimeField::NUM_BITS` to bound the scalar size during point-scalar multiplication

fjarri:
This probably won't affect actual curves much, but it is useful for test curves which have a small order, but the representation size still has to be large. See also https://github.com/RustCrypto/formats/pull/1649

### DIFF
--- a/primeorder/src/projective.rs
+++ b/primeorder/src/projective.rs
@@ -10,7 +10,7 @@ use core::{
 };
 use elliptic_curve::{
     array::ArraySize,
-    bigint::{ArrayEncoding, Bounded},
+    bigint::ArrayEncoding,
     group::{
         self,
         cofactor::CofactorGroup,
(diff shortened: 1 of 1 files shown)


fjarri:
Actually, this should speed up some regular curves too, e.g. `p521` which uses `U576` for its scalars.

tarcieri (approved):

tarcieri:
Sidebar: seems like we could also add a `mul_vartime` which bounds by the bit length of the input scalar

primeorder/src/projective.rs
```diff
@@ -122,7 +122,7 @@ where
         }
 
         let mut q = Self::IDENTITY;
-        let mut pos = C::Uint::BITS as usize - 4;
+        let mut pos = (<Scalar<C> as PrimeField>::NUM_BITS.div_ceil(8) * 8) as usize - 4;
```
fjarri, line 125: Actually, now that I think about it, `.div_ceil(4) * 4` would be more accurate, since we need an integer number of window sizes, not bytes.

tarcieri: Maybe extract a function for calculating this? It could also be used by a hypothetical `mul_vartime`

PR #1121, opened 2025-02-14
Add ed448

mikelodder7:
This adds the Ed448-Goldilocks curve and signing implementation from the latest commit at [Mikelodder7/Ed448-Goldilocks](https://github.com/mikelodder7/Ed448-Goldilocks).


ed448/rustfmt.toml
```diff
@@ -0,0 +1,5 @@
+newline_style = "Unix"
```
dignifiedquire, line 1: probably drop this file, to align with the rest of the repo

ed448/src/curve/montgomery.rs
```diff
@@ -0,0 +1,236 @@
+// The original file was a part of curve25519-dalek.
+// Copyright (c) 2016-2019 Isis Lovecruft, Henry de Valence
+// Copyright (c) 2020 Kevaundray Wedderburn
+// See LICENSE for licensing information.
```
dignifiedquire, line 4: lets add a reference to the actual licnes file from curve25519-dalek

dignifiedquire:
@tarcieri @mikelodder7 I am not seeing much in terms of test vectors or the likes, do you happen to know where to find some?

dignifiedquire:
working on a pr to update to the current set of rustcrypto deps

mikelodder7:
Co authors
@divergentdave
@conradoplg
@stackinspector
@anonion0
@kevaundray

mikelodder7:
> @tarcieri @mikelodder7 I am not seeing much in terms of test vectors or the likes, do you happen to know where to find some?

Here's the Ed448 test vectors

https://github.com/RustCrypto/elliptic-curves/pull/1121/files#diff-16bd59631ab39d33a42cd634a7603565919c4243d9d2b21b3d975d194457f22bR351

ed448/rustfmt.toml
mikelodder7: Dropped!

dignifiedquire:
created https://github.com/mikelodder7/elliptic-curves/pull/1

mikelodder7:
> created [mikelodder7#1](https://github.com/mikelodder7/elliptic-curves/pull/1)

👀

dkg:
what's the status on this?  It would be great to see ed448 under the RustCrypto umbrella.

tarcieri:
@dkg waiting on @mikelodder7 to rebase and get the tests passing

mikelodder7:
So I have rebased but I'm not sure what to modify for bign256

stackinspector:
There's a little stuff about my previous contribution that I feel the need to renegotiate before the code gets into a serious place like RustCrypto.
The `hex_to_field` ([example](https://github.com/RustCrypto/elliptic-curves/pull/1121/files#diff-9c2883ba45e8a409b745f80fd0c0939eef6fa6643e51e3fbfb3f93c80eee5dabR184)) in test cases, which I [originally implemented with `macro_rules`](https://github.com/crate-crypto/Ed448-Goldilocks/pull/27/commits/6c839a218cf4adaba321b5faba220ea01d75fc12#diff-a37869a90c7a53189661bac24fba649872d61cf4001698361bd1a6c2e718d17cR421), passes string literial as a macro argument to `hex-literial::hex!`. But the maintainer of the original repository preferred `hex_to_field` as a function. But the problem is that `hex-literial::hex!` can only accept string literials, and the function `hex-literial::decode`, which can accept `&'static str` directly, is an internal API, even though we can do the checks that should be done, the API is theoretically unstable! Some of the other drawbacks I explained in the [comment](https://github.com/crate-crypto/Ed448-Goldilocks/pull/27#issuecomment-1554487875) at the time. At final the function version was merged, but I still prefer to switch back to the `macro_rules` implementation and pass literals.

ed448/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+description = """A pure-Rust implementation of Ed448 and Curve448 and Decaf.
+This crate also includes signing and verifying of Ed448 signatures.
+"""
+documentation = "https://docs.rs/ed448-goldilocks"
+exclude = [".gitignore", ".github/*"]
+edition = "2021"
+homepage = "https://docs.rs/ed448-goldilocks/"
+keywords = ["cryptography", "decaf", "ed448", "ed448-goldilocks"]
+license = "BSD-3-Clause"
+name = "ed448"
```
tarcieri, line 13: If we keep the name as `ed448-goldilocks`, it will [allow the `ed448` name to be used for the signature abstraction crate](https://github.com/RustCrypto/signatures/blob/master/ed448/Cargo.toml#L2)

(and that also ensures existing users get updates automatically without having to change to a new name)

dignifiedquire:
@tarcieri what's the blockers for moving this forward, given CI is green now?

dignifiedquire:
@tarcieri what would be the best place for me to put a `x448` implementation based on this?

kevaundray:
> @tarcieri what would be the best place for me to put a `x448` implementation based on this? 

I have an impl here as a reference btw: https://github.com/crate-crypto/x448

dignifiedquire:
@kevaundray very nice, so I guess we just need to update that to the latest version of this 448 crate. Do you want to keep that crate under your management and update it, or should we move it into rustcrypto as well?

kevaundray:
> @kevaundray very nice, so I guess we just need to update that to the latest version of this 448 crate. Do you want to keep that crate under your management and update it, or should we move it into rustcrypto as well? 

I think it would make sense to move it to rustcrypto as well for completeness if you haven't made an implementation

tarcieri:
@dignifiedquire I'd like to do a final review and won't have time to do that for about a week

ed448/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+readme = "README.md"
+repository = "https://github.com/RustCrypto/elliptic-curves/tree/master/ed448"
+rust-version = "1.85"
+version = "0.17.0-pre.0"
+
+[dependencies]
+crypto-bigint = { version = "0.6.0-rc.3", features = ["hybrid-array"], default-features = false }
+crypto_signature = { version = "2.3.0-pre.6", default-features = false, features = ["digest", "rand_core"], optional = true, package = "signature" }
```
dignifiedquire, line 21: if we put this on the pre release for `signature` and `rand@0.9` we will run into the issue that this is going to be unusable as in https://github.com/RustCrypto/signatures/issues/930

ed448/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+elliptic-curve = { version = "0.14.0-rc.0", features = ["arithmetic", "bits", "hash2curve", "jwk", "pkcs8", "pem", "sec1"] }
+pkcs8 = { version = "0.10", features = ["alloc"], optional = true }
+rand_core = { version = "0.9", default-features = false }
+serdect = { version = "0.3.0", optional = true }
```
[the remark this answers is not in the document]
tarcieri: huh?

ed448/Cargo.toml
tarcieri: Let’s get this merged as-is and follow up on that

tarcieri: It’s also not the same problem as this crate has existing releases with older deps whereas `ml-dsa`/`slh-dsa` do not

ed448/Cargo.toml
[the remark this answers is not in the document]
dignifiedquire: I tried updating `x448` to use this version and am getting this

```
error: failed to select a version for `serdect`.
    ... required by package `elliptic-curve v0.14.0-rc.0`
    ... which satisfies dependency `elliptic-curve = "^0.14.0-rc.0"` of package `ed448 v0.17.0-pre.0 (https://github.com/mikelodder7/elliptic-curves#363ed9dc)`
    ... which satisfies git dependency `ed448` of package `x448 v0.6.0 (/Users/dignified/opensource/x448)`
versions that meet the requirements `=0.3.0-pre.0` are: 0.3.0-pre.0

all possible versions conflict with previously selected packages.

  previously selected package `serdect v0.3.0`
    ... which satisfies dependency `serdect = "^0.3.0"` of package `ed448 v0.17.0-pre.0 (https://github.com/mikelodder7/elliptic-curves#363ed9dc)`
    ... which satisfies git dependency `ed448` of package `x448 v0.6.0 (/Users/dignified/opensource/x448)`
```

with these dependencies

```toml
ed448 = { git = "https://github.com/mikelodder7/elliptic-curves", features = ["zeroize"] }
rand_core = { version = "0.9", default-features = false, features = ["getrandom"] }
zeroize = { version = "1.8", default-features = false, optional = true }
```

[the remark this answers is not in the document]
tarcieri: For now you'll also need: 

```toml
[patch.crates-io]
elliptic-curve = { git = "https://github.com/RustCrypto/traits.git" }
```

(also `rand_core` v0.9 doesn't have a `getrandom` feature, it's `os_rng`)

ed448/Cargo.toml
dignifiedquire: fyi I made https://github.com/dignifiedquire/cx448/ based on this code, which I will publish for now, so I can use a more up to date version

ed448/Cargo.toml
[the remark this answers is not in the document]
YugenSource: Why not use rand_core 0.8, rand_core introduced more dependencies.

daxpedda (commented):
I tried to use this to implement decaf448 for [`voprf`](https://crates.io/crates/voprf), but wasn't able to get things to match the test vectors ([see the specification](https://www.rfc-editor.org/rfc/rfc9497#name-decaf448-shake256)).

I played around with it for quite a while and eventually got it working, see https://github.com/mikelodder7/elliptic-curves/pull/2. But unfortunately this broke a bunch of tests, no surprise: I'm not a cryptographer and I don't know what I'm doing.

I also added a test with a single test vector from the OPRF specification just to show how and where it breaks. I'm happy to add a full OPRF test vector suite as in https://github.com/RustCrypto/elliptic-curves/pull/503, https://github.com/RustCrypto/elliptic-curves/pull/600 and https://github.com/RustCrypto/elliptic-curves/pull/964 when it passes.

ed448/src/field/scalar.rs
```diff
@@ -0,0 +1,1106 @@
...
+        }
+
+        // XXX :We are doing LSB first
+        bits
+    }
+
+    /// Construct a `Scalar` from a little-endian byte representation.
+    pub fn from_bytes(bytes: &[u8; 56]) -> Scalar {
+        Self(U448::from_le_slice(bytes))
+    }
```
daxpedda, line 747: This can produce invalid `Scalar`s. Just missing modulo `ORDER`.

ed448/src/field/scalar.rs
```diff
@@ -0,0 +1,1106 @@
...
+    where
+        X: for<'a> ExpandMsg<'a>,
+    {
+        let mut random_bytes = Array::<u8, U84>::default();
+        let dst = [dst];
+        let mut expander =
+            X::expand_message(&[msg], &dst, random_bytes.len()).expect("invalid dst");
+        expander.fill_bytes(&mut random_bytes);
+        Self::from_okm(&random_bytes)
+    }
```
daxpedda, line 902: The [specification says](https://www.rfc-editor.org/rfc/rfc9496.html#name-scalar-field-2) to use 64 bytes here. An external method taking 64 bytes, like `Scalar::from_bytes_mod_order_wide()`, would probably be useful as well.

ed448/src/constants.rs
```diff
@@ -0,0 +1,15 @@
+use crate::*;
+use crate::{decaf::DecafPoint, Scalar};
+
+pub const DECAF_BASEPOINT: DecafPoint = DecafPoint(curve::twedwards::extended::ExtendedPoint {
+    X: TWISTED_EDWARDS_BASE_POINT.X,
+    Y: TWISTED_EDWARDS_BASE_POINT.Y,
+    Z: TWISTED_EDWARDS_BASE_POINT.Z,
+    T: TWISTED_EDWARDS_BASE_POINT.T,
+});
```
daxpedda, line 9: While I have no idea why the twisted curve is used for the internal representation, at least the public `DecafPoint::GENERATOR` has to be correct.

```suggestion
pub const DECAF_BASEPOINT: DecafPoint = DecafPoint(curve::twedwards::extended::ExtendedPoint {
    X: FieldElement(ConstMontyType::new(&U448::from_be_hex("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa955555555555555555555555555555555555555555555555555555555"))),
    Y: FieldElement(ConstMontyType::new(&U448::from_be_hex("ae05e9634ad7048db359d6205086c2b0036ed7a035884dd7b7e36d728ad8c4b80d6565833a2a3098bbbcb2bed1cda06bdaeafbcdea9386ed"))),
    Z: FieldElement(ConstMontyType::new(&U448::from_u64(1))),
    T: FieldElement(ConstMontyType::new(&U448::from_be_hex("696d84643374bace9d70983a12aa9d461da74d2d5c35e8d97ba72c3aba4450a5d29274229bd22c1d5e3a6474ee4ffb0e7a9e200a28eee402"))),
});
```
I got the definition of the base point [from the specification](https://www.rfc-editor.org/rfc/rfc9496.html#name-scalar-field:~:text=66666666%2066666666%2066666666%2066666666%2066666666%2066666666%2066666666%0A33333333%2033333333%2033333333%2033333333%2033333333%2033333333%2033333333).

This is covered in https://github.com/mikelodder7/elliptic-curves/pull/2#discussion_r2038550669 as well.

ed448/src/field/scalar.rs
```diff
@@ -0,0 +1,1106 @@
...
+        let tmp = self.to_bytes_rfc_8032();
+        for &b in tmp.iter() {
+            write!(f, "{:02X}", b)?;
+        }
+        Ok(())
+    }
+}
+
+impl FromOkm for Scalar {
+    type Length = U84;
```
daxpedda, line 420: This should be 64 bytes as well, see https://github.com/RustCrypto/elliptic-curves/pull/1121#discussion_r2038632770.

ed448/src/field/scalar.rs
daxpedda: My assessment was incorrect here, the specification part I linked here is not related to the `hash_to_field` implementation.

ed448/src/field/scalar.rs
daxpedda: As pointed out in https://github.com/RustCrypto/elliptic-curves/pull/1121#discussion_r2038632770, this is wrong.

tarcieri:
> I ended up converting some of it to primeorder

That's a little weird since ed(wards)448 isn't a prime order curve

tarcieri:
Merged in #1219

PR #1046, opened 2024-05-10
Add ECDH + PKCS8 to bign256

makavity:
Hello!
* Added pkcs8 (working, tested with https://github.com/bcrypto/bee2evp)
* Added ECDH
* Fixed signature
* Removed sec1 partially

And also, don't know why tests is failed..

PR #1125, opened 2025-02-26
Upgrade all crates to the 2024 edition and bump MSRV to 1.85

baloo:


k256/src/ecdsa.rs
```diff
@@ -28,10 +28,10 @@
 //!     ecdsa::{SigningKey, Signature, signature::Signer},
 //!     SecretKey,
 //! };
-//! use rand_core::OsRng; // requires 'getrandom' feature
+//! use rand_core::{OsRng, TryRngCore}; // requires 'os_rng' feature
 //!
 //! // Signing
-//! let signing_key = SigningKey::random(&mut OsRng); // Serialize with `::to_bytes()`
+//! let signing_key = SigningKey::random(&mut OsRng.unwrap_mut()); // Serialize with `::to_bytes()`
```
baloo, line 34: We'll need https://github.com/RustCrypto/signatures/pull/915 to do better than that here.

bign256/README.md
```diff
@@ -33,17 +33,9 @@ USE AT YOUR OWN RISK!
...
-
-Rust **1.81** or higher.
-
-Minimum supported Rust version can be changed in the future, but it will be
-done with a minor version bump.
-
 ## SemVer Policy
 
 - All on-by-default features of this library are covered by SemVer
-- MSRV is considered exempt from SemVer as noted above
```
newpavlov, line 46: Delete this section completely similarly to the other repos?

.github/workflows/sm2.yml
```diff
@@ -19,44 +19,22 @@ env:
...
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: ${{ matrix.rust }}
-          targets: ${{ matrix.target }}
-      - run: cargo build --target ${{ matrix.target }} --release --no-default-features
-      - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features alloc
-      - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features arithmetic
-      - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features pkcs8
-      - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features alloc,arithmetic,pkcs8
```
tarcieri, line 42: Why did you remove this? It's how we check for `no_std` compatibility (perhaps it should be renamed to `no_std`)

bign256/src/lib.rs
```diff
@@ -132,7 +132,7 @@ impl FieldBytesEncoding<BignP256> for U256 {
 #[cfg(feature = "arithmetic")]
 pub type NonZeroScalar = elliptic_curve::NonZeroScalar<BignP256>;
 
-/// BIGN P-256 public key.
+// /// BIGN P-256 public key.
 // #[cfg(feature = "arithmetic")]
 // pub type PublicKey = elliptic_curve::PublicKey<BignP256>;
```
tarcieri, line 137: Sidebar: hmm, why's this commented out

baloo: I have no idea, that got added in #1046
The comment on the documentation was added by rustfmt.

.github/workflows/sm2.yml
baloo: getrandom now requires the backend to be specified. It won't work with either thumbv7 nor wasm32 without the consumer of the crate specifying it.

For wasm32 I could make a mock consumer-crate for this purpose and inject the `wasm_js` feature flag, but for thumbv7 I don't have a getrandom backend in the first place.
I can't pull `wasm_js` either directly.

baloo: Actually? `getrandom` shouldn't be pulled in the first place? There is something going on in that dependency tree.

baloo: That last commit should fix that. (now requires an external Rng)

tarcieri (approved):

PR #1146, opened 2025-04-21
p256: use of `DigestSigner` with a shorter hash

baloo:
This showcases the use of `DigestSigner` with a shorter hash than what p256 is usually used with.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -741,6 +741,7 @@ dependencies = [
  "proptest",
  "rand_core 0.9.3",
  "serdect",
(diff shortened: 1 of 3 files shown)


tarcieri:
I’d probably prefer not to add a `sha1` feature. Otherwise yes everything should be built to support different hash sizes

baloo:
well this is not a feature, just a dev-dependency.
Yeah I was convinced the hash output size had to line up with the field bytes size for ever. And I just wrote a test to check.

Not a big deal.

PR #1157, opened 2025-04-30
primefield+p384: upgrade to `fiat-crypto` v0.1.4

tarcieri:
Updates the field element type generating macros in the `primefield` crate to be compatible with the latest `fiat-crypto` code as postprocessed by `fiat-constify`, and regenerates the `p384` field impls using `fiat-crypto` v0.1.4.

This update leverages `const_mut_refs` to be much closer to the upstream `fiat-crypto` output, while still providing `const fn` support.

This change also starts extracting a `fiat` module within `primeorder` for macros that are specific to `fiat-crypto` output.

With this upgrade in place, upgrading the rest of the curve crates should be comparatively straightforward.


tarcieri:
This saw similar speedups in scalar multiplication and inversion as were seen in `p521` in #1156:

```
point operations/point-scalar mul
                        time:   [432.29 µs 434.73 µs 437.29 µs]
                        change: [-3.0895% -2.4875% -1.9227%] (p = 0.00 < 0.05)
                        Performance has improved.
scalar operations/invert
                        time:   [29.923 µs 29.973 µs 30.035 µs]
                        change: [-3.6955% -3.4156% -3.1371%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Other operations were not impacted.

primefield/src/fiat.rs
```diff
@@ -0,0 +1,239 @@
...
+
+            /// Add elements.
+            #[inline]
+            pub const fn add(&self, rhs: &Self) -> Self {
+                let mut out = $mont_type([0; <$uint>::LIMBS]);
+                $add(
+                    &mut out,
+                    &$mont_type(self.0.to_words()),
+                    &$mont_type(rhs.0.to_words()),
+                );
```
tarcieri, line 61: The newtypes introduced in newer versions of `fiat-crypto` make these conversions somewhat annoying, and they're occurring wherever we invoke any generated code.

Fortunately sprinkling `#[inline]` on it seems to be sufficient to get the compiler to optimize them away, but it might make sense to support casting between generated field element types (which contain `crypto-bigint::Uint`s) and references to their `fiat-crypto` generated newtypes.

The main thing that's missing right now is the `fiat-crypto` newtypes aren't annotated as `#[repr(transparent)]`, as what's really needed here is a cast between the inner array type and the newtype. That shouldn't be too difficult to add in `fiat-constify` though.

PR #1184, opened 2025-05-08
bp256+bp384: update README.md links

dishmaker:


tarcieri (approved):
Hmm, that uses up a lot of horizontal space, but maybe we can get rid of the arithmetic column now that it's redundant

dishmaker:
Also remove

> as noted in the arithmetic column below:

PR #1187, opened 2025-05-12
chore(deps): bump `crypto-bigint` to `0.7.0-pre.3`

baloo:
depends:
 - https://github.com/RustCrypto/traits/pull/1845

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -321,9 +321,9 @@ checksum = "43da5946c66ffcc7745f48db692ffbb10a83bfe0afd96235c5c2a4fb23994929"
 
 [[package]]
 name = "crypto-bigint"
-version = "0.7.0-pre.2"
+version = "0.7.0-pre.3"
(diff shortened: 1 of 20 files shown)


baloo:
This still throws warnings like:
```
warning: use of deprecated method `elliptic_curve::crypto_bigint::limb::mul::<impl elliptic_curve::crypto_bigint::Limb>::mac`: please use `carrying_mul_add` instead
   --> p256/src/arithmetic/field/field64.rs:125:26
    |
125 |     let (a1, carry) = a1.mac(a0, modulus[1], a0);
    |                          ^^^
    |
    = note: `#[warn(deprecated)]` on by default
```

but sadly, following the suggestion breaks the tests. I'm not sure I understand why.

fjarri:
The argument order changed. Sorry, should have included that in the deprecation note.

baloo:
I missed that! thanks a lot!

tarcieri (approved):

PR #1193, opened 2025-05-17
Implement `TryFrom<Scalar/Projective/AffinePoint>` for `NonIdentity/ZeroScalar`

daxpedda:
This implement `TryFrom<Scalar> for NonZeroScalar`, `TryFrom<ProjectivePoint> for NonIdentity<ProjectivePoint>` and `TryFrom<AffinePoint> for NonIdentity<AffinePoint>` in the same spirit as #1188 and #1190.

### DIFF
--- a/bign256/src/arithmetic/scalar.rs
+++ b/bign256/src/arithmetic/scalar.rs
@@ -223,6 +223,7 @@ impl From<&SecretKey> for Scalar {
     }
 }
 
+/// The constant-time alternative is available at [`NonZeroScalar::new()`].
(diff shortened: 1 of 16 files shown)


tarcieri:
These should probably carry a warning that they're not constant-time and at least document the constant-time alternative

daxpedda:
But the comparison itself is constant-time AFAIU. Handling the output isn't, but that seems communicated by the type to me as its not returning a `CtOption`.

Let me know if I'm making any sense here, otherwise I'm happy to add the warning.

tarcieri (approved):

PR #1205, opened 2025-05-28
Add `BatchInvert` tests

daxpedda:
Shows that https://github.com/RustCrypto/traits/pull/1864 is working correctly.

Updates `elliptic-curve` to include https://github.com/RustCrypto/traits/pull/1864.

### DIFF
--- a/p256/src/arithmetic/scalar.rs
+++ b/p256/src/arithmetic/scalar.rs
@@ -717,7 +717,8 @@ mod tests {
     use elliptic_curve::array::Array;
     use elliptic_curve::group::ff::{Field, PrimeField};
     use elliptic_curve::ops::{BatchInvert, ReduceNonZero};
-    use rand_core::OsRng;
+    use proptest::prelude::any;
+    use proptest::{prop_compose, proptest};
 
     /// t = (modulus - 1) >> S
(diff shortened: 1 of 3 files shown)


p256/src/arithmetic/scalar.rs
```diff
@@ -813,4 +814,16 @@ mod tests {
...
+    fn batch_invert() {
+        let scalars: [Scalar; 10] =
+            core::array::from_fn(|_| NonZeroScalar::try_from_rng(&mut OsRng).unwrap().into());
+
+        let inverted_scalars = Scalar::batch_invert(scalars.as_slice()).unwrap();
+
+        for (scalar, inverted_scalar) in scalars.into_iter().zip(inverted_scalars) {
+            assert_eq!(inverted_scalar, scalar.invert().unwrap());
+        }
+    }
```
tarcieri, line 828: I guess these are fine for now. It'd be good to get them in `primefield::test_field_invert` eventually but the `OsRng` dependency is a complication. It would be good if it used `proptest` instead for randomized testing, but similarly `primefield` isn't set up for that.

Maybe add some TODO comments to that effect? (move to `primefield::test_field_invert`, switch to `proptest`)

PR #1216, opened 2025-05-30
Add re-export p256::ecdsa::RecoveryId

### LINKED ISSUE #1215: Missing export p256::ecdsa::RecoveryId
There is `k256::ecdsa::RecoveryId` but no similar `p256::ecdsa::RecoveryId` exported. Is this intentional?

As `RecoveryId` is a parameter type of a public function `p256::ecdsa::VerifyingKey::recover_from_digest`, users need to create an explicit `ecdsa` dependency or use `k256::ecdsa::RecoveryId` for p256. Both are not ideal I guess.

Should this be added just like in k256?

webmaster128:
Closes #1215

### DIFF
--- a/p256/src/ecdsa.rs
+++ b/p256/src/ecdsa.rs
@@ -39,7 +39,10 @@
 //! # }
 //! ```
 
-pub use ecdsa_core::signature::{self, Error};
+pub use ecdsa_core::{
+    RecoveryId,
+    signature::{self, Error},
+};
 
 use super::NistP256;
 use ecdsa_core::EcdsaCurve;


p256/src/ecdsa.rs
```diff
@@ -39,10 +39,12 @@
 //! # }
 //! ```
 
-pub use ecdsa_core::signature::{self, Error};
+pub use ecdsa_core::{
+    EcdsaCurve, RecoveryId,
+    signature::{self, Error},
+};
```
webmaster128, line 45: This also makes `EcdsaCurve` public as I copied the block from k256. Is this a good thing? Or should I remove the `EcdsaCurve` change?

tarcieri (approved):

PR #1217, opened 2025-05-31
chore(deps): bump `crypto-bigint` to `0.7.0-pre.4`

baloo:


baloo:
I wish I caught that earlier.

tarcieri:
Why isn't there a Cargo.lock update?

tarcieri:
I'll probably just need to cut new releases of `primefield` and all of the elliptic curve crates

baloo:
> Why isn't there a Cargo.lock update?

because I forgot to run `cargo`, I've used it as a git dependency directly.

PR #1219, opened 2025-06-01
ed448-goldilocks: initial import

baloo:
Imports source code from: https://github.com/crate-crypto/Ed448-Goldilocks

Another take on #1208 that doesn't pull the macros from primeorder.


ed448/src/curve/edwards/extended.rs
```diff
@@ -755,7 +756,7 @@ impl EdwardsPoint {
     /// see <https://datatracker.ietf.org/doc/rfc9380/>
     pub fn hash<X>(msg: &[u8], dst: &[u8]) -> Self
     where
-        X: for<'a> ExpandMsg<'a>,
+        X: ExpandMsg<U32>,
```
baloo: cc @daxpedda 

I would have expected to use U84, but I'm not sure I understand why that doesn't work.

daxpedda: The generic here should be the [variable `k`](https://www.rfc-editor.org/rfc/rfc9380.html#section-8-5.6), which for [Curve/Ed448 is 224 bits](https://www.rfc-editor.org/rfc/rfc9380.html#section-8.6-3.5). So this should be `U28`.

ed448/src/macros.rs
```diff
@@ -130,3 +130,60 @@ macro_rules! define_mul_assign_variants {
         }
     };
 }
+
+macro_rules! scalar_from_impls {
```
baloo: @tarcieri I ended up duplicating that from primeorder. I don't know if it's worth the effort moving that to `elliptic-curve`

ed448/src/curve/edwards/extended.rs
baloo: hence the `CollisionResistance` relationship

ed448/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+description = """A pure-Rust implementation of Ed448 and Curve448 and Decaf.
+This crate also includes signing and verifying of Ed448 signatures.
+"""
+documentation = "https://docs.rs/ed448-goldilocks"
+exclude = [".gitignore", ".github/*"]
+edition = "2024"
+homepage = "https://docs.rs/ed448-goldilocks/"
+keywords = ["cryptography", "decaf", "ed448", "ed448-goldilocks"]
+license = "BSD-3-Clause"
+name = "ed448"
```
tarcieri: Would suggest the crate name remain [`ed448-goldilocks`](https://crates.io/crates/ed448-goldilocks) so `ed448` can be [used for the signature algorithm](https://github.com/RustCrypto/signatures/tree/master/ed448).

FWIW, RFC8032 calls the signature algorithm "Ed448" and the curve "edwards448"

tarcieri: Also, the `name` field should probably be at the top, and the other fields should follow the standard order: https://github.com/RustCrypto/meta/issues/22

ed448/src/macros.rs
tarcieri: That's fine

tarcieri:
@baloo can you also rename the containing directory and CI workflow to `ed448-goldilocks`?

(FWIW, it's the original published name of this curve: https://eprint.iacr.org/2015/625.pdf)

Also looks like you need a `rustfmt`

tarcieri (approved):

tarcieri:
@kevaundray can you add me as an owner of the `ed448-goldilocks` crate?

baloo:
(and likely https://crates.io/crates/ed448 to push https://github.com/RustCrypto/signatures/tree/master/ed448)

tarcieri:
@baloo yeah, already asked here: https://github.com/RustCrypto/signatures/pull/990#issuecomment-2946019178

dignifiedquire:
what would be the best place to PR in x448?

kevaundray:
> @kevaundray can you add me as an owner of the `ed448-goldilocks` crate?

I just added your username, can you confirm you received it?

tarcieri:
@kevaundray got them both, thanks!

Not sure about `x448`. We can either put it under `elliptic-curves` or possibly KEMs, though it isn't really a KEM.

@kevaundray can you add me as an owner of `x448` too?

kevaundray:
> @kevaundray got them both, thanks!
> 
> Not sure about `x448`. We can either put it under `elliptic-curves` or possibly KEMs, though it isn't really a KEM.
> 
> @kevaundray can you add me as an owner of `x448` too?

added you to that too :)

PR #1224, opened 2025-06-03
ed448-goldilocks: reuse types from `ed448`

baloo:
Depends:
 - https://github.com/RustCrypto/elliptic-curves/pull/1219
 - https://github.com/RustCrypto/signatures/pull/983
 

This is a followup to #1219

### DIFF
--- a/ed448-goldilocks/Cargo.toml
+++ b/ed448-goldilocks/Cargo.toml
@@ -18,6 +18,7 @@ This crate also includes signing and verifying of Ed448 signatures.
 [dependencies]
 crypto_signature = { version = "3.0.0-rc.0", default-features = false, features = ["digest", "rand_core"], optional = true, package = "signature" }
 elliptic-curve = { version = "0.14.0-rc.5", features = ["arithmetic", "hash2curve", "pkcs8"] }
+ed448 = { version = "=0.5.0-pre.0", default-features = false, optional = true }
 rand_core = { version = "0.9", default-features = false }
 serdect = { version = "0.3.0", optional = true }
 sha3 = { version = "0.11.0-rc.0", default-features = false }
@@ -25,13 +26,12 @@ subtle = { version = "2.6", default-features = false }
 
 [features]
 default = ["std", "signing", "pkcs8"]
-alloc = ["crypto_signature/alloc", "elliptic-curve/alloc", "serdect/alloc"]
+alloc = ["crypto_signature/alloc", "ed448?/alloc", "elliptic-curve/alloc", "serdect/alloc"]
 std = ["alloc"]
-
 bits = ["elliptic-curve/bits"]
-pkcs8 = ["elliptic-curve/pkcs8"]
-signing = ["dep:crypto_signature"]
-serde = ["dep:serdect"]
+pkcs8 = ["ed448?/pkcs8", "elliptic-curve/pkcs8"]
+signing = ["dep:crypto_signature", "ed448"]
+serde = ["dep:serdect", "ed448?/serde_bytes"]
 
 [dev-dependencies]
 hex-literal = "1"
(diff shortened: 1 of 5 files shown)


tarcieri:
@baloo this looks like it needs to be updated to include RustCrypto/signatures#985

tarcieri:
GitHub seems to be having a bad time. WTF is this?

> bign256 / cross ... Cancelled after 1064745725m

Edit: githubstatus.com confirms they were having problems and just pushed out a fix. Gonna try kicking the failed jobs.

Cargo.toml
```diff
@@ -18,7 +18,8 @@ members = [
 
 [profile.dev]
 opt-level = 2
-
 [patch.crates-io]
 primefield = { path = "primefield" }
 primeorder = { path = "primeorder" }
+
+ed448-signature = { git = "https://github.com/RustCrypto/signatures.git" }
```
tarcieri: Hmm, can we rename this crate to `ed448` first (to match the `ed25519` crate, and because the proper name of this EdDSA variant is "Ed448")

tarcieri: Note: opened https://github.com/RustCrypto/signatures/pull/990

ed448-goldilocks/Cargo.toml
```diff
@@ -19,6 +19,7 @@ This crate also includes signing and verifying of Ed448 signatures.
 crypto-bigint = { version = "=0.7.0-pre.4", features = ["hybrid-array"], default-features = false }
 crypto_signature = { version = "3.0.0-rc.0", default-features = false, features = ["digest", "rand_core"], optional = true, package = "signature" }
 elliptic-curve = { version = "0.14.0-rc.3", features = ["arithmetic", "bits", "hash2curve", "jwk", "pkcs8", "pem", "sec1"] }
+ed448 = { version = "0.0.0", default-features = false, optional = true }
```
tarcieri: I can cut an RC for this soon, but this is out:

```suggestion
ed448 = { version = "=0.5.0-pre.0", default-features = false, optional = true }
```

ed448-goldilocks/src/curve/edwards/extended.rs
```diff
@@ -211,6 +193,13 @@ impl<'de> serdect::serde::Deserialize<'de> for CompressedEdwardsY {
     }
 }
 
+#[cfg(feature = "ed448")]
+impl From<ed448::ComponentBytes> for CompressedEdwardsY {
+    fn from(point: ed448::ComponentBytes) -> Self {
+        Self(point)
+    }
+}
```
tarcieri:
```suggestion
impl From<PointBytes> for CompressedEdwardsY {
    fn from(point: PointBytes) -> Self {
        Self(point)
    }
}
```

tarcieri (approved):

tarcieri:
Oh fun, rustfmt

baloo:
my editor runs rustfmt on save, and for a reason I don't understand, for ed448-goldilocks it insists on having it formatted using edition 2021

EDIT: because there was a `rustfmt.toml` in the subdirectory that I've missed.

tarcieri:
If your editor is IntelliJ, I have the same problem

baloo:
no I use neovim and rust.vim, but it doesn't read the edition from the `Cargo.toml` and just runs `rustfmt` which is only configured using a `rustfmt.toml`.

I place a small `rustfmt.toml` in the root of the workspaces once they moved to 2024 edition:
```
edition = "2024"
```

and that has been working for me so far. I'd prefer having rust.vim or rustfmt respect the edition from `Cargo.toml` but I have absolutely no talent when it comes to writing vim plugins.

tarcieri (approved):

PR #1237, opened 2025-06-06
chore(deps): bump `der` and `ecdsa` to `v0.8.0-rc.4` and `v0.17.0-rc.1`

baloo:
supersedes #1235

Both need to be updated at the same time because of API breaks


tarcieri:
Didn't notice this until it's too late, but #1240 achieved the same effect

PR #1242, opened 2025-06-07
ed448-goldilocks: use `cargo hack test`

tarcieri:
Test all feature permutations


tarcieri:
Really we should do this on all the crates in this repo, but `ed448-goldilocks` is the one in most need of feature refactoring

PR #1248, opened 2025-06-08
Update to changes in `BatchNormalize`

daxpedda:
This updates to changes made in `BatchNormalize` in https://github.com/RustCrypto/traits/pull/1896. I also added a test for `NonIdentity::batch_normalize` for `p256`.

I looked into moving this test into `test_projective_arithmetic`, but I don't see a good way to currently do this because of how we can't get that array type for `prop_compose`. Alternatively we could require the corresponding types to implement [`Strategy`](https://docs.rs/proptest/latest/proptest/strategy/trait.Strategy.html). `proptest` also just updated to `rand` v0.9 so I think this would be a good idea. WDYT?

Depends on https://github.com/RustCrypto/traits/pull/1896.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -403,8 +403,7 @@ checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 [[package]]
 name = "elliptic-curve"
 version = "0.14.0-rc.5"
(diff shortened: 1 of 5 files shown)


daxpedda:
Closing in favor of https://github.com/RustCrypto/elliptic-curves/pull/1257.

PR #1250, opened 2025-06-09
x448: initial import

baloo:
This imports https://crates.io/crates/x448 in https://github.com/RustCrypto/elliptic-curves/


baloo:
This is currently based on https://github.com/RustCrypto/elliptic-curves/pull/1224 (mostly for sanity and avoid merge conflicts), I'll rebase once that merges.

baloo:
cc @dignifiedquire @kevaundray

x448/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "x448"
+version = "0.7.0-pre"
+authors = ["RustCrypto Developers"]
+categories = ["cryptography"]
+keywords = ["cryptography", "crypto", "x448", "diffie-hellman", "curve448",]
+homepage = "https://docs.rs/x448/"
+repository = "https://github.com/RustCrypto/elliptic-curves/tree/master/x448"
+documentation = "https://docs.rs/ed448-goldilocks"
+license = "BSD-3-Clause"
```
tarcieri: Can we get this changed to Apache 2.0 + MIT (with permission) before importing?

```suggestion
license = "Apache-2.0 OR MIT"
```

tarcieri: Requesting permission from the three contributors to relicense the `x448` crate as Apache 2.0 + MIT:

- [x] @kevaundray
- [x] @anonion0
- ~~@StormWindStudios~~

anonion0: I grant permission to relicense my contributions as Apache 2.0+MIT.

baloo: I don't think @StormWindStudios's contributions are making it in this import

We'll just need @kevaundray permission to relicense.

baloo:
This kind of broke when we merged https://github.com/RustCrypto/elliptic-curves/pull/1284

cc @daxpedda

I'm confused. I don't understand how https://www.rfc-editor.org/rfc/rfc7748#section-5 asks for the scalar to be a 56-bytes string and now the `EdwardsScalar` is 57 bytes.

Am I supposed to use the `DecafScalar` instead?

daxpedda:
Apologies, I noticed indeed that this should have been `DecafScalar`.
I'm currently preparing another PR where I try to fix those names, because having a separate scalar for each curve when only one is special, Ed448, is a bit much.

Feel free to fix this in this PR.
Basically just replace all uses of `EdwardsScalar` in `montgomery.rs` with `DecafScalar`.

baloo:
No worries, I'm not sure I'm following what you're saying, I guess I'll wait for your PR ^^

daxpedda:
I have a new and improved PR up for `MontgomeryScalar`: #1306.

x448/Cargo.toml
kevaundray: Hi sorry for late reply -- I grant permission to relicense my contribution as Apache 2.0 + MIT

kevaundray: Do you think you could add a copyright 20xx -2025 ? Seems I forgot to add it in the original repo

baloo:
This is rebased on:
 - #1306,
 - #1305, and
 - #1304

tarcieri:
@baloo is there a specific reason this is still draft? It'd be good to get it landed so we make sure it gets updated when `ed448-goldilocks` is changed

daxpedda:
Its currently relying on https://github.com/RustCrypto/elliptic-curves/pull/1306.
I could try to extract `MontgomeryScalar` from #1306 to make it easily reviewable first if that's desirable.

baloo:
> @baloo is there a specific reason this is still draft? It'd be good to get it landed so we make sure it gets updated when `ed448-goldilocks` is changed

I struggled for a bit with various `Scalar`s, #1306 fixed it for me.

I'll keep rebasing and testing along with the merges until we can get this out of draft :)

baloo:
I guess the scalar wasn't that hard to figure out in the end.

I've sent a PR over to @daxpedda 's branch to migrate over to `MontgomeryXpoint` https://github.com/khonsulabs/elliptic-curves/pull/1

baloo:
Is there anything else required here?

x448/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+            return None;
+        }
+        let shared_key = &public_key.0 * &self.as_scalar();
+        Some(SharedSecret(shared_key))
+    }
+
+    /// Performs a Diffie-hellman key exchange once between the secret key and an external public key
+    pub fn to_diffie_hellman(self, public_key: &PublicKey) -> Option<SharedSecret> {
+        self.as_diffie_hellman(public_key)
+    }
```
tarcieri, line 128: This doesn't seem like a helpful distinction. I think we should just have the borrowing version. Even for ephemeral D-H keys, there are valid reasons to perform more than one D-H operation (i.e. 3DH)

x448/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+
+    /// Converts a byte slice into a secret and clamp
+    pub fn from_bytes(bytes: &[u8]) -> Option<Secret> {
+        // First check if we have 56 bytes
+        if bytes.len() != 56 {
+            return None;
+        }
+
+        let secret = Secret::from(slice_to_array(bytes));
+        Some(secret)
```
tarcieri, line 138: Seems like this could use `TryFrom`

x448/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    /// Converts a secret into a byte array
+    pub fn as_bytes(&self) -> &[u8; 56] {
+        &self.0
+    }
```
tarcieri, line 144: We should probably have different types for ephemeral vs static secrets ala `x25519-dalek`. For ephemeral secrets ideally the caller couldn't export this value (for forward secrecy).

PR #1256, opened 2025-06-13
Update to hash2curve changes

daxpedda:
Companion PR to https://github.com/RustCrypto/traits/pull/1901.

PR #1257, opened 2025-06-13
Update `elliptic-curve`

daxpedda:
Updates `elliptic-curve` to master.
This merges https://github.com/RustCrypto/elliptic-curves/pull/1248 and https://github.com/RustCrypto/elliptic-curves/pull/1256.

### DIFF
--- a/primeorder/src/projective.rs
+++ b/primeorder/src/projective.rs
@@ -13,7 +13,7 @@ use elliptic_curve::{
     array::ArraySize,
     bigint::ArrayEncoding,
     group::{
-        self, Group, GroupEncoding,
+        Curve, Group, GroupEncoding,
         prime::{PrimeCurve, PrimeGroup},
     },
     ops::{BatchInvert, LinearCombination},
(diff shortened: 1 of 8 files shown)


primeorder/src/projective.rs
```diff
@@ -308,7 +308,7 @@ where
     }
 }
 
-impl<C> group::Curve for ProjectivePoint<C>
+impl<C> Curve for ProjectivePoint<C>
```
tarcieri, line 311: I used `group::Curve` intentionally to keep it namespaced since `Curve` is incredibly overloaded (really it should probably be named `CurveGroup`, see zkcrypto/group#51).

I guess `elliptic-curve` could re-export it as `CurveGroup`.

All that said, I guess this is fine for now and I'll follow up.

tarcieri (approved):

primeorder/src/projective.rs
daxpedda: Ah, apologies. I was importing it to be able to use it. But I can just import it as `use ...::Curve as _`.

Let me know if you want me to fix this up.

tarcieri: Can fix it up after this: https://github.com/RustCrypto/traits/pull/1902

PR #1269, opened 2025-06-19
Update SIGMA protocol reference link in documentation

VolodymyrBg:
Replaced the outdated SIGMA protocol reference URL in the ECDH module documentation with the current official Springer link to the original SIGMA paper by Hugo Krawczyk. This ensures the documentation points to a reliable and accessible source for the protocol details.

### DIFF
--- a/bign256/src/ecdh.rs
+++ b/bign256/src/ecdh.rs
@@ -20,7 +20,7 @@
 //! [`diffie_hellman`] function.
 //!
 //! [AKE]: https://en.wikipedia.org/wiki/Authenticated_Key_Exchange
-//! [SIGMA]: https://link.springer.com/chapter/10.1007/978-3-540-45146-4_24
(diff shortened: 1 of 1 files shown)


tarcieri:
Why didn't you use the same IACR reference as https://github.com/RustCrypto/traits/pull/1910?

VolodymyrBg:
> Why didn't you use the same IACR reference as [RustCrypto/traits#1910](https://github.com/RustCrypto/traits/pull/1910)?

I use google search to find correct links, i didn't remember about that link, so I chose the first one correct link.

tarcieri:
I would suggest using the same link as the first PR. IACR is the most popular source for cryptography papers.

VolodymyrBg:
> I would suggest using the same link as the first PR. IACR is the most popular source for cryptography papers.

Done, sorry for this inconsistency

PR #1270, opened 2025-06-19
primeorder: add optimized addition formulas for curves with a=0

andrewwhitehead:
These formulas are from the same source as the existing implementations and apply to curves such as secp256k1.


tarcieri (approved):
utACK having compared the code to Algorithms 7, 8, and 9 in the paper by hand and I wasn't able to find any transcription errors.

A real test would be great but all I can think to do is try to incorporate it into `k256`, where the laziness of the field elements makes them incompatible with `primeorder`.

PR #1275, opened 2025-06-28
k256: Remove implementation of `PrehashSignature`

carloskiki:
As discussed in https://github.com/RustCrypto/traits/issues/1917.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,5 +20,6 @@ members = [
 opt-level = 2
 
 [patch.crates-io]
+ecdsa = { git = "https://github.com/carloskiki/signatures.git" }
 primefield = { path = "primefield" }
 primeorder = { path = "primeorder" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,6 @@ members = [
 opt-level = 2
 
 [patch.crates-io]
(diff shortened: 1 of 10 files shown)


tarcieri:
@carloskiki maybe you can turn this into a full companion PR to https://github.com/RustCrypto/traits/issues/1917 which both removes `PrehashSignature` and tests the `ecdsa` changes against the concrete curve crates?

carloskiki:
I now pull the `ecdsa` dependency from https://github.com/RustCrypto/signatures/pull/1007. Do you want me to write tests or are we just checking if current tests pass & everything compiles? If you want me to write tests, what type of behavior should I be checking?

tarcieri:
@carloskiki no additional tests are needed, but can you use a toplevel patch.crates-io directive to check the whole repo?

Cargo.toml
```diff
@@ -20,5 +20,6 @@ members = [
 opt-level = 2
 
 [patch.crates-io]
+ecdsa = { git = "https://github.com/carloskiki/signatures.git" }
```
tarcieri:
```suggestion
ecdsa = { git = "https://github.com/RustCrypto/signatures.git" }
```

PR #1283, opened 2025-07-07
ed448-goldilocks: fix Decaf448 map2curve and add test vectors

### LINKED ISSUE #1282: `decaf448_map` is faulty
While there are unfortunately no official [`decaf448_map`](https://www.rfc-editor.org/rfc/rfc9380.html#appendix-C-2) test vectors, we can use the [OPRF](https://www.rfc-editor.org/rfc/rfc9497.html#name-decaf448-shake256) ones.
You can find an example of a failing test [here](https://github.com/khonsulabs/elliptic-curves/blob/6d0edf22315d35587d9f412ce2243d982dca6342/ed448-goldilocks/src/decaf.rs#L20-L68).

While comparing the output with other implementations ([OPRF PoC](https://github.com/cfrg/draft-irtf-cfrg-voprf), [crrl](https://github.com/pornin/crrl) and [noble-curves](https://github.com/paulmillr/noble-curves)), which all work correctly, I determined that the issue specifically happens in [`FieldElement::map_to_curve_decaf448()`](https://github.com/RustCrypto/elliptic-curves/blob/67756d7babcf7ca6bb9bd7afb633d4d1b50014c8/ed448-goldilocks/src/field/element.rs#L385).
This implements itself seems to match the [specification of `decaf448_map`](https://www.rfc-editor.org/rfc/rfc9496#section-5.3.4-8). I've compared the [output `X`, `Y`, `Z` and `T` coordinates](https://github.com/RustCrypto/elliptic-curves/blob/67756d7babcf7ca6bb9bd7afb633d4d1b50014c8/ed448-goldilocks/src/field/element.rs#L410-L415) with the other libraries and they all match. But apparently something goes wrong after [twisting the resulting `EdwardsPoint`](https://github.com/RustCrypto/elliptic-curves/blob/67756d7babcf7ca6bb9bd7afb633d4d1b50014c8/ed448-goldilocks/src/field/element.rs#L416), the resulting serialized `DecafPoint` doesn't match with the other libraries.

So either something is wrong with the [twisting implementation](https://github.com/RustCrypto/elliptic-curves/blob/67756d7babcf7ca6bb9bd7afb633d4d1b50014c8/ed448-goldilocks/src/curve/edwards/extended.rs#L696-L698) or the serialization, which I doubt.

I could not find a library that implements Decaf448 using a twisted representation. Let me know if there is one, then I'm sure I could figure out the difference.

daxpedda:
This PR adds the official hash2curve test vectors for Decaf448: https://www.rfc-editor.org/rfc/rfc9496#name-group-elements-from-uniform-.

~~As pointed out in #1282, this currently fails. I have confirmed that the same test vectors work with other libraries, e.g. [OPRF PoC](https://github.com/cfrg/draft-irtf-cfrg-voprf), [crrl](https://github.com/pornin/crrl) and [noble-curves](https://github.com/paulmillr/noble-curves).~~

I eventually figured out the correct implementation and fixed it.

Fixes #1282.

### DIFF
--- a/ed448-goldilocks/src/decaf.rs
+++ b/ed448-goldilocks/src/decaf.rs
@@ -9,3 +9,78 @@ mod scalar;
 pub use affine::AffinePoint;
 pub use points::{CompressedDecaf, DecafPoint};
 pub use scalar::{DecafScalar, DecafScalarBytes, WideDecafScalarBytes};
+
+// https://www.rfc-editor.org/rfc/rfc9496#name-group-elements-from-uniform-
+#[test]
+fn hash_to_curve() {
+    use hex_literal::hex;
+
+    const TEST_VECTORS: [TestVector; 7] = [
+        TestVector {
+            input: hex!(
+                "cbb8c991fd2f0b7e1913462d6463e4fd2ce4ccdd28274dc2ca1f4165d5ee6cdccea57be3416e166fd06718a31af45a2f8e987e301be59ae6673e963001dbbda80df47014a21a26d6c7eb4ebe0312aa6fffb8d1b26bc62ca40ed51f8057a635a02c2b8c83f48fa6a2d70f58a1185902c0"
+            ),
+            output: hex!(
+                "0c709c9607dbb01c94513358745b7c23953d03b33e39c7234e268d1d6e24f34014ccbc2216b965dd231d5327e591dc3c0e8844ccfd568848"
+            ),
+        },
+        TestVector {
+            input: hex!(
+                "b6d8da654b13c3101d6634a231569e6b85961c3f4b460a08ac4a5857069576b64428676584baa45b97701be6d0b0ba18ac28d443403b45699ea0fbd1164f5893d39ad8f29e48e399aec5902508ea95e33bc1e9e4620489d684eb5c26bc1ad1e09aba61fabc2cdfee0b6b6862ffc8e55a"
+            ),
+            output: hex!(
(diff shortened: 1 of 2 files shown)


tarcieri:
@daxpedda should we remove it for now?

daxpedda:
I think we should. While I am unable to judge if the current output has the security guarantees set by the specification or not, it is not compatible with other implementations and will cause protocols to not work correctly.

Will make a PR when I have time.

daxpedda:
After some more digging I stumbled upon the [Elligator 2 mapping in the Decaf paper](https://www.shiftleft.org/papers/decaf/decaf.pdf#section.A.3).

While I was unable to translate it directly, I found [a working implementation in the reference implementation](https://sourceforge.net/p/ed448goldilocks/code/ci/e5cc6240690d3ffdfcbdb1e4e851954b789cd5d9/tree/src/per_curve/elligator.tmpl.c#l28), which I didn't even know existed.

Potentially the implementation could be cleaned up to align more with the paper, but I couldn't really figure out the mathematical gymnastics going on in the reference implementation.

I never figured out why the current implementation didn't work though. The implementation is taken from [RFC9496](https://www.rfc-editor.org/rfc/rfc9496#name-element-derivation-2) which assumes an extended Edwards coordinate representation. Our current implementation did reflect this but still didn't work. I'm still assuming something is wrong with our `EdwardsPoint` to `TwistedExtendedPoint` conversion. In any case, the new implementation should be much more performant anyway.

TLDR: I fixed the mapping.

PR #1284, opened 2025-07-08
ed448-goldilocks: split `Scalar` into `EdwardsScalar`/`DecafScalar`

### LINKED ISSUE #1229: ed448-goldilocks: 56-byte decaf x-coordinates
Decaf x-coordinates are supposed to be serialized as 56-bytes, but we're serializing them as 57-bytes, with a TODO to fix this:

https://github.com/RustCrypto/elliptic-curves/blob/c797d60/ed448-goldilocks/src/decaf/affine.rs#L63-L68

This has turned into a blocker for this PR to expose `AffineCoordinates::y`, since that trait assumes both the x and y-coordinates have the same size:

https://github.com/RustCrypto/traits/pull/1891

I think for now I am just going to comment out the `AffineCoordinates` and `CurveArithmetic` trait impls for Decaf, until such a time as that TODO can actually be addressed.

daxpedda:
Because Ed448 and Decaf448 have different scalar representations, we need two separate types.

This PR achieves this by making `Scalar` generic and adding two new types:
- `type EdwardsScalar = Scalar<Ed448>`
- `type DecafScalar = Scalar<Decaf448>`

`Scalar` itself must still be publicly visible to allow its methods and implementations to show up in the documentation. A new private trait `CurveWithScalar` is introduced to let implementations differentiate between both curves. Because `Scalar<C: CurveWithScalar>` relies on this private trait `Scalar` can't be used by users with other curves.

Alternatively we could also achieve the same result with a huge macro instead of a shared generic `Scalar` type. Let me know if you prefer that instead.

Resolves #1229.

### DIFF
--- a/ed448-goldilocks/README.md
+++ b/ed448-goldilocks/README.md
@@ -17,25 +17,25 @@ It is intended to be portable, fast, and safe.
 ## Usage
 
 ```rust
-use ed448_goldilocks::{EdwardsPoint, CompressedEdwardsY, Scalar, sha3::Shake256};
+use ed448_goldilocks::{EdwardsPoint, CompressedEdwardsY, EdwardsScalar, sha3::Shake256};
 use elliptic_curve::Field;
 use hash2curve::ExpandMsgXof;
 use rand_core::OsRng;
 
-let secret_key = Scalar::TWO;
+let secret_key = EdwardsScalar::TWO;
 let public_key = EdwardsPoint::GENERATOR * &secret_key;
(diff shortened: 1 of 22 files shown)


daxpedda:
Just realized that the scalar representation isn't the only difference that necessitates this change. Decaf448 also requires a different `FromOkm` implementation for hash2curve.

I went ahead and fixed that as well here.

tarcieri (approved):

PR #1286, opened 2025-07-08
Add `hash2curve` crate

daxpedda:
This moves the hash2curve implementation from `elliptic-curve` to its own crate in this (`elliptic-curves`) repo as discussed in https://github.com/RustCrypto/traits/issues/1928.

I also temporarily moved the `OprfParameters` trait as well, as it can't live in `elliptic-curve` anymore because it depends on hash2curve traits, but the new `hash2curve` crate depends on `elliptic-curve`, creating a circular dependency.

I'm not sure what the plan is with OPRF, but I am currently working on upstreaming an OPRF implementation, where the trait can be moved to.

Companion PR: https://github.com/RustCrypto/traits/pull/1929.


hash2curve/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "hash2curve"
+version = "0.2.0-rc.0"
```
tarcieri: Perhaps this to keep the version number in sync with all of the other elliptic curve crates?

```suggestion
version = "0.14.0-rc.0"
```

daxpedda: Done.

PR #1287, opened 2025-07-08
ed448-goldilocks: impl `GroupDigest` for `Ed448` and `Decaf448`

daxpedda:
This implements `GroupDigest` for `Ed448` and `Decaf448` and removes the already existing standalone functions. Test vectors from the hash2curve specification were already in place.

I also added test vectors for `hash_to_scalar` from OPRF, like we already have in place for P-256, P-384 and P-521.

### DIFF
--- a/ed448-goldilocks/README.md
+++ b/ed448-goldilocks/README.md
@@ -17,9 +17,9 @@ It is intended to be portable, fast, and safe.
 ## Usage
 
 ```rust
-use ed448_goldilocks::{EdwardsPoint, CompressedEdwardsY, EdwardsScalar, sha3::Shake256};
+use ed448_goldilocks::{Ed448, EdwardsPoint, CompressedEdwardsY, EdwardsScalar, sha3::Shake256};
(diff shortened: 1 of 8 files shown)


daxpedda:
Apparently there is an issue in `hash2curve`: `hash_from_bytes()` isn't compatible with Decaf448's standard implementation. It requires a length of 112. But `hash_from_bytes()` calculates the length as `2 * FromOkm::Length`, which says 168 in this case.

Will have to think how to address this in `hash2curve`, or alternatively just overwrite the default implementation.

daxpedda:
I fixed it by using wrapper types around `FieldElement`: `Ed448FieldElement` and `Decaf448FieldElement`. That way we can implement `FromOkm` selectively depending on the curve. These types are only used in `MapToCurve`s `type FieldElement` and nowhere else.

I spent some time experimenting with a generic `FieldElement` which could work as well. But seeing that the workaround was so simple I let it be. Maybe we can get back to that if `FieldElement` needs to be differentiated further in the future.

PR #1291, opened 2025-07-11
Implement `Curve448` with full coordinates

daxpedda:
This PR implements a set of types for Montgomery points with a full coordinate system and corresponding `Curve448` type with a `CurveArithmetic` implementation. Our current `MontgomeryPoint` x-coordinate only remains in place for use with `X448` and is renamed to `MontgomeryXpoint`.

For context: the x-only coordinate system obviously has a lot of missing information to properly implement a full set of arithmetic operations. Crucially, while the y-coordinate could be recovered, it is costly and is missing the sign. Some protocols exist that set the sign of the y-coordinate, which is why we have some methods in place for `MontgomeryXpoint` to do exactly that.

I made sure to add a full set of conversion methods between the new types.


daxpedda:
I just stumbled on ["Faster Complete Addition Laws for MontgomeryCurves"](https://tches.iacr.org/index.php/TCHES/article/view/11808/11313) and will be looking into that.

daxpedda:
> I just stumbled on ["Faster Complete Addition Laws for MontgomeryCurves"](https://tches.iacr.org/index.php/TCHES/article/view/11808/11313) and will be looking into that.

I'm going to leave this to a future somebody to look into.
Will open an issue to track it after the PR is merged.

daxpedda:
I found a bug in the current implementation of Montgomery scalar multiplication.
Our differential addition and double algorithm assumed that `diff` is normalized (Z=1), which it wasn't in our case.
It wasn't caught by tests because we only did one multiplication, which always started with a normalized point.

It is possible to switch our differential addition and double algorithm to not assume Z=1, which would save us the initial invert operation (normalization). But it adds one extra multiplication, so one inversion at the start of our scalar multiplication is probably cheaper then 448 multiplications.

daxpedda:
Closing this in favor of splitting it into multiple smaller PRs.

Issue #1295, opened 2025-07-13
`hash2curve`: Bringing breaking changes to ease maintainability & usability.


carloskiki:
There are a few changes I would like to bring to `hash2curve`, mainly to reduce the amount of code the crate has and make it independent of the `elliptic-curve` crate.

## Motivations

1. Reducing the amount of code is always a plus IMHO - makes the crate more maintainable & approachable.
2. `elliptic-curve` is almost unused as a dependency, but brings a lot of transitive dependencies (see below for comparison).
3. Some functions use out parameters and I would like to change that where possible (`Expander` and `hash_to_field`).

## Changes

- Change `GroupDigest` to something like a `Suite` trait.
```rust
/// A hash to curve suite.
///
/// <https://www.rfc-editor.org/rfc/rfc9380.html#name-suites-for-hashing>
pub trait Suite {
    const ID: &'static str;

    type Point: MapToCurve;
    type SecurityLevel: Unsigned;
    type ExpandMsg: ExpandMsg<Self::SecurityLevel>;

    fn hash_from_bytes(msg: &[&[u8]], dst: &[&[u8]]) -> Option<Self::Point> {
        // ...
    }

    fn encode_from_bytes(msg: &[&[u8]], dst: &[&[u8]]) -> Option<Self::Point> {
        // ...
    }
}
```
- MapToCurve will be bound by `Group` instead of `CurveArithmetic`.
- All places that return `elliptic_curve::Result<_>` will return `Option` instead (`elliptic_curve::Error` was already as ZST).
- Make `hash_to_field` output `Array<F, C>` for `F: FromOkm` and `C: ArraySize` instead of taking `&mut [F]` as out parameter.
- Make `ExpandMsg` return `Iterator<Item = u8>` instead of an `Expander` (This may cause a slight performance regression, will be benchmarked before it is added).

## Dependency tree with & without `elliptic-curve`

With `elliptic-curve`

```
hash2curve v0.14.0-rc.0
├── digest v0.11.0-rc.0
│   ├── block-buffer v0.11.0-rc.4
│   │   └── hybrid-array v0.3.1
│   │       ├── typenum v1.18.0
│   │       └── zeroize v1.8.1
│   └── crypto-common v0.2.0-rc.3
│       └── hybrid-array v0.3.1 (*)
├── elliptic-curve v0.14.0-rc.10
│   ├── base16ct v0.2.0
│   ├── crypto-bigint v0.7.0-pre.6
[...]

PR #1296, opened 2025-07-13
`hash2curve`: make `hash_to_field` output an array instead of taking an out parameter.

carloskiki:
Discussed as part of #1295.

### DIFF
--- a/hash2curve/src/hash2field.rs
+++ b/hash2curve/src/hash2field.rs
@@ -38,21 +38,23 @@ pub trait FromOkm {
 /// [`ExpandMsgXmd`]: crate::hash2field::ExpandMsgXmd
 /// [`ExpandMsgXof`]: crate::hash2field::ExpandMsgXof
 #[doc(hidden)]
-pub fn hash_to_field<E, K, T>(data: &[&[u8]], domain: &[&[u8]], out: &mut [T]) -> Result<()>
+pub fn hash_to_field<E, K, T, C>(data: &[&[u8]], domain: &[&[u8]]) -> Result<Array<T, C>>
 where
     E: ExpandMsg<K>,
     T: FromOkm + Default,
+    C: ArraySize,
 {
     let len_in_bytes = T::Length::USIZE
-        .checked_mul(out.len())
+        .checked_mul(C::USIZE)
         .and_then(|len| len.try_into().ok())
         .and_then(NonZeroU16::new)
         .ok_or(Error)?;
     let mut tmp = Array::<u8, <T as FromOkm>::Length>::default();
     let mut expander = E::expand_message(data, domain, len_in_bytes)?;
+    let mut out = Array::<T, C>::default();
     for o in out.iter_mut() {
         expander.fill_bytes(&mut tmp);
         *o = T::from_okm(&tmp);
     }
-    Ok(())
+    Ok(out)
 }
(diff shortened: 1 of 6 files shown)


daxpedda:
This was attempted before in https://github.com/RustCrypto/traits/pull/872#issuecomment-1007409241.

AFAICS we decided against it for two reason:
- Not all lengths can be expressed with `ArraySize`.
- Monomorphization causing extreme compilation and binary bloat.

carloskiki:
> * Not all lengths can be expressed with `ArraySize`.

I used `hybrid_array::Array` because that's what the crate uses elsewhere but in hindsight we could use primitive (`[T; N]`) arrays instead.

> Monomorphization causing extreme compilation and binary bloat.

I don't think this is really that much of an issue, we are only using the function to generate arrays of 1 or 2 elements, and we may not even want to expose this function publicly. I think out parameters are generally harder to optimize than returning a value: https://www.youtube.com/watch?v=FnGCDLhaxKU&t=5458s, but that may not be true at all & is probably not worth changing a function for it.

I don't think this is a big deal at all, feel free to close this PR. I don't want to bikeshed over useless details, this was just something silly I came across when trying to grasp the codebase. I can also refactor to use `[T; N]` if you want the change.

daxpedda:
Oh no, this is totally my bad. I totally jumped to conclusions without taking a proper look.
I was thinking of the length parameter in `hash_from_bytes()`, `encode_from_bytes()` and `hash_to_scalar()`.

This is a great change! And yes, if you could change it to `const N` that would be nice.
FYI: I'm not a maintainer here, I'm just dropping my unasked 2¢ left and right :)

hash2curve/src/hash2field.rs
```diff
@@ -38,21 +38,23 @@ pub trait FromOkm {
 /// [`ExpandMsgXmd`]: crate::hash2field::ExpandMsgXmd
 /// [`ExpandMsgXof`]: crate::hash2field::ExpandMsgXof
 #[doc(hidden)]
-pub fn hash_to_field<E, K, T>(data: &[&[u8]], domain: &[&[u8]], out: &mut [T]) -> Result<()>
+pub fn hash_to_field<E, K, T, C>(data: &[&[u8]], domain: &[&[u8]]) -> Result<Array<T, C>>
 where
     E: ExpandMsg<K>,
     T: FromOkm + Default,
+    C: ArraySize,
```
daxpedda: Actually, if we leave this to `ArraySize` instead of `const C`, we could do the following:
```suggestion
    T: FromOkm + Default,
    C: ArraySize,
    T::Length: Mul<C, Output: IsLess<U65536, Output = True>>,
```
Eliminating one more run-time error.

However propagating this requirement up for users might be quite ugly. So I'm in favor of actually going ahead with the const generic, but thought I'd drop the thought at least.

carloskiki: Yes this is a good idea, but I don't like the complex trait bounds either. With inline consts we could do something like:
```rust
const { if N * T::Length > u16::MAX {
    compile_error!("...");
} };
```
Although this causes [non-local errors](https://hackmd.io/dTZWM-kVTlquPXMJ3KKsVw) and so I did not add it for now.

daxpedda (approved):

tarcieri (approved):

PR #1302, opened 2025-07-18
ed448-goldilocks: assorted fixes and improvements

daxpedda:
This PR includes some assorted fixes and improvements from #1291.
They are split into descriptive commits for your convenience.

### DIFF
--- a/ed448-goldilocks/src/field/element.rs
+++ b/ed448-goldilocks/src/field/element.rs
@@ -335,7 +335,7 @@ impl FieldElement {
         l0 = l2 * l1;
         l2 = l0.square_n::<9>();
         l1 = l0 * l2;
-        l0 = l1 * l1;
+        l0 = l1.square();
         l2 = l0 * self;
         l0 = l2.square_n::<18>();
         l2 = l1 * l0;


daxpedda:
I was wondering why this PR looks so big. But I just forgot to rebase it on top of #1300.
It should be fairly small now.

ed448-goldilocks/src/constants.rs
```diff
@@ -12,9 +12,3 @@ pub const DECAF_BASEPOINT: DecafPoint = DecafPoint(curve::twedwards::extended::E
 /// \ell = 2^\{446\} + 0x8335dc163bb124b65129c96fde933d8d723a70aadc873d6d54a7bb0d.
 /// $$
 pub const EDWARDS_BASEPOINT_ORDER: EdwardsScalar = EdwardsScalar::new(ORDER);
-
-/// `BASEPOINT_ORDER` is the order of the Decaf448 basepoint, i.e.,
-/// $$
-/// \ell = 2^\{446\} + 0x8335dc163bb124b65129c96fde933d8d723a70aadc873d6d54a7bb0d.
-/// $$
-pub const DECAF_BASEPOINT_ORDER: DecafScalar = DecafScalar::new(ORDER);
```
tarcieri: Hmm, these both seem problematic, especially in the `pub` API, as they're out-of-range scalars (i.e. the highest valid scalar is `ORDER - 1`)

ed448-goldilocks/src/decaf/points.rs
```diff
@@ -226,7 +226,7 @@ impl CofactorGroup for DecafPoint {
     }
 
     fn is_torsion_free(&self) -> Choice {
-        (self * DECAF_BASEPOINT_ORDER).ct_eq(&Self::IDENTITY)
+        self.ct_eq(&Self::IDENTITY)
     }
```
tarcieri: Shouldn't this always be true because Decaf is a prime order group?

daxpedda: Yeah, that's what I wrote in the git commit message as well ...
Fixed it.

ed448-goldilocks/src/field/element.rs
```diff
@@ -276,13 +275,10 @@ impl FieldElement {
     }
 
     /// Squares a field element  `n` times
-    fn square_n(&self, mut n: u32) -> FieldElement {
-        let mut result = self.square();
+    fn square_n<const N: u32>(&self) -> FieldElement {
```
tarcieri, line 278: Curious if this change actually impacts codegen in any way (i.e. for the better)

ed448-goldilocks/src/constants.rs
daxpedda: Hm, I'm not sure how else to do this.
I'm thinking of scrapping the `CofactorGroup` implementation entirely, WDYT?

tarcieri: For Decaf or the Edwards form? The impl for `Decaf` seems trivial.

For Edwards I think you can leave it as is, but change it to `pub(crate) const EDWARDS_BASEPOINT_ORDER` so you don't leak the invalid scalar in the public API.

daxpedda: I just inlined it, WDYT?

tarcieri: That works!

ed448-goldilocks/src/field/element.rs
daxpedda: I didn't check the codegen, but I benchmarked the inversion method and there were no changes.

tarcieri: Feels kind of six of one, half dozen of another in that case

daxpedda: I made this change to prevent accidental variable time usage.
But let me know and I can just remove it.

tarcieri (approved):

PR #1304, opened 2025-07-18
ed448-goldilocks: export LOW_A/LOW_B/LOW_C points

baloo:
This is used in x448 (https://github.com/RustCrypto/elliptic-curves/pull/1250)


daxpedda (changes requested):
Lets make them associated constants on `MontgomeryPoint`.
Otherwise LGTM!

ed448-goldilocks/src/montgomery.rs
```diff
@@ -18,20 +18,22 @@ use core::fmt;
 use core::ops::Mul;
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
 
-// Low order points on Curve448 and it's twist
-const LOW_A: MontgomeryPoint = MontgomeryPoint([
+/// First low order point on Curve448 and it's twist
+pub const LOW_A: MontgomeryPoint = MontgomeryPoint([
```
tarcieri:
```suggestion
impl MontgomeryPoint {
    /// First low order point on Curve448 and it's twist
    pub const LOW_A: Self = Self([
```

tarcieri (approved):

PR #1305, opened 2025-07-18
ed448-goldilocks: expose `CurveWithScalar`

baloo:
This is used in x448 (https://github.com/RustCrypto/elliptic-curves/pull/1250)


daxpedda (commented):
I was able to build `x448` without exposing `CurveWithScalar`.
Why was this necessary on your end?

baloo:
I think it built, but it was not exposed in the doc, which made it harder to reason about.
This might not be necessary now we have the `MontgomeryXpoint`

daxpedda:
My intention actually was to hide them from the docs.
All the methods provided by this trait are re-exposed on the `Scalar` directly.

If you think there is still an issue in the docs, let me know in detail as I would like to take a look at it.

I'm still playing with the alternative to turn `scalar.rs` into a huge macro so the scalars can become types instead of typedef's ... but maybe we can find another alternative.

baloo:
Yeah you're correct, it was not required.

PR #1306, opened 2025-07-19
`Curve448` with full coordinates

daxpedda:
> This PR implements a set of types for Montgomery points with a full coordinate system and corresponding `Curve448` type with a `CurveArithmetic` implementation. Our current `MontgomeryPoint` x-coordinate only remains in place for use with `X448` and is renamed to `MontgomeryXpoint`.
>
> For context: the x-only coordinate system can't implement a full set of arithmetic operations because of the missing y-coordinate. While the y-coordinate could be recovered, it is costly and is missing the sign. Some protocols exist that set the sign of the y-coordinate, which is why we have some methods in place for `MontgomeryXpoint` to do exactly that.
>
> I made sure to add a full set of conversion methods between the new types.

I'm aware that I'm proposing very large changes that have not previously been discussed. I'm happy to take any feedback.
It should be much easier to review than #1291 on account that it can be done commit by commit.

Take 2 on #1291.

### DIFF
--- a/ed448-goldilocks/src/montgomery/point.rs
+++ b/ed448-goldilocks/src/montgomery/point.rs
@@ -129,8 +129,11 @@ impl From<&AffineMontgomeryPoint> for AffinePoint {
         t4 = t1 * t2; // 2x^2y^2+2y^2
         let yDen = t5 - t4; // yDen = x^5-2x^3+x-2x^2y^2-2y^2
 
-        let x = xNum * xDen.invert();
-        let y = yNum * yDen.invert();
+        let commonDen = xDen * yDen;
+        let invDen = commonDen.invert();
+
+        let x = xNum * yDen * invDen;
+        let y = yNum * xDen * invDen;
 
         AffinePoint::conditional_select(
             &AffinePoint { x, y },
--- a/ed448-goldilocks/src/edwards/extended.rs
+++ b/ed448-goldilocks/src/edwards/extended.rs
@@ -333,12 +333,14 @@ impl EdwardsPoint {
 
         let affine = self.to_affine();
 
-        // TODO: optimize to a single inversion.
         let xx = affine.x.square();
         let yy = affine.y.square();
 
-        let u = yy * xx.invert();
-        let v = (FieldElement::TWO - xx - yy) * affine.y * (xx * affine.x).invert();
+        let v_den = xx * affine.x;
+        let den = (xx * v_den).invert();
+
+        let u = yy * v_den * den;
+        let v = (FieldElement::TWO - xx - yy) * affine.y * xx * den;
 
         AffineMontgomeryPoint::conditional_select(
             &AffineMontgomeryPoint::new(u, v),


ed448-goldilocks/src/montgomery/x.rs
```diff
@@ -0,0 +1,457 @@
...
+const LOW_C: MontgomeryXpoint = MontgomeryXpoint([
+    0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xfe, 0xff, 0xff, 0xff,
+    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+]);
+
+/// A point in Montgomery form
+#[derive(Copy, Clone)]
+pub struct MontgomeryXpoint(pub [u8; 56]);
```
baloo, line 41: Pedantic question: shouldn't that be named `MontgomeryXPoint` to abide by https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#general-conventions-[rfc-#430] ?

```suggestion
pub struct MontgomeryXPoint(pub [u8; 56]);
```

tarcieri: @baloo debatable. I’d refer to this version: https://rust-lang.github.io/api-guidelines/naming.html

> Usize rather than USize or Stdin rather than StdIn

The rules seem to favor eliminating superfluous capital letters where it makes sense

daxpedda: Yep, Rust API guidelines is also what I followed in this case.

carloskiki:
Do you have any protocol in mind where this would be used?

daxpedda:
Yes, I am planning on using it with OPRF inside OPAQUE.

ed448-goldilocks/src/constants.rs
```diff
@@ -18,3 +18,9 @@ pub const EDWARDS_BASEPOINT_ORDER: EdwardsScalar = EdwardsScalar::new(ORDER);
 /// \ell = 2^\{446\} + 0x8335dc163bb124b65129c96fde933d8d723a70aadc873d6d54a7bb0d.
 /// $$
 pub const DECAF_BASEPOINT_ORDER: DecafScalar = DecafScalar::new(ORDER);
+
+/// `BASEPOINT_ORDER` is the order of the Curve448 basepoint, i.e.,
```
baloo:
```suggestion
/// [`MONTGOMERY_BASEPOINT_ORDER`] is the order of the Curve448 basepoint, i.e.,
```

daxpedda: We completely removed that constant: https://github.com/RustCrypto/elliptic-curves/pull/1302#discussion_r2217808854.

ed448-goldilocks/src/montgomery/point.rs
```diff
@@ -0,0 +1,574 @@
...
+};
+use crate::field::{ConstMontyType, FieldElement};
+use crate::{AffinePoint, Curve448, Curve448FieldBytes, ORDER};
+
+/// A point in Montgomery form including the y-coordinate.
+#[derive(Copy, Clone, Debug, Default, Eq)]
+pub struct MontgomeryPoint {
+    pub(super) x: FieldElement,
+    pub(super) y: FieldElement,
+}
```
tarcieri: I know the coordinate names for Montgomery are confusing as hell to begin with, but I would probably expect this to match the coordinate names for `ProjectiveMontgomeryPoint`, i.e. `U` and `V`.

ed448-goldilocks/src/montgomery/point.rs
```diff
@@ -0,0 +1,574 @@
...
+use super::{
+    DEFAULT_ENCODE_TO_CURVE_SUITE, DEFAULT_HASH_TO_CURVE_SUITE, MontgomeryScalar, MontgomeryXpoint,
+    ProjectiveMontgomeryXpoint,
+};
+use crate::field::{ConstMontyType, FieldElement};
+use crate::{AffinePoint, Curve448, Curve448FieldBytes, ORDER};
+
+/// A point in Montgomery form including the y-coordinate.
+#[derive(Copy, Clone, Debug, Default, Eq)]
+pub struct MontgomeryPoint {
```
tarcieri: Perhaps this should be `AffineMontgomeryPoint`?

I think calling it `MontgomeryPoint` makes it seem like the main type users should be reaching for when it's really a niche type

ed448-goldilocks/src/montgomery/point.rs
daxpedda: Hm, I was going after the papers here, which consistently referred to X and Y when dealing in affine form and U, V and W when dealing with projective points (AFAIR). But yeah, I don't mind ignoring all that and making it a bit easier on ourselves.

ed448-goldilocks/src/montgomery/point.rs
daxpedda: Yeah I think this is a great solution actually.

ed448-goldilocks/src/montgomery/point.rs
tarcieri: IMO there isn't a whole lot of consistency. Some papers use x,y and some use u,v. This one for example uses u,v:

https://tches.iacr.org/index.php/TCHES/article/view/11808/11313

> 2.2  Montgomery Curve
>
> A Montgomery curve over 𝔽q is defined by the affine equation:
> Bv² = u³ + Au² + u

I just thought we could be internally consistent with the coordinate names.

Edit: I think the main reason for using u,v for the Montgomery form is to disambiguate from the Weierstrass form, where every curve in the Montgomery form can also be expressed in the Weierstrass form.

daxpedda: I see.
Thank you for the context and actually finding an example on the quick!

ed448-goldilocks/src/montgomery/point.rs
```diff
@@ -0,0 +1,580 @@
...
+            CtOption::new(
+                ProjectiveMontgomeryXpoint {
+                    U,
+                    W: FieldElement::ONE,
+                }
+                .to_extended(sign),
+                valid,
+            )
+        })
+    }
```
daxpedda, line 445: Just wanted to highlight that I'm using SEC1 encoding here considering that I found no standard at all to encode full-coordinate Curve448.

Alternatively we can just follow Edwards448 here.

tarcieri: These curves definitely don't use SEC1. You can shove the sign bit in the most significant bit (i.e. bit 255), but that's nonstandard as it's specified to be set to 0 (at least for X25519).

daxpedda: Yeah that's why I went with SEC1, because the SEC1 specification doesn't say anywhere its limited to Weierstrass curves.

But I'm happy to follow the route you are suggesting here.

daxpedda: Done.

PR #1308, opened 2025-07-19
Optimize Montgomery to Edwards conversions

daxpedda:
This PR updates the Montgomery to Edwards isogeny mapping to a more optimized version.
Notably reducing inversions from two to one.

This is currently only used for hash2curve, which actually has a more optimized mapping I will add in a different PR. This functionality will however be used when adding Montgomery to Edwards conversions in #1306.

### DIFF
--- a/ed448-goldilocks/src/edwards/affine.rs
+++ b/ed448-goldilocks/src/edwards/affine.rs
@@ -95,35 +95,48 @@ impl AffinePoint {
         }
     }
 
+    // https://www.rfc-editor.org/rfc/rfc7748#section-4.2
     pub(crate) fn isogeny(&self) -> Self {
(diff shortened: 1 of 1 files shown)


tarcieri:
I'm a little confused what's happening here. The input and output types are both `AffinePoint` which is ostensibly the untwisted Edwards form. What is this isogeny between? It looks like you're computing the Montgomery coordinates from the Edwards affine ones?

daxpedda:
Indeed. As I was saying in the OP: this is currently only used in hash2curve.

This PR just improves existing code, I didn't add this function. But #1306 properly exposes this function to users and the Montgomery to Edwards conversion.

Alternatively we can just close this as this improvement is included in #1306.

PR #1309, opened 2025-07-19
Use optimized `map_to_curve_elligator2_curve448()`

daxpedda:
This implements Elligator2 mapping according to [RFC 9380](https://www.rfc-editor.org/rfc/rfc9380.html#name-curve448-q-3-mod-4-k-1) for Curve448 and Ed448.

The current Curve448 mapping was using either the unoptimized or an outdated version. For Ed448 we were currently using the isogeny mapping, which works correctly ofc. However, RFC 9380 specifically outlines an optimized mapping for Ed448 in combination of the Curve448 mapping. This avoids unnecessary inversions between mapping from Curve448 to Ed448.

While Curve448 will first be introduced in #1306, the Ed448 alligator mapping uses the Curve448 mapping.

I simply commented out `fn isoginy()` as it is now unused. However, it will be adapted in #1306 to facilitate Montgomery to Edwards conversions.


ed448-goldilocks/src/edwards/affine.rs
```diff
@@ -69,36 +69,36 @@ impl AffinePoint {
...
+    //     let yNum = t4 - t5; // yNum = -(x^5-2x^3+x-4xy^2)
+
+    //     t4 = t1 * t2; // 2x^2y^2+2y^2
+    //     let yDen = t5 - t4; // yDen = x^5-2x^3+x-2x^2y^2-2y^2
+
+    //     Self {
+    //         x: xNum * xDen.invert(),
+    //         y: yNum * yDen.invert(),
+    //     }
+    // }
```
tarcieri: Can you just remove this?

Getting this PR reviewed seems doable, but #1306 is quite large and might take awhile

daxpedda: Sure!

daxpedda: Done.

PR #1311, opened 2025-07-19
[WIP] primefield: generic `MontyFieldElement` type

tarcieri:
The previous implementation was written entirely in terms of macros.

Leveraging types from `crypto-bigint`, this provides a generic field element type with an internal Montgomery form representation.


primefield/src/monty.rs
```diff
@@ -0,0 +1,178 @@
...
+        let array = Array::try_from(slice).ok()?;
+        Self::from_bytes(&array).into()
+    }
+
+    /// Decode a field element from hex-encoded bytes.
+    ///
+    /// Does *not* perform a check that the field element does not overflow the order.
+    ///
+    /// This method is primarily intended for defining internal constants.
+    pub const fn from_hex(hex: &str) -> Self {
```
tarcieri: One drawback is this is no longer `pub(crate)`.

I tried adding compile-time checks the modulus is in range, but it seems the relevant methods to do `const fn` comparisons are currently `pub(crate)` (i.e. `Uint::lt`)

primefield/src/monty.rs
```diff
@@ -0,0 +1,444 @@
...
+    #[must_use]
+    pub const fn square(&self) -> Self {
+        self.multiply(self)
+    }
+
+    /// Compute field inversion: `1 / self`.
+    #[inline]
+    pub fn invert<const UNSAT_LIMBS: usize>(&self) -> CtOption<Self>
+    where
+        Odd<Uint<LIMBS>>: PrecomputeInverter<Inverter = SafeGcdInverter<LIMBS, UNSAT_LIMBS>, Output = Uint<LIMBS>>,
```
tarcieri: This bound is atrocious but can hopefully go away with binary XGCD (RustCrypto/crypto-bigint#875)

tarcieri: Simplified it to:

```rust
MontyForm<MOD, LIMBS>: Invert<Output = CtOption<MontyForm<MOD, LIMBS>>>
```

primefield/src/monty.rs
tarcieri: I added the checks which makes it slightly less worrisome

PR #1316, opened 2025-07-19
Optimize Edwards isogeny map

daxpedda:
This PR optimizes the Edwards isogeny map. No special algorithms was used, just variables cached and the common denominator calculated to save an inversion.


tarcieri:
The 4-isogeny described in #1349 appears to be more efficient and avoids inversions entirely

tarcieri (approved):
Per your comments on #1349 I'm curious why the twisted-to-untwisted map from libgoldilocks isn't working.

That said, the one you've implemented is both well-cited and has reduced the number of inversions to one, which is certainly a big improvement.

daxpedda:
> Per your comments on #1349 I'm curious why the twisted-to-untwisted map from libgoldilocks isn't working.

I'm gonna spend some time figuring out how to make it work. I didn't dive into the mathematical gymnastics done in https://github.com/otrv4/libgoldilocks/blob/d07cb5b423995bae1155702aa949846c95d855c1/src/goldilocks.c#L980-L994 as well, but I would prefer to understand what has been done there instead of just leaving a comment saying "copied code from XYZ".

tarcieri:
I'm going to go ahead and merge this but leave #1349 open

tarcieri:
@daxpedda here's the paper that Mike Hamburg cites on Edwards 4-isogenies: https://eprint.iacr.org/2011/135.pdf

tarcieri:
>  I didn't dive into the mathematical gymnastics done in https://github.com/otrv4/libgoldilocks/blob/d07cb5b423995bae1155702aa949846c95d855c1/src/goldilocks.c#L980-L994 as well, but I would prefer to understand what has been done there instead of just leaving a comment saying "copied code from XYZ".

That's the isogeny from this Mike Hamburg paper: https://eprint.iacr.org/2014/027.pdf

Screenshot 2025-08-03 at 9 08 38 AM

...except it's in extended Edwards coordinates, so `x = X/Z, y = Y/Z, T = XY/Z^2`

daxpedda:
Yeah, just figured that out as well.
I also remembered that both conversions are actually the same, the only difference is `a`.
Already got it working, just adding comments how it was derived and such.

PR #1317, opened 2025-07-20
Use `Iterator<Item = u8>` instead of `Expander`

carloskiki:
Discussed in https://github.com/RustCrypto/elliptic-curves/issues/1295.

## Changes

- `Expander` trait removed in favor of `Iterator`.
- `ExpandMsgXmd` and `ExpandMsgXof` stop generating bytes after having reached `len_in_bytes`.
- `ExpandMsgXmd` and `ExpandMsgXof` are themselves the "expanders", instead of having an `Expander` associated type.

## Regression

On small workloads (`len_in_bytes < 256`), both xmd and xof expanders are ~ 20% slower with `Iterator` than with `Expander`. On the largest workloads (`len_in_bytes = 4k`), xmd is 50% slower and xof is 100% slower with iterator. I think `ExpandMsgXof` could be made equally as fast if `XofReader` allowed a method that generates an Iterator over its inner buffer, just like how `ExpandMsgXmd` is implemented. I also think that adding bound checks (like there should have been) on `Expander` would cause similar slow downs on small workloads.

### Metrics

small (frequent use case):
```
expand_xmd              time:   [168.61 ns 168.68 ns 168.76 ns]

expand_xmd_iter         time:   [226.10 ns 226.51 ns 227.08 ns]

expand_xof              time:   [489.15 ns 490.44 ns 492.19 ns]

expand_xof_iter         time:   [532.37 ns 532.61 ns 532.92 ns]
```

large:
```
expand_xmd              time:   [12.074 µs 12.078 µs 12.083 µs]

expand_xmd_iter         time:   [18.433 µs 18.440 µs 18.449 µs]

expand_xof              time:   [7.0494 µs 7.0514 µs 7.0538 µs]

expand_xof_iter         time:   [15.713 µs 15.718 µs 15.725 µs]
```

I did not include the benchmarks in the PR because I did not want to add criterion as a `dev-dependency`. Let me know if you want me to add them.

I let the maintainers decide whether this regression is worth it.


hash2curve/src/hash2field.rs
```diff
@@ -38,20 +38,24 @@ pub trait FromOkm {
...
         .and_then(NonZeroU16::new)
         .ok_or(Error)?;
-    let mut tmp = Array::<u8, <T as FromOkm>::Length>::default();
     let mut expander = E::expand_message(data, domain, len_in_bytes)?;
     Ok(core::array::from_fn(|_| {
-        expander.fill_bytes(&mut tmp);
+        let tmp = Array::<u8, <T as FromOkm>::Length>::from_iter(
+            expander.by_ref().take(T::Length::USIZE),
+        );
         T::from_okm(&tmp)
```
daxpedda:
```suggestion
        T::from_okm(&expander.by_ref().take(T::Length::USIZE).collect())
```

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -37,23 +45,18 @@ where
...
+    ) -> Result<Self> {
         let b_in_bytes = HashT::OutputSize::USIZE;
 
         // `255 * <b_in_bytes>` can not exceed `u16::MAX`
         if usize::from(len_in_bytes.get()) > 255 * b_in_bytes {
             return Err(Error);
         }
 
-        let ell = u8::try_from(usize::from(len_in_bytes.get()).div_ceil(b_in_bytes))
-            .expect("should never pass the previous check");
```
daxpedda, line 55: Lets turn this into a `debug_assert!`.

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -75,74 +78,49 @@ where
...
+            .zip(&self.b_vals[..])
+            .enumerate()
+            .for_each(|(j, (b0val, bi1val))| tmp[j] = b0val ^ bi1val);
+        let mut b_vals = HashT::default();
+        b_vals.update(&tmp);
+        b_vals.update(&[self.index]);
+        self.domain.update_hash(&mut b_vals);
+        b_vals.update(&[self.domain.len()]);
+        self.b_vals = b_vals.finalize_fixed();
+        Some(self.b_vals[0])
```
daxpedda: I didn't get at first why this change was necessary, but apparently you discovered a bug! We previously calculated the limit by just assuming that we will run out of bytes when we reach the end of `ell`. But all the hashes concatenated can actually exceed `len_in_bytes`.

The [spec is also clear about this](https://www.rfc-editor.org/rfc/rfc9380.html#section-5.3.1-6):
> 12. return substr(uniform_bytes, 0, len_in_bytes)

However, I suggest simplifying the code to the following:
```suggestion
        if self.remaining == 0 {
            return None;
        }

        if self.offset == self.b_vals.len() {
            self.index += 1;
            self.offset = 0;
            // b_0 XOR b_(idx - 1)
            let mut tmp = Array::<u8, HashT::OutputSize>::default();
            self.b_0
                .iter()
                .zip(&self.b_vals[..])
                .enumerate()
                .for_each(|(j, (b0val, bi1val))| tmp[j] = b0val ^ bi1val);
            let mut b_vals = HashT::default();
            b_vals.update(&tmp);
            b_vals.update(&[self.index]);
            self.domain.update_hash(&mut b_vals);
            b_vals.update(&[self.domain.len()]);
            self.b_vals = b_vals.finalize_fixed();
        }

        let byte = self.b_vals[self.offset];
        self.offset += 1;
        self.remaining -= 1;
        Some(byte)
```
Where `ExpandMsgXmd::remaining` is just `len_in_bytes`.
I believe this code should make it quite clear on whats going on here.

hash2curve/src/hash2field/expand_msg/xof.rs
```diff
@@ -66,16 +62,27 @@ where
...
+    type Item = u8;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.length == 0 {
+            return None;
+        }
+        self.length -= 1;
+        let mut buf = [0u8; 1];
+        self.reader.read(&mut buf);
+        Some(buf[0])
```
daxpedda: See [`array::from_mut()`](https://doc.rust-lang.org/nightly/core/array/fn.from_mut.html) to get rid of the array.
```suggestion
        let mut byte = 0;
        self.reader.read(array::from_mut(&mut byte));
        Some(byte)
```

ed448-goldilocks/src/field/element.rs
```diff
@@ -463,16 +463,16 @@ mod tests {
                 (84 * 2).try_into().unwrap(),
             )
             .unwrap();
-            let mut data = Array::<u8, U84>::default();
-            expander.fill_bytes(&mut data);
+            let mut data = Array::<u8, U84>::from_iter(expander.by_ref().take(84));
             // TODO: This should be `Curve448FieldElement`.
             let u0 = Ed448FieldElement::from_okm(&data).0;
```
daxpedda:
```suggestion
            // TODO: This should be `Curve448FieldElement`.
            let u0 = Ed448FieldElement::from_okm(&expander.by_ref().take(84).collect()).0;
```
We can do this in a lot of other places as well.

daxpedda (changes requested):
I'm not terribly concerned about the slowdown. Overall executing the whole hash2curve operation is quite expensive compared to the `Expander` part, I would assume.

However I'm uncertain now what we are trying to achieve here. We are getting rid of a trait in favor of worse performance and improving the feedback (returning `None` when we are done).

In hindsight, what we want to model is the [`Read`](https://doc.rust-lang.org/nightly/std/io/trait.Read.html) API. This would be an accurate representation to whats going on there. Unfortunately the `Read` trait isn't available on no_std.

I will make a PR to propose an alternative, then we can do a comparison and see what we get.

daxpedda:
See #1318.

hash2curve/src/hash2field/expand_msg/xmd.rs
carloskiki: I think the logic here is succinct enough to not need a dbg assert. Since `len_in_bytes <= 255 * b_in_bytes`, `len_in_bytes / b_in_bytes <= 255`.

hash2curve/src/hash2field/expand_msg/xmd.rs
carloskiki: Good catch! I wasn't sure if this was an oversight or an assumption that the users would not ask for more bytes.

carloskiki:
#1318 raises the concern that the `ExpandMsg` trait now requires a lifetime parameter. I could change that by returning an opaque `impl Iterator<Item = u8> + use<'dst>`, or use an associated type as before if desired.

hash2curve/src/hash2field/expand_msg/xmd.rs
daxpedda: Seems fine to me!

daxpedda:
> #1318 raises the concern that the `ExpandMsg` trait now requires a lifetime parameter. I could change that by returning an opaque `impl Iterator<Item = u8> + use<'dst>`, or use an associated type as before if desired.

Yes please, removing that lifetime from `ExpandMsg` made it a whole lot easier to use that trait in more complicated generic setups downstream.

carloskiki:
The trait no longer has an input lifetime, it has an associated type `Expanded`, analogous to the current `Expander`.

tarcieri:
Closing in favor of #1318

PR #1318, opened 2025-07-20
hash2curve `Expander` improvements

daxpedda:
This PR does three changes:
- Fixes `Expander`s outputting more than `len_in_bytes`.
- Lets `Expander::fill_bytes()` return `Result`, to more closely mimic a rusty `Read`er.
- Optimizes `ExpandMsgXmd` to copy whole slices instead of byte by byte.

This is an alternative take to addressing the issues #1317 found.
However, this PR doesn't get rid of any traits or types. I originally intended to get rid of `ExpanderXmd`, but I remembered again why we kept it in the first place: to remove the lifetime from `ExpandMsg`.

### DIFF
--- a/hash2curve/src/hash2field.rs
+++ b/hash2curve/src/hash2field.rs
@@ -51,7 +51,9 @@ where
     let mut tmp = Array::<u8, <T as FromOkm>::Length>::default();
     let mut expander = E::expand_message(data, domain, len_in_bytes)?;
     Ok(core::array::from_fn(|_| {
-        expander.fill_bytes(&mut tmp);
+        expander
+            .fill_bytes(&mut tmp)
+            .expect("never exceeds `len_in_bytes`");
         T::from_okm(&tmp)
     }))
 }
--- a/hash2curve/src/hash2field/expand_msg.rs
+++ b/hash2curve/src/hash2field/expand_msg.rs
@@ -39,8 +39,12 @@ pub trait ExpandMsg<K> {
 
 /// Expander that, call `read` until enough bytes have been consumed.
 pub trait Expander {
-    /// Fill the array with the expanded bytes
-    fn fill_bytes(&mut self, okm: &mut [u8]);
+    /// Fill the array with the expanded bytes, returning how many bytes were read.
+    ///
+    /// # Errors
+    ///
+    /// If no bytes are left.
+    fn fill_bytes(&mut self, okm: &mut [u8]) -> Result<usize>;
 }
 
 /// The domain separation tag
--- a/hash2curve/src/hash2field/expand_msg/xmd.rs
+++ b/hash2curve/src/hash2field/expand_msg/xmd.rs
@@ -171,6 +171,31 @@ mod test {
     use hex_literal::hex;
     use sha2::Sha256;
 
+    #[test]
+    fn edge_cases() {
+        fn generate() -> ExpanderXmd<'static, Sha256> {
+            <ExpandMsgXmd<Sha256> as ExpandMsg<U4>>::expand_message(
+                &[b"test message"],
+                &[b"test DST"],
+                NonZero::new(64).unwrap(),
+            )
+            .unwrap()
+        }
+
+        assert_eq!(generate().fill_bytes(&mut [0; 0]), Ok(0));
+        assert_eq!(generate().fill_bytes(&mut [0; 1]), Ok(1));
+        assert_eq!(generate().fill_bytes(&mut [0; 64]), Ok(64));
+        assert_eq!(generate().fill_bytes(&mut [0; 65]), Ok(64));
+
+        let mut expander = generate();
+        assert_eq!(expander.fill_bytes(&mut [0; 0]), Ok(0));
+        assert_eq!(expander.fill_bytes(&mut [0; 32]), Ok(32));
+        assert_eq!(expander.fill_bytes(&mut [0; 0]), Ok(0));
+        assert_eq!(expander.fill_bytes(&mut [0; 31]), Ok(31));
+        assert_eq!(expander.fill_bytes(&mut [0; 2]), Ok(1));
+        assert_eq!(expander.fill_bytes(&mut [0; 1]), Err(Error));
+    }
+
     fn assert_message<HashT>(
         msg: &[u8],
         domain: &Domain<'_, HashT::OutputSize>,
--- a/hash2curve/src/hash2field/expand_msg/xmd.rs
+++ b/hash2curve/src/hash2field/expand_msg/xmd.rs
@@ -51,8 +51,10 @@ where
             return Err(Error);
         }
 
-        let ell = u8::try_from(usize::from(len_in_bytes.get()).div_ceil(b_in_bytes))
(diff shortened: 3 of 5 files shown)


carloskiki:
I find the `Read` API overkill for this. We know there can't be errors while reading from the buffer, and we already know in advance how many bytes will be read. Even before calling the function it is known that the call will return `Ok(min(len_in_bytes, buffer.len()))`. Also by convention, [`read`](https://doc.rust-lang.org/stable/std/io/trait.Read.html#tymethod.read) should return `Ok(0)` when EOF is reached, not an error.

> If the return value of this method is [Ok(n)](https://doc.rust-lang.org/stable/std/result/enum.Result.html#variant.Ok), then implementations must guarantee that 0 <= n <= buf.len(). A nonzero n value indicates that the buffer buf has been filled in with n bytes of data from this source. If n is 0, then it can indicate one of two scenarios:
>  - This reader has reached its “end of file” and will likely no longer be able to produce bytes. Note that this does not mean that the reader will always no longer be able to produce bytes. As an example, on Linux, this method will call the recv syscall for a [TcpStream](https://doc.rust-lang.org/stable/std/net/struct.TcpStream.html), where returning zero indicates the connection was shut down correctly. While for [File](https://doc.rust-lang.org/stable/std/fs/struct.File.html), it is possible to reach the end of file and get zero as result, but if more data is appended to the file, future calls to read will return more data.
> - The buffer specified was 0 bytes in length.

As also written, `read` may be able to produce more bytes at a later time, which does not really fit our use case.

hash2curve/src/hash2field.rs
```diff
@@ -51,7 +51,9 @@ where
     let mut tmp = Array::<u8, <T as FromOkm>::Length>::default();
     let mut expander = E::expand_message(data, domain, len_in_bytes)?;
     Ok(core::array::from_fn(|_| {
-        expander.fill_bytes(&mut tmp);
+        expander
+            .fill_bytes(&mut tmp)
+            .expect("never exceeds `len_in_bytes`");
```
carloskiki, line 56: Hmmm, this is still wrong. If `fill_bytes` does not completely fill the array, then the untouched suffix contains garbage bytes.

To implement this correctly, we would have a sort of "event loop" that checks if the correct amount was filled and tries again to fill the array suffix if it was not filled completely... Otherwise we would need a strict contract for the trait implementation, which feels unrusty.

hash2curve/src/hash2field/expand_msg.rs
```diff
@@ -39,8 +39,12 @@ pub trait ExpandMsg<K> {
...
 /// Expander that, call `read` until enough bytes have been consumed.
 pub trait Expander {
-    /// Fill the array with the expanded bytes
-    fn fill_bytes(&mut self, okm: &mut [u8]);
+    /// Fill the array with the expanded bytes, returning how many bytes were read.
+    ///
+    /// # Errors
+    ///
+    /// If no bytes are left.
+    fn fill_bytes(&mut self, okm: &mut [u8]) -> Result<usize>;
```
carloskiki: We should clearly write the contract here for external implementors. They _must_ fill the array completely, and failure to do so signifies that the `Expander` is now empty, otherwise the hash2curve implementation generates undefined results. I feel like we should be using the type system to enforce these things at compile time, however.

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -156,6 +171,31 @@ mod test {
...
+        assert_eq!(generate().fill_bytes(&mut [0; 64]), Ok(64));
+        assert_eq!(generate().fill_bytes(&mut [0; 65]), Ok(64));
+
+        let mut expander = generate();
+        assert_eq!(expander.fill_bytes(&mut [0; 0]), Ok(0));
+        assert_eq!(expander.fill_bytes(&mut [0; 32]), Ok(32));
+        assert_eq!(expander.fill_bytes(&mut [0; 0]), Ok(0));
+        assert_eq!(expander.fill_bytes(&mut [0; 31]), Ok(31));
+        assert_eq!(expander.fill_bytes(&mut [0; 2]), Ok(1));
+        assert_eq!(expander.fill_bytes(&mut [0; 1]), Err(Error));
```
carloskiki, line 222: This case trips me up a bit. The standard `Read` api would return `Ok(0)` instead.

carloskiki (commented):
I feel like having this sort of API over complicates things here.

daxpedda:
> I find the `Read` API overkill for this. We know there can't be errors while reading from the buffer, and we already know in advance how many bytes will be read. Even before calling the function it is known that the call will return `Ok(min(len_in_bytes, buffer.len()))`.

But it is a reader API. This PR is choosing to expose it as-is instead of wrapping it into a less complicated one.

> Also by convention, [`read`](https://doc.rust-lang.org/stable/std/io/trait.Read.html#tymethod.read) should return `Ok(0)` when EOF is reached, not an error.


> As also written, `read` may be able to produce more bytes at a later time, which does not really fit our use case.

This isn't actually a copy of `Read`, it is trying to expose something useful, appropriate and familiar API to users. But, you can look at it like [`read_exact()`](https://doc.rust-lang.org/1.88.0/std/io/trait.Read.html#method.read_exact), which does return an error on EOF.

hash2curve/src/hash2field.rs
daxpedda: You would be correct if this is actually the `Read` API, but it isn't. This works correctly as-is.

hash2curve/src/hash2field/expand_msg.rs
daxpedda: Same as above.

hash2curve/src/hash2field/expand_msg/xmd.rs
daxpedda: I took more inspiration from `read_exact()` than `read()` here.

daxpedda:
I believe that invoking the ghost of the `Read` trait here was a mistake. I was just trying to take some inspiration and familiarity from it not actually copying it, despite how it looks like in the OP. My bad!

So why am I proposing this API?
The calls we are doing internally here are writing into slices. This is a fact, see `copy_from_slice()` and `XofReader::read()`. So it seems straightforward to me to expose it as-is. Is this the most convenient, easy-to-use and least-complicated API? No. But it is the most correct and direct API. This isn't supposed to be a high-level API. The high-level API is `GroupDigest`.

Couldn't we just have left the API as-is?
Right now there was no way for users to check how many bytes were written or when they actually did something wrong, like requested more bytes when there are none. Ofc we can just document exactly how many bytes this API will spit out and call it a day. But Rust gives us a way to encode this information via the type-system, I really don't see a reason not to do so.

In conclusion, I feel like the improvements here have a very clear purpose and direction to me. I'm not seeing the issue nor the downside here exactly.

However, while this change doesn't make the API more complicated to use compared to the current one (we are just exposing *more* information the user can happily ignore), it is significantly more complicated than #1317.

tarcieri (approved):

PR #1324, opened 2025-07-22
primefield: have `test_primefield_constants!` compute `T`

tarcieri:
...where `T` is `(modulus - 1) >> S`.

Previously this was precomputed manually on a curve-by-curve basis. This commit changes it to compute `T` from the other constants.

### DIFF
--- a/primefield/src/dev.rs
+++ b/primefield/src/dev.rs
@@ -1,8 +1,8 @@
 /// Implement all tests for a type which impls the `PrimeField` trait.
 #[macro_export]
 macro_rules! test_primefield {
-    ($fe:tt, $t:expr) => {
-        $crate::test_primefield_constants!($fe, $t);
+    ($fe:tt, $uint:ident) => {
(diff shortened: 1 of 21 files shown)


primefield/src/dev.rs
```diff
@@ -12,16 +12,36 @@ macro_rules! test_primefield {
...
+            core::array::from_fn(|i| {
+                let hi = words.get((2 * i) + 1).copied().unwrap_or_default();
+                let lo = words[2 * i];
+                (hi as u64) << 32 | (lo as u64)
+            })
+        }
+        #[cfg(target_pointer_width = "64")]
+        fn pow_args(n: &$uint) -> [u64; $uint::LIMBS] {
+            *n.as_words()
+        }
```
tarcieri, line 36: This part wound up being a little weird on 32-bit platforms since `pow_vartime` unconditionally takes a `u64` iterator. Perhaps some functionality can be added to `crypto-bigint` to iterate over `Uint` as either `u32` or `u64`

PR #1325, opened 2025-07-22
sm2: pke no std support

MaxKingPor:
add sm2 pke no std support
ref #1099

### DIFF
--- a/sm2/src/dsa.rs
+++ b/sm2/src/dsa.rs
@@ -50,11 +50,12 @@ use signature::{Error, Result, SignatureEncoding};
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
 
-#[cfg(all(feature = "alloc", feature = "pkcs8"))]
+#[cfg(feature = "pkcs8")]
 use crate::pkcs8::{
-    AlgorithmIdentifierRef, ObjectIdentifier, der, der::AnyRef,
-    spki::AssociatedAlgorithmIdentifier, spki::SignatureBitStringEncoding,
+    AlgorithmIdentifierRef, ObjectIdentifier, der::AnyRef, spki::AssociatedAlgorithmIdentifier,
 };
+#[cfg(all(feature = "alloc", feature = "pkcs8"))]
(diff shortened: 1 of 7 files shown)


sm2/src/dsa.rs
```diff
@@ -50,12 +50,11 @@ use signature::{Error, Result, SignatureEncoding};
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
 
-#[cfg(feature = "pkcs8")]
+#[cfg(all(feature = "alloc", feature = "pkcs8"))]
```
tarcieri: Why are you changing this? Everything but `SignatureBitStringEncoding` is available in no-`alloc` environments

tarcieri: Why did you resolve this comment without answering or changing anything?

MaxKingPor:
@tarcieri  I'm sorry, I've replied, but it seems that you can't see it over there
IMG_20250724_001532

---

These imports are only used in `impl` near line 215
https://github.com/RustCrypto/elliptic-curves/blob/20eeefff523f46c9ad8effb23add8164d4b53381/sm2/src/dsa.rs#L208-L232

tarcieri:
Oh, your comment is marked pending. I’m on a phone but I’m confused why there wouldn’t be unused import warnings in thar case.

MaxKingPor:
The current main branch uses std by default

tarcieri:
By "default" in the sense of a default-enabled `std` (and `alloc`) feature, not unconditionally:

https://github.com/RustCrypto/elliptic-curves/blob/8f6a343/sm2/Cargo.toml#L38-L40

MaxKingPor:
Sorry I didn't know this, but now I've changed it back

tarcieri:
Closing as stale

MaxKingPor:
@tarcieri Is this still needed?

tarcieri:
Yes, it would be a useful improvement

MaxKingPor:
I've reworked it, could you please take a look at #1687?

PR #1328, opened 2025-07-26
Fix `ConstantTimeEq` for `DecafPoint`

daxpedda:
Comparison of Decaf points requires normalization. To avoid inversion operations, we use scaling.

However, the current implementation for `DecafPoint`s forgets about having a Z-coordinate. This PR fixes that by just delegating the comparison implementation to the already correct implementation of `ExtendedPoint`.


daxpedda:
The current implementation is actually correct: https://www.rfc-editor.org/rfc/rfc9496#name-equals-2.

PR #1332, opened 2025-07-29
Use `ConstMontyForm::invert()` in place of `pow()`

daxpedda:
I didn't add this in #1302 because the performance was much worse. Now with the latest changes in crypto-bigint this isn't the case anymore. Instead the new `ConstMontyForm::invert()` improves the performance by 85% (on my PC).

### DIFF
--- a/ed448-goldilocks/src/field/element.rs
+++ b/ed448-goldilocks/src/field/element.rs
@@ -263,12 +263,8 @@ impl FieldElement {
     }
 
     /// Inverts a field element
-    /// Previous chain length: 462, new length 460
     pub fn invert(&self) -> Self {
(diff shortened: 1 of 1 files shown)


ed448-goldilocks/src/field/element.rs
```diff
@@ -265,10 +265,7 @@ impl FieldElement {
     /// Inverts a field element
     /// Previous chain length: 462, new length 460
```
tarcieri: I assume this comment was referring to an addition chain at one point but perhaps could be removed now?

daxpedda: Done!

PR #1333, opened 2025-07-29
Move function requiring `AffinePoint` from `EdwardsPoint` to `AffinePoint`

daxpedda:
A bunch of functionality implemented on `EdwardsPoint` internally just converts to `AffinePoint`. This functionality should rather live in `AffinePoint`.

This PR also implement `BatchNormalize` for `EdwardsPoint`, which wouldn't have much of a point without these changes to begin with.

### DIFF
--- a/ed448-goldilocks/README.md
+++ b/ed448-goldilocks/README.md
@@ -19,6 +19,7 @@ It is intended to be portable, fast, and safe.
 ```rust
 use ed448_goldilocks::{Ed448, EdwardsPoint, CompressedEdwardsY, EdwardsScalar, sha3::Shake256};
 use elliptic_curve::Field;
+use elliptic_curve::group::GroupEncoding;
 use hash2curve::{ExpandMsgXof, GroupDigest};
 use rand_core::OsRng;
 
@@ -29,9 +30,9 @@ assert_eq!(public_key, EdwardsPoint::GENERATOR + EdwardsPoint::GENERATOR);
 
 let secret_key = EdwardsScalar::try_from_rng(&mut OsRng).unwrap();
 let public_key = EdwardsPoint::GENERATOR * &secret_key;
-let compressed_public_key = public_key.compress();
+let compressed_public_key = public_key.to_bytes();
 
-assert_eq!(compressed_public_key.to_bytes().len(), 57);
+assert_eq!(compressed_public_key.len(), 57);
 
 let hashed_scalar = Ed448::hash_to_scalar::<ExpandMsgXof<Shake256>>(&[b"test"], &[b"edwards448_XOF:SHAKE256_ELL2_RO_"]).unwrap();
 let input = hex_literal::hex!("c8c6c8f584e0c25efdb6af5ad234583c56dedd7c33e0c893468e96740fa0cf7f1a560667da40b7bde340a39252e89262fcf707d1180fd43400");
@@ -40,7 +41,7 @@ assert_eq!(hashed_scalar, expected_scalar);
 
 let hashed_point = Ed448::hash_from_bytes::<ExpandMsgXof<Shake256>>(&[b"test"], &[b"edwards448_XOF:SHAKE256_ELL2_RO_"]).unwrap();
 let expected = hex_literal::hex!("d15c4427b5c5611a53593c2be611fd3635b90272d331c7e6721ad3735e95dd8b9821f8e4e27501ce01aa3c913114052dce2e91e8ca050f4980");
-let expected_point = CompressedEdwardsY(expected).decompress().unwrap();
+let expected_point = CompressedEdwardsY(expected).decompress().unwrap().to_edwards();
 assert_eq!(hashed_point, expected_point);
 
 let hashed_point = EdwardsPoint::hash_with_defaults(b"test");
--- a/ed448-goldilocks/src/edwards/affine.rs
+++ b/ed448-goldilocks/src/edwards/affine.rs
@@ -1,8 +1,10 @@
 use crate::field::FieldElement;
 use crate::*;
+use core::fmt::{Display, Formatter, LowerHex, Result as FmtResult, UpperHex};
 use core::ops::Mul;
-use elliptic_curve::{Error, Result, point::NonIdentity, zeroize::DefaultIsZeroes};
-use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
+use elliptic_curve::{Error, point::NonIdentity, zeroize::DefaultIsZeroes};
+use rand_core::TryRngCore;
+use subtle::{Choice, ConditionallyNegatable, ConditionallySelectable, ConstantTimeEq, CtOption};
 
 /// Affine point on untwisted curve
 #[derive(Copy, Clone, Debug)]
@@ -69,6 +71,21 @@ impl AffinePoint {
         y: FieldElement::ONE,
     };
 
+    /// Generate a random [`AffinePoint`].
+    pub fn try_from_rng<R>(rng: &mut R) -> Result<Self, R::Error>
+    where
+        R: TryRngCore + ?Sized,
+    {
+        let mut bytes = CompressedEdwardsY::default();
+
+        loop {
+            rng.try_fill_bytes(&mut bytes.0)?;
+            if let Some(point) = bytes.decompress().into() {
+                return Ok(point);
+            }
+        }
+    }
+
     pub(crate) fn isogeny(&self) -> Self {
         let x = self.x;
         let y = self.y;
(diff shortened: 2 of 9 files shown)


tarcieri:
This is definitely pushing my knowledge of both Edwards curves and Ed448-Goldilocks, and sent me on a rabbit hole reading the paper.

I think both versions of `edwards_isogeny` are naively implemented, and we would ideally be converting directly from `EdwardsPoint` to `TwistedExtendedPoint` (sidebar: which should probably be called `TwistedEdwardsPoint` for consistency, I guess the former was renamed from `ExtendedPoint` at some point) without first converting to an intermediate `AffinePoint` with its associated inversions. There's even a comment to that effect:

> So what we can do is optimise each respective isogeny method for a=1 or a = -1 (currently, I just made it really slow and simple)

Edit: opened an issue about this #1349, and it seems to be the reason you weren't seeing speedups using twisted scalar multiplication in #1337

tarcieri:
To expound on that:

- `EdwardsPoint` should probably remain the primary user-facing type with the majority of the functionality
- `AffinePoint` and `TwistedExtendedPoint` are implementation details

Conversions look like:

```
CompressedEdwardsY <-> AffinePoint <-> EdwardsPoint <-> TwistedExtendedPoint
```

...but I think it's fine to preserve methods like `compress` and `decompress` directly on `EdwardsPoint`, but with the actual point compression/decompression implementation factored onto `AffinePoint`, and the methods on `EdwardsPoint` delegating to `AffinePoint` and doing affine conversions as needed.

Likewise, scalar multiplication would ideally do a more direct conversion from `EdwardsPoint` to `TwistedExtendedPoint`, and delegate to the implementation there.

daxpedda:
`compress()` and `decompress()` remain available on `EdwardsPoint` through `GroupEncoding`. Eventually we should remove these functions entirely and implement `GroupEncoding` on `AffinePoint` as well (tracking that in #1348).

I hope I'm on the right track here? I'm kinda trying to align things with the other RustCrypto crates, where most types have very few methods and its all mostly implemented through traits. It also makes sense to me, because otherwise we have all those duplicate methods.

ed448-goldilocks/src/edwards/extended.rs
```diff
@@ -950,14 +654,14 @@ impl Mul<&EdwardsScalar> for &EdwardsPoint {
 
     /// Scalar multiplication: compute `scalar * self`.
     fn mul(self, scalar: &EdwardsScalar) -> EdwardsPoint {
-        self.scalar_mul(scalar)
+        self.to_affine().scalar_mul(scalar)
     }
```
tarcieri: This in particular seems problematic. Using the map from #1316 you can convert directly from the untwisted to twisted form without an intermediate affine conversion.

So scalar multiplication, and all of the functionality associated with it which has been removed in this PR, should remain here on `EdwardsPoint`, which IMO should be the main user-facing API the library presents (ala `curve25519-dalek`).

I think it's okay to have scalar multiplication on `AffinePoint`, but it should work by homogenizing into `EdwardsPoint` first, then performing scalar multiplication with that.

ed448-goldilocks/src/edwards/extended.rs
```diff
@@ -549,18 +328,6 @@ impl EdwardsPoint {
...
-    pub fn scalar_mul(&self, scalar: &EdwardsScalar) -> Self {
-        // Compute floor(s/4)
-        let mut scalar_div_four = *scalar;
-        scalar_div_four.div_by_four();
-
-        // Use isogeny and dual isogeny to compute phi^-1((s/4) * phi(P))
-        let partial_result = variable_base(&self.to_twisted(), &scalar_div_four).to_untwisted();
-        // Add partial result to (scalar mod 4) * P
-        partial_result.add(&self.scalar_mod_four(scalar))
-    }
```
tarcieri: This can remain here, and with #1316 the affine conversion will be avoided.

ed448-goldilocks/src/edwards/extended.rs
daxpedda: Yes indeed, now that we have the direct mapping I will rebase this PR so it makes more sense.

daxpedda:
Putting on draft until I rebase. Still working on those optimized implementations from #1316.

ed448-goldilocks/src/edwards/extended.rs
daxpedda: Done.

ed448-goldilocks/src/edwards/extended.rs
daxpedda: Done.

ed448-goldilocks/README.md
```diff
@@ -29,7 +29,7 @@ assert_eq!(public_key, EdwardsPoint::GENERATOR + EdwardsPoint::GENERATOR);
 
 let secret_key = EdwardsScalar::try_from_rng(&mut OsRng).unwrap();
 let public_key = EdwardsPoint::GENERATOR * &secret_key;
-let compressed_public_key = public_key.compress();
+let compressed_public_key = public_key.to_affine().compress();
```
tarcieri: Can you import the trait so these "just work" without the intermediate affine conversion?

ed448-goldilocks/src/edwards/affine.rs
```diff
@@ -100,6 +117,34 @@ impl AffinePoint {
         }
     }
 
+    /// Standard compression; store Y and sign of X
+    // XXX: This needs more docs and is `compress` the conventional function name? I think to_bytes/encode is?
```
tarcieri: `compress` seems fine to me for this

ed448-goldilocks/src/edwards/affine.rs
```diff
@@ -100,6 +117,34 @@ impl AffinePoint {
...
+    pub fn is_on_curve(&self) -> Choice {
+        // X^2 + Y^2 == 1 + D * X^2 * Y^2
+
+        let XX = self.x.square();
+        let YY = self.y.square();
+        let lhs = YY + XX;
+        let rhs = FieldElement::ONE + FieldElement::EDWARDS_D * XX * YY;
+
+        lhs.ct_eq(&rhs)
+    }
```
tarcieri, line 145: Is this really needed, especially as a `pub fn`? The alternative approach used by the rest of the crates in this repo is to never instantiate an `AffinePoint` without first checking it's on the curve, and since Curve448/Ed448 always uses compressed coordinates for serialization, you effectively get that for free

ed448-goldilocks/src/edwards/affine.rs
```diff
@@ -149,3 +194,234 @@ define_mul_variants!(
...
+}
+
+#[cfg(feature = "alloc")]
+impl TryFrom<Box<[u8]>> for CompressedEdwardsY {
+    type Error = &'static str;
+
+    fn try_from(value: Box<[u8]>) -> Result<Self, Self::Error> {
+        Self::try_from(value.as_ref())
+    }
+}
```
tarcieri, line 329: The impls for `Vec<u8>` and `Box<[u8]>` seem a little superfluous to me when there's one for byte slices

ed448-goldilocks/src/edwards/affine.rs
daxpedda: All the methods in this PR are just moved from `EdwardsPoint`, I didn't really touch them otherwise. My intention was to go over removing all these functions in a separate PR, tracking it in #1348.

But am happy to do it in this PR as well.

ed448-goldilocks/src/edwards/affine.rs
daxpedda: As above, just moved over from the `edwards/extended` module.
I'm happy to remove them in this PR.

tarcieri: Sure

ed448-goldilocks/src/edwards/affine.rs
daxpedda: This was copied over as well. Anyway, removed it.

ed448-goldilocks/README.md
daxpedda: Done.

ed448-goldilocks/src/sign/expanded.rs
```diff
@@ -61,7 +64,7 @@ impl ExpandedSecretKey {
 
         let point = EdwardsPoint::GENERATOR * scalar;
         let public_key = VerifyingKey {
-            compressed: point.compress(),
+            compressed: point.to_affine().compress(),
```
daxpedda, line 67: Couldn't remove this because we are storing `CompressedEdwardsY` in `VerifyingKey` and `GroupEncoding` doesn't return that.

I'm happy to change the type.
I'm planning on removing the type entirely anyway to align with the other curves in this repo, so we can wait for that as well.

tarcieri (approved):

PR #1336, opened 2025-07-30
Remove unnecessary checks from `CompressedEdwardsY::decompress()`

daxpedda:
Currently `CompressedEdwardsY::decompress()` checks if the point is on the curve. This PR removes this check.

Please correct me if I'm wrong here, but I believe this check can never fail because we don't allow invalid y-coordinates (we reduce the input) and the x-coordinate is derived.

### DIFF
--- a/ed448-goldilocks/src/edwards/affine.rs
+++ b/ed448-goldilocks/src/edwards/affine.rs
@@ -388,39 +388,37 @@ impl CompressedEdwardsY {
     ///
     /// Returns `None` if the input is not the \\(y\\)-coordinate of a
     /// curve point.
+    // See https://www.rfc-editor.org/rfc/rfc8032.html#section-5.2.3.
     pub fn decompress_unchecked(&self) -> CtOption<AffinePoint> {
         // Safe to unwrap here as the underlying data structure is a slice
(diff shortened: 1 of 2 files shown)


tarcieri:
> we don't allow invalid y-coordinates (we reduce the input)

That doesn't mean a prospective y-coordinate will be a valid solution to the curve equation.

However, as far as I can tell `CompressedEdwardsY::decompress_unchecked` does ensure a valid solution to the curve equation, and will return the `CtOption` equivalent of `None` if it isn't

daxpedda:
> > we don't allow invalid y-coordinates (we reduce the input)
> 
> That doesn't mean a prospective y-coordinate will be a valid solution to the curve equation.

I see, thanks!

> However, as far as I can tell `CompressedEdwardsY::decompress_unchecked` does ensure a valid solution to the curve equation, and will return the `CtOption` equivalent of `None` if it isn't

I went ahead and adjusted the implementation to follow RFC 8032 and left comments here and there. If we assume that the RFC algorithm only produces points on the curve, our current implementation must be fine then.

daxpedda:
I went ahead and removed the check for torsion-freeness as well, following the RFC.

PR #1334, opened 2025-07-29
ed448-goldilocks: add basic benchmark suite

daxpedda:
This adds a basic benchmark suite to `ed448-goldilocks` for various curves.

PR #1337, opened 2025-07-30
Use untwisted Edwards Curve for Ed448 scalar multiplication

daxpedda:
The current scalar multiplication algorithm for `EdwardsPoint` does the computation on the twisted Edwards curve. ~~However, the conversion is quite expensive.~~ (turns out with #1332 the conversion is significantly faster)

This PR moves the computation to happen on the untwisted Edwards curve:
```
Ed448/scalar multiplication
                        time:   [256.58 µs 256.76 µs 256.94 µs]
                        change: [−7.9084% −7.7969% −7.6979%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Based on #1303, #1313, #1314, #1329, #1330, #1332 and #1334 to make sure all optimizations are pulled in before comparing performance.

With further improvements to inversion, the difference could go away entirely, so we might want to hold off until we get a clearer picture from crypto-bigint.


daxpedda:
I think we should delete the whole `mod scalar_mul` and move its implementation to Decaf, the only place its going to be used after this PR.

The only alternative I see is to somehow make all this code generic to be able to share it between Decaf and Edwards, but it seems that will actually produce more code than it will save.

daxpedda:
I forgot to add https://github.com/RustCrypto/elliptic-curves/pull/1332, which makes the conversion to twisted form and back much faster.

So now we are only seeing a 8% improvement, not sure if its worth it. If further optimizations to inversion are incoming this all might not be worth it.

tarcieri:
I think instead of this we should use a better conversion to/from the twisted form which avoids inversions: #1349

daxpedda:
Yeah, with #1316 the performance of this PR is actually worse.

daxpedda:
Just wanted to report back: with all optimizations applied the difference between the twisted and untwisted curve is only one extra multiplication when doing additions.

In total this adds up to 120 extra multiplications in the untwisted curve.
The conversion back and forth adds another 8 multiplications and 4 squarings.

So in total doing the multiplication in the twisted curve saves 112 multiplications but adds 4 squarings.
Clearly an improvement.

PR #1338, opened 2025-07-31
`hash2curve`: Error rework

carloskiki:
Discussed in https://github.com/RustCrypto/elliptic-curves/issues/1295. This was planned as a follow up to https://github.com/RustCrypto/elliptic-curves/pull/1317, but I decided to make it as its own PR because #1317 is not unanimous.

Replaces `elliptic_curve::{Error, Result}` with an associated type on `ExpandMsg`.

### DIFF
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -1,8 +1,8 @@
 //! Traits for handling hash to curve.
 
 use super::{ExpandMsg, MapToCurve, hash_to_field};
-use elliptic_curve::array::typenum::Unsigned;
-use elliptic_curve::{ProjectivePoint, Result};
+use digest::array::typenum::Unsigned;
+use elliptic_curve::ProjectivePoint;
 
 /// Hash arbitrary byte sequences to a valid group element.
 pub trait GroupDigest: MapToCurve {
@@ -32,7 +32,7 @@ pub trait GroupDigest: MapToCurve {
     ///
     /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
     /// [`ExpandMsgXof`]: crate::ExpandMsgXof
-    fn hash_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>>
+    fn hash_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>, X::Error>
     where
         X: ExpandMsg<Self::K>,
     {
@@ -62,7 +62,7 @@ pub trait GroupDigest: MapToCurve {
     ///
     /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
     /// [`ExpandMsgXof`]: crate::ExpandMsgXof
-    fn encode_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>>
+    fn encode_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>, X::Error>
     where
         X: ExpandMsg<Self::K>,
     {
@@ -85,7 +85,7 @@ pub trait GroupDigest: MapToCurve {
     ///
     /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
     /// [`ExpandMsgXof`]: crate::ExpandMsgXof
-    fn hash_to_scalar<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<Self::Scalar>
+    fn hash_to_scalar<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<Self::Scalar, X::Error>
     where
         X: ExpandMsg<Self::K>,
     {
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -79,8 +79,11 @@ pub trait GroupDigest: MapToCurve {
     ///
     /// # Errors
     ///
-    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
-    /// and [`crate::ExpandMsgXof`] for examples.
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
+    /// and [`ExpandMsgXofError`] for examples.
+    ///
+    /// [`ExpandMsgXmdError`]: crate::ExpandMsgXmdError
+    /// [`ExpandMsgXofError`]: crate::ExpandMsgXofError
     fn hash_to_scalar<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<Self::Scalar, X::Error>
     where
         X: ExpandMsg<Self::K>,
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -56,8 +56,11 @@ pub trait GroupDigest: MapToCurve {
     ///
     /// # Errors
     ///
-    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
-    /// and [`crate::ExpandMsgXof`] for examples.
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
+    /// and [`ExpandMsgXofError`] for examples.
+    ///
+    /// [`ExpandMsgXmdError`]: crate::ExpandMsgXmdError
+    /// [`ExpandMsgXofError`]: crate::ExpandMsgXofError
     fn encode_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>, X::Error>
     where
         X: ExpandMsg<Self::K>,
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -23,15 +23,9 @@ pub trait GroupDigest: MapToCurve {
     /// > hash function is used.
     ///
     /// # Errors
-    /// - `len_in_bytes > u16::MAX`
-    /// - See implementors of [`ExpandMsg`] for additional errors:
-    ///   - [`ExpandMsgXmd`]
-    ///   - [`ExpandMsgXof`]
     ///
-    /// `len_in_bytes = <Self::FieldElement as FromOkm>::Length * 2`
-    ///
-    /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
-    /// [`ExpandMsgXof`]: crate::ExpandMsgXof
+    /// When the chosen `ExpandMsg` implementation returns an error. See [`crate::ExpandMsgXmd`]
+    /// and [`crate::ExpandMsgXof`] for examples.
     fn hash_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>, X::Error>
     where
         X: ExpandMsg<Self::K>,
@@ -53,15 +47,9 @@ pub trait GroupDigest: MapToCurve {
     /// > points in this set are more likely to be output than others.
     ///
     /// # Errors
-    /// - `len_in_bytes > u16::MAX`
-    /// - See implementors of [`ExpandMsg`] for additional errors:
-    ///   - [`ExpandMsgXmd`]
-    ///   - [`ExpandMsgXof`]
     ///
-    /// `len_in_bytes = <Self::FieldElement as FromOkm>::Length`
-    ///
-    /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
-    /// [`ExpandMsgXof`]: crate::ExpandMsgXof
+    /// When the chosen `ExpandMsg` implementation returns an error. See [`crate::ExpandMsgXmd`]
+    /// and [`crate::ExpandMsgXof`] for examples.
     fn encode_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>, X::Error>
     where
         X: ExpandMsg<Self::K>,
@@ -76,15 +64,9 @@ pub trait GroupDigest: MapToCurve {
     /// and returns a scalar.
     ///
     /// # Errors
-    /// - `len_in_bytes > u16::MAX`
-    /// - See implementors of [`ExpandMsg`] for additional errors:
-    ///   - [`ExpandMsgXmd`]
-    ///   - [`ExpandMsgXof`]
     ///
-    /// `len_in_bytes = <Self::Scalar as FromOkm>::Length`
-    ///
-    /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
-    /// [`ExpandMsgXof`]: crate::ExpandMsgXof
+    /// When the chosen `ExpandMsg` implementation returns an error. See [`crate::ExpandMsgXmd`]
+    /// and [`crate::ExpandMsgXof`] for examples.
     fn hash_to_scalar<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<Self::Scalar, X::Error>
     where
         X: ExpandMsg<Self::K>,
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -2,7 +2,7 @@
 
 use super::{ExpandMsg, MapToCurve, hash_to_field};
 use digest::array::typenum::Unsigned;
-use elliptic_curve::ProjectivePoint;
+use elliptic_curve::{ProjectivePoint, array::typenum::Unsigned};
 
 /// Hash arbitrary byte sequences to a valid group element.
 pub trait GroupDigest: MapToCurve {
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -27,7 +27,11 @@ pub trait GroupDigest: MapToCurve {
     /// # Errors
     ///
     /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
-    /// and [`crate::ExpandMsgXof`] for examples.
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
+    /// and [`ExpandMsgXofError`] for examples.
+    ///
+    /// [`ExpandMsgXmdError`]: crate::ExpandMsgXmdError
+    /// [`ExpandMsgXofError`]: crate::ExpandMsgXofError
     fn hash_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>, X::Error>
     where
         X: ExpandMsg<Self::K>,
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -1,7 +1,6 @@
 //! Traits for handling hash to curve.
 
 use super::{ExpandMsg, MapToCurve, hash_to_field};
-use digest::array::typenum::Unsigned;
 use elliptic_curve::{ProjectivePoint, array::typenum::Unsigned};
 
 /// Hash arbitrary byte sequences to a valid group element.
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -22,6 +22,8 @@ pub trait GroupDigest: MapToCurve {
     /// > oracle returning points in G assuming a cryptographically secure
     /// > hash function is used.
     ///
+    /// For the `expand_message` call, `len_in_bytes = <Self::FieldElement as FromOkm>::Length * 2`.
+    ///
     /// # Errors
     ///
     /// When the chosen `ExpandMsg` implementation returns an error. See [`crate::ExpandMsgXmd`]
@@ -45,6 +47,8 @@ pub trait GroupDigest: MapToCurve {
     /// > uniformly random in G: the set of possible outputs of
     /// > encode_to_curve is only a fraction of the points in G, and some
     /// > points in this set are more likely to be output than others.
+    /// 
+    /// For the `expand_message` call, `len_in_bytes = <Self::FieldElement as FromOkm>::Length`.
     ///
     /// # Errors
     ///
@@ -62,7 +66,9 @@ pub trait GroupDigest: MapToCurve {
     /// Computes the hash to field routine according to
     /// <https://www.rfc-editor.org/rfc/rfc9380.html#section-5-4>
     /// and returns a scalar.
-    ///
+    ///   
+    /// For the `expand_message` call, `len_in_bytes = <Self::FieldElement as FromOkm>::Length`.
+    /// 
     /// # Errors
     ///
     /// When the chosen `ExpandMsg` implementation returns an error. See [`crate::ExpandMsgXmd`]
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -22,6 +22,7 @@ pub trait GroupDigest: MapToCurve {
     /// > hash function is used.
     ///
     /// For the `expand_message` call, `len_in_bytes = <Self::FieldElement as FromOkm>::Length * 2`.
+    /// This value must be less than `u16::MAX` or otherwise a compiler error will occur.
     ///
     /// # Errors
     ///
@@ -48,6 +49,7 @@ pub trait GroupDigest: MapToCurve {
     /// > points in this set are more likely to be output than others.
     ///
     /// For the `expand_message` call, `len_in_bytes = <Self::FieldElement as FromOkm>::Length`.
+    /// This value must be less than `u16::MAX` or otherwise a compiler error will occur.
     ///
     /// # Errors
     ///
@@ -67,6 +69,7 @@ pub trait GroupDigest: MapToCurve {
     /// and returns a scalar.
     ///   
     /// For the `expand_message` call, `len_in_bytes = <Self::FieldElement as FromOkm>::Length`.
+    /// This value must be less than `u16::MAX` or otherwise a compiler error will occur.
     ///
     /// # Errors
     ///
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -26,7 +26,6 @@ pub trait GroupDigest: MapToCurve {
     ///
     /// # Errors
     ///
-    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
     /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
     /// and [`ExpandMsgXofError`] for examples.
     ///
--- a/hash2curve/src/group_digest.rs
+++ b/hash2curve/src/group_digest.rs
@@ -26,7 +26,7 @@ pub trait GroupDigest: MapToCurve {
     ///
     /// # Errors
     ///
-    /// When the chosen `ExpandMsg` implementation returns an error. See [`crate::ExpandMsgXmd`]
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
     /// and [`crate::ExpandMsgXof`] for examples.
     fn hash_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>, X::Error>
     where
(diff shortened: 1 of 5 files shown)


hash2curve/src/hash2field/expand_msg.rs
```diff
@@ -56,19 +61,15 @@ pub(crate) enum Domain<'a, L: ArraySize> {
...
         X: Default + ExtendableOutput + Update,
     {
         // https://www.rfc-editor.org/rfc/rfc9380.html#section-3.1-4.2
         if dst.iter().map(|slice| slice.len()).sum::<usize>() == 0 {
-            Err(Error)
+            Err(EmptyDST)
         } else if dst.iter().map(|slice| slice.len()).sum::<usize>() > MAX_DST_LEN {
-            if L::USIZE > u8::MAX.into() {
-                return Err(Error);
-            }
```
daxpedda: This should remain a run-time error. It should still be valid to use a non-oversized DST with a hash that has a big output size.

hash2curve/src/hash2field.rs
```diff
@@ -28,26 +27,23 @@ pub trait FromOkm {
...
     E: ExpandMsg<K>,
     T: FromOkm + Default,
 {
-    let len_in_bytes = T::Length::USIZE
-        .checked_mul(N)
-        .and_then(|len| len.try_into().ok())
-        .and_then(NonZeroU16::new)
-        .ok_or(Error)?;
+    // Completely degenerate case; `N` and `T::Length` would need to be extremely large.
+    const { assert!(T::Length::USIZE * N <= u16::MAX as usize) }
```
daxpedda: Wow, this is news to me! This actually fails during compile-time ... this is amazing 🤯! I'm also surprised `T::Length` is accessible in a const block!

We should include a proper error message in this `assert!`.
`checked_mul()` should still be used here to protect against overflows (indeed very degenerate case).

hash2curve/src/hash2field.rs
```diff
@@ -28,26 +27,23 @@ pub trait FromOkm {
...
+    let Some(len_in_bytes) = NonZeroU16::new(T::Length::U16 * N as u16) else {
+        // Since `T::Length: NonZero`, only `N = 0` can lead to this case.
+        return Ok(core::array::from_fn(|_| T::default()));
```
daxpedda: Well, we can just turn this into a compile-time error as well.

hash2curve/src/group_digest.rs
```diff
@@ -23,16 +23,10 @@ pub trait GroupDigest: MapToCurve {
     /// > hash function is used.
     ///
     /// # Errors
-    /// - `len_in_bytes > u16::MAX`
-    /// - See implementors of [`ExpandMsg`] for additional errors:
-    ///   - [`ExpandMsgXmd`]
-    ///   - [`ExpandMsgXof`]
     ///
-    /// `len_in_bytes = <Self::FieldElement as FromOkm>::Length * 2`
```
daxpedda, line 31: This is necessary so users can figure out the `len_in_bytes` that triggered the error.

hash2curve/src/hash2field/expand_msg.rs
```diff
@@ -153,3 +150,15 @@ impl<'a, L: ArraySize> Domain<'a, L> {
         assert_eq!(self.len(), bytes[bytes.len() - 1]);
     }
 }
+
+/// Error when an empty domain separation tag is used.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct EmptyDST;
```
daxpedda:
```suggestion
pub struct EmptyDst;
```

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -146,12 +147,40 @@ where
     }
 }
 
+/// Error type for [`ExpandMsgXmd`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ExpandMsgXmdError {
+    /// The domain separation tag is empty.
+    DSTError(EmptyDST),
```
daxpedda: We shouldn't use the term "error" for an error variant.
```suggestion
    Dst(EmptyDST),
```
I'm also not sure if its really sensible to have a wrapped error type like this instead of just a variant called `EmptyDst`.

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -146,12 +147,40 @@ where
...
+    /// The length in bytes is too large.
+    ///
+    /// `len_in_bytes` must be at most `255 * HashT::OutputSize`
+    OversizedLen,
```
daxpedda: We should avoid using the term "oversized" here, its a bit inflated in hash2curve with "oversized DST".
```suggestion
    Length,
```
This seems enough to me, but `TooLong` is also an option. Its bikeshedding at this point.

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -146,12 +147,40 @@ where
...
+}
+
+impl core::fmt::Display for ExpandMsgXmdError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::DSTError(e) => write!(f, "{e}"),
+            Self::OversizedLen => write!(f, "length in bytes is too large"),
```
daxpedda:
```suggestion
            Self::OversizedLen => write!(f, "`len_in_bytes` is too large"),
```
Lets keep consistent so users can figure out what they need here.

hash2curve/src/hash2field/expand_msg/xof.rs
```diff
@@ -1,22 +1,23 @@
...
 };
 
 /// Implements `expand_message_xof` via the [`ExpandMsg`] trait:
 /// <https://www.rfc-editor.org/rfc/rfc9380.html#name-expand_message_xof>
 ///
 /// # Errors
-/// - `dst` contains no bytes
-/// - `dst > 255 && K * 2 > 255`
+///
+/// `expand_message` caa return an [`EmptyDST`] error.
```
daxpedda:
```suggestion
/// `expand_message` can return an [`EmptyDST`] error.
```

daxpedda (changes requested):

hash2curve/src/hash2field.rs
carloskiki: I think it is fine to use it here because it is a case we should not ever see, but we need to be careful to not over use it because this type of compile assertion causes non-local errors: https://hackmd.io/dTZWM-kVTlquPXMJ3KKsVw.

I think most of the issues related to this are fixed, but previously it was sometimes the case that `cargo check` passed (because it does not monomorphise the functions), but `cargo build` failed (once monomorphised over a specific constant, there is a compile error).

carloskiki: addressed in https://github.com/RustCrypto/elliptic-curves/pull/1338/commits/387cf666cb24ba67152bf581f42b9545398d7527.

hash2curve/src/hash2field.rs
carloskiki: Functions in `std` seem to generally accept `N = 0` and return a zero length array, e.g.: https://doc.rust-lang.org/stable/std/primitive.slice.html#method.first_chunk.

Nevertheless, I made the change in https://github.com/RustCrypto/elliptic-curves/pull/1338/commits/387cf666cb24ba67152bf581f42b9545398d7527.

hash2curve/src/hash2field/expand_msg.rs
carloskiki: Oooh, makes total sense.

hash2curve/src/hash2field/expand_msg.rs
carloskiki: fixed in https://github.com/RustCrypto/elliptic-curves/pull/1338/commits/a659f8bc6c9f2579909b8854d53d5f4e282ed001.

hash2curve/src/hash2field/expand_msg/xmd.rs
carloskiki: Fixed in https://github.com/RustCrypto/elliptic-curves/pull/1338/commits/59b7896492da2f48f6751e729da56bb32f177f1b.

hash2curve/src/hash2field/expand_msg/xmd.rs
carloskiki: Fixed in https://github.com/RustCrypto/elliptic-curves/pull/1338/commits/5304eb19c9a8e92fcc8fd06a9a1531e5ea1c7215.

hash2curve/src/group_digest.rs
carloskiki: Fixed in https://github.com/RustCrypto/elliptic-curves/pull/1338/commits/4e5226771b148b4f9d70dd0cd8fa249056a726b6.

hash2curve/src/hash2field/expand_msg.rs
carloskiki: Addressed in https://github.com/RustCrypto/elliptic-curves/pull/1338/commits/0ea02c8be7947fc22565b20f888785d1c793ecd5. I added different errors for Xmd and Xof, let me know if you were thinking of something else.

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -11,15 +11,13 @@ use digest::{
...
 
 /// Implements `expand_message_xof` via the [`ExpandMsg`] trait:
 /// <https://www.rfc-editor.org/rfc/rfc9380.html#name-expand_message_xmd>
 ///
 /// # Errors
-/// - `dst` contains no bytes
-/// - `dst > 255 && HashT::OutputSize > 255`
-/// - `len_in_bytes > 255 * HashT::OutputSize`
+///
+/// `expand_message` can return an [`ExpandMsgXmdError`].
```
daxpedda: If we don't list out error cases here we should make referral explicit.
```suggestion
///
/// See [`ExpandMsgXmdError`] for more details.
```

hash2curve/src/hash2field/expand_msg.rs
```diff
@@ -5,9 +5,10 @@ pub(super) mod xof;
 
 use core::num::NonZero;
 
-use digest::{Digest, ExtendableOutput, Update, XofReader};
-use elliptic_curve::array::{Array, ArraySize};
-use elliptic_curve::{Error, Result};
+use digest::{
+    Digest, ExtendableOutput, Update, XofReader,
+    array::{Array, ArraySize},
+};
```
daxpedda:
```suggestion
use digest::{Digest, ExtendableOutput, Update, XofReader};
use elliptic_curve::array::{Array, ArraySize};
```
Or why the change here?

hash2curve/src/hash2field/expand_msg.rs
```diff
@@ -153,3 +154,33 @@ impl<'a, L: ArraySize> Domain<'a, L> {
...
+pub enum DstError {
+    /// Empty domain separation tag.
+    Empty,
+    /// The domain separation tag is too long and needs to be hashed, but the hash function
+    /// selected has an output size too large (greater than `255`).
+    XmdHash,
+    /// The domain separation tag is too long and needs to be hashed, but the chosen `K` (target
+    /// security level in bytes) is too large (greater than `127`),
+    XofSecurityLevel,
+}
```
daxpedda: We should fold these error variants into `ExpandMsgXmdError` and a dedicated `ExpandMsgXofError`. We have dedicated methods on `Domain` anyway to handle separate error types.

That way users don't have to juggle with nested error types and we can also clearly separate XMD and XOF variants.

hash2curve/src/group_digest.rs
```diff
@@ -22,17 +22,13 @@ pub trait GroupDigest: MapToCurve {
     /// > oracle returning points in G assuming a cryptographically secure
     /// > hash function is used.
     ///
-    /// # Errors
-    /// - `len_in_bytes > u16::MAX`
```
daxpedda, line 26: I think we should keep the documentation of these compile-time errors. After reading about [non-local errors](https://hackmd.io/dTZWM-kVTlquPXMJ3KKsVw) it makes sense to me that we need to make these errors more visible to the user.

hash2curve/src/group_digest.rs
```diff
@@ -22,17 +22,13 @@ pub trait GroupDigest: MapToCurve {
...
-    /// - See implementors of [`ExpandMsg`] for additional errors:
-    ///   - [`ExpandMsgXmd`]
-    ///   - [`ExpandMsgXof`]
+    /// For the `expand_message` call, `len_in_bytes = <Self::FieldElement as FromOkm>::Length * 2`.
```
daxpedda, line 24: This might be a bit misleading because the error can also come from `hash_to_field()`. Related to https://github.com/RustCrypto/elliptic-curves/pull/1338/files#r2249369763.

hash2curve/src/group_digest.rs
```diff
@@ -22,17 +22,13 @@ pub trait GroupDigest: MapToCurve {
...
     ///
-    /// `len_in_bytes = <Self::FieldElement as FromOkm>::Length * 2`
+    /// # Errors
     ///
-    /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
-    /// [`ExpandMsgXof`]: crate::ExpandMsgXof
```
daxpedda, line 34: Whats wrong with the links? Makes it easier to read the documentation.

hash2curve/src/hash2field.rs
```diff
@@ -8,11 +8,10 @@ use core::num::NonZeroU16;
 
 pub use expand_msg::{xmd::*, xof::*, *};
 
-use elliptic_curve::array::{
+use digest::array::{
```
daxpedda: Why the change?

hash2curve/src/hash2field.rs
```diff
@@ -28,26 +27,26 @@ pub trait FromOkm {
...
-        .ok_or(Error)?;
+    // Completely degenerate case; `N` and `T::Length` would need to be extremely large.
+    const {
+        assert!(
+            T::Length::USIZE.saturating_mul(N) <= u16::MAX as usize,
+            "The product of `T::Length` and `N` must not exceed `u16::MAX`."
+        );
+        assert!(N > 0, "N must be greater than 0.");
+    }
+    let len_in_bytes = NonZeroU16::new(T::Length::U16 * N as u16).expect("N is greater than 0");
```
daxpedda:
```suggestion
    let len_in_bytes = const {
        assert!(
            T::Length::USIZE.saturating_mul(N) <= u16::MAX as usize,
            "The product of `T::Length` and `N` must not exceed `u16::MAX`."
        );
        NonZeroU16::new(T::Length::U16 * N as u16).expect("N is greater than 0")
    };
```

hash2curve/src/hash2field/expand_msg.rs
carloskiki: These are the same type re-exported by different crates. While getting rid of `elliptic_curve::{Error, Result}` I must have removed the whole use statement and re-imported using my editor, which made these types come from `digest` instead. I don't think this is a big deal at all even though `fmt` adds two more lines of code.

hash2curve/src/hash2field/expand_msg.rs
carloskiki: Good idea!

hash2curve/src/group_digest.rs
carloskiki: Do you mean when reading the docs in your editor or when reading them on `docs.rs`?
In docs.rs, the links still there: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html. You can run `cargo doc --open` to verify.

If you mean when reading the docs through you editor, `rust-analyzer` has "go to definition" capabilities for docs as well. So you can "go to definition" with any item enclosed in ``[`...`]``

hash2curve/src/group_digest.rs
carloskiki: The reason I used const assert there is that I was pretty certain this situation would never happen. Even if this happens, the error message is still somewhat helpful because of the text added to the assert.

But then, I tried to reproduce failing the const assert through `GroupDigest` to see what kind of error it would give. I realized `hybrid_array` [does not even implement `ArraySize` for types that big](https://docs.rs/hybrid-array/latest/hybrid_array/sizes/index.html), even with the `extra-sizes` feature. So one would need to implement `ArraySize` on a local type with a larger size (quite far fetched). Turns out, you can't even do that(!), since `ArraySize` is bounded by `typenum::Unsigned`, and `typenum::Usigned` is sealed.

hash2curve/src/hash2field.rs
carloskiki: Same as for https://github.com/RustCrypto/elliptic-curves/pull/1338#discussion_r2249930523.

hash2curve/src/group_digest.rs
carloskiki: In other words, to fail this we need to directly provide a relatively large `N` to `hash_to_field`, which then is no longer a non-local error (because it happens at the call site of `hash_to_field`, not `hash_to_bytes`).

hash2curve/src/hash2field/expand_msg.rs
daxpedda: Well, for consistency across the crate/repo/organization, we should stick with importing from `elliptic_curve`.

On a more general note, my experience so far is that its considered good practice to keep the diff as small as possible to make it easier to review. For smaller changes we can always make separate PRs and discuss it there instead of trying to coalesce unrelated changes into a single PR.

hash2curve/src/group_digest.rs
daxpedda: Not really following. If `FromOkm::Length` has been set to a very big size, its already possible to trigger that error. E.g. [`U35664 * 2 > u16::MAX`](https://docs.rs/hybrid-array/0.3.1/hybrid_array/sizes/type.U35664.html).

hash2curve/src/group_digest.rs
daxpedda: No, I just meant `ExpandMsgXmd` is easier to read then `crate::ExpandMsgXmd`, this is why these links were there in the first place.

hash2curve/src/group_digest.rs
carloskiki: Oh okay I did not see that type (the `hybrid_array` docs is misleading). I will change that.

hash2curve/src/hash2field/expand_msg.rs
carloskiki: This is now implemented.

hash2curve/src/group_digest.rs
carloskiki: This was added back in https://github.com/RustCrypto/elliptic-curves/pull/1338/commits/0d5fb7c43b8387e703225cb920df35ac14c08a43.

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -146,6 +145,35 @@ where
     }
 }
 
+/// Error type for [`ExpandMsgXmd`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ExpandMsgXmdError {
+    /// The domain separation tag is invalid because it is empty.
```
daxpedda:
```suggestion
    /// The domain separation tag must not be empty.
```

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -146,6 +145,35 @@ where
...
+    EmptyDst,
+    /// The domain separation tag is too long and needs to be hashed, but the hash function
+    /// selected has an output size too large (greater than `255`).
```
daxpedda: This error message requires too much knowledge about `expand_message_xmd()`.
```suggestion
    /// The hash's output size must not be greater then `255`
    /// if the domain separation tag is longer than `255`.
```

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -146,6 +145,35 @@ where
...
+    DstHash,
+    /// The length in bytes is too large.
+    ///
+    /// `len_in_bytes` must be at most `255 * HashT::OutputSize`
```
daxpedda:
```suggestion
    /// `len_in_bytes` must be at most `255 * HashT::OutputSize`.
```

hash2curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -146,6 +145,35 @@ where
...
+    Length,
+}
+
+impl core::fmt::Display for ExpandMsgXmdError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::EmptyDst => write!(f, "the domain separation tag is empty"),
+            Self::DstHash => write!(
+                f,
+                "`dst` is too long and the hash function has an output size greater than 255"
```
daxpedda:
```suggestion
                "hash output size is too large"
```
or
```suggestion
                "hash output size is too large with a domain separation tag this size/larger than 255"
```

hash2curve/src/hash2field/expand_msg/xof.rs
```diff
@@ -79,10 +77,32 @@ where
     }
 }
 
+/// Error type for [`ExpandMsgXof`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ExpandMsgXofError {
+    /// The domain separation tag is invalid because it is empty.
+    EmptyDst,
+    /// The domain separation tag is too long and needs to be hashed, but the selected
+    /// target security level in bytes (`K`) is too large (greater than `127`).
```
daxpedda:
```suggestion
    /// The target security level (`K`) must not be greater then `127`
    /// if the domain separation tag is longer than `255`.
```

hash2curve/src/hash2field/expand_msg/xof.rs
```diff
@@ -79,10 +77,32 @@ where
...
+    DstSecurityLevel,
+}
+
+impl core::fmt::Display for ExpandMsgXofError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::EmptyDst => write!(f, "the domain separation tag is empty"),
+            Self::DstSecurityLevel => write!(
+                f,
+                "the domain separation tag is too long and the target security level is too large"
```
daxpedda:
```suggestion
                "target security level is too large"
```
or
```suggestion
                "target security level is too large with a domain separation tag this size/larger than 255"
```

hash2curve/src/hash2field/expand_msg.rs
```diff
@@ -57,18 +58,17 @@ pub(crate) enum Domain<'a, L: ArraySize> {
 }
 
 impl<'a, L: ArraySize> Domain<'a, L> {
-    pub fn xof<X>(dst: &'a [&'a [u8]]) -> Result<Self>
+    pub fn xof<X>(dst: &'a [&'a [u8]]) -> Result<Self, xof::ExpandMsgXofError>
```
daxpedda: I think we should import at this point (using it 3 times).

hash2curve/src/group_digest.rs
```diff
@@ -22,17 +21,14 @@ pub trait GroupDigest: MapToCurve {
...
+    /// This value must be less than `u16::MAX` or otherwise a compiler error will occur.
     ///
-    /// `len_in_bytes = <Self::FieldElement as FromOkm>::Length * 2`
+    /// # Errors
     ///
-    /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
-    /// [`ExpandMsgXof`]: crate::ExpandMsgXof
-    fn hash_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>>
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
+    /// and [`crate::ExpandMsgXof`] for examples.
```
daxpedda:
```suggestion
    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
    /// and [`ExpandMsgXofError`] for examples.
    ///
    /// [`ExpandMsgXmdError`]: crate::ExpandMsgXmdError
    /// [`ExpandMsgXofError`]: crate::ExpandMsgXofError
```

hash2curve/src/hash2field.rs
```diff
@@ -28,26 +27,25 @@ pub trait FromOkm {
 /// <https://www.rfc-editor.org/rfc/rfc9380.html#name-hash_to_field-implementatio>
 ///
 /// # Errors
-/// - `len_in_bytes > u16::MAX`
-/// - See implementors of [`ExpandMsg`] for additional errors:
-///   - [`ExpandMsgXmd`]
-///   - [`ExpandMsgXof`]
 ///
-/// `len_in_bytes = T::Length * out.len()`
```
daxpedda: This part is still missing. But it doesn't really matter because this isn't really a public function.

daxpedda (changes requested):
Just documentation nits at this point.
Otherwise LGTM!

hash2curve/src/hash2field.rs
carloskiki: I also think that the types now serve as the documentation. I'll at least add that `len_in_bytes = T::Length * out.len()`.

hash2curve/src/hash2field/expand_msg.rs
carloskiki: Changed.

This makes me want to share the way I usually write code in personal projects. I would do: `expand_msg::xof::Error` instead of `ExpandMsgXofError`. Types inside the modules have generic names (like "Error"), but they clearly define what they do in the scope of the module. Using the module path feels more natural and you have to import less stuff. Types would also not be re-exported at the root of the library.

daxpedda: This is how I did it in the past as well. Unfortunately I found out that its not that popular because people will still try to import and now have to rename types manually. I guess it just goes against the current Rust culture.

I still do it this way in private projects though.

hash2curve/src/group_digest.rs
```diff
@@ -22,17 +21,18 @@ pub trait GroupDigest: MapToCurve {
...
     /// # Errors
-    /// - `len_in_bytes > u16::MAX`
-    /// - See implementors of [`ExpandMsg`] for additional errors:
-    ///   - [`ExpandMsgXmd`]
-    ///   - [`ExpandMsgXof`]
     ///
-    /// `len_in_bytes = <Self::FieldElement as FromOkm>::Length * 2`
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
+    /// and [`ExpandMsgXofError`] for examples.
```
daxpedda:
```suggestion
    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
    /// and [`ExpandMsgXofError`] for examples.
```

hash2curve/src/group_digest.rs
```diff
@@ -52,17 +52,14 @@ pub trait GroupDigest: MapToCurve {
...
+    /// This value must be less than `u16::MAX` or otherwise a compiler error will occur.
     ///
-    /// `len_in_bytes = <Self::FieldElement as FromOkm>::Length`
+    /// # Errors
     ///
-    /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
-    /// [`ExpandMsgXof`]: crate::ExpandMsgXof
-    fn encode_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>>
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
+    /// and [`crate::ExpandMsgXof`] for examples.
```
daxpedda:
```suggestion
    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
    /// and [`ExpandMsgXofError`] for examples.
    ///
    /// [`ExpandMsgXmdError`]: crate::ExpandMsgXmdError
    /// [`ExpandMsgXofError`]: crate::ExpandMsgXofError
```

hash2curve/src/group_digest.rs
```diff
@@ -74,18 +71,15 @@ pub trait GroupDigest: MapToCurve {
...
-    ///   - [`ExpandMsgXmd`]
-    ///   - [`ExpandMsgXof`]
-    ///
-    /// `len_in_bytes = <Self::Scalar as FromOkm>::Length`
     ///
-    /// [`ExpandMsgXmd`]: crate::ExpandMsgXmd
-    /// [`ExpandMsgXof`]: crate::ExpandMsgXof
-    fn hash_to_scalar<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<Self::Scalar>
+    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`crate::ExpandMsgXmd`]
+    /// and [`crate::ExpandMsgXof`] for examples.
```
daxpedda:
```suggestion
    /// When the chosen [`ExpandMsg`] implementation returns an error. See [`ExpandMsgXmdError`]
    /// and [`ExpandMsgXofError`] for examples.
    ///
    /// [`ExpandMsgXmdError`]: crate::ExpandMsgXmdError
    /// [`ExpandMsgXofError`]: crate::ExpandMsgXofError
```

daxpedda (changes requested):
LGTM!

I don't know what RustCrypto's stance is on custom error types is in general, but I think this is a good idea. Otherwise we should at least extract the compile-time error cases as I believe those are a great improvement.

tarcieri (approved):

Issue #1042, opened 2024-04-25
p256::Scalar: implement Reduce<U512>


survived:
I noticed that `p256::Scalar` does not implement `Reduce<U512>` trait. Other crates do have that: for instance, `k256::Scalar` does implement this trait. It makes it difficult to write generic code that uses this trait. Is there a reason for not implementing `Reduce<U512>` for `p256::Scalar`?

tarcieri:
It's a missing feature: `p256` lacks support for wide reductions.

`crypto-bigint` now contains a generic implementation of modular arithmetic which should make it somewhat straightforward to implement.

PR #1341, opened 2025-08-01
feat: support U384 in scalar arithmetic

### LINKED ISSUE #1199: `ReduceNonZero` should require at least minimum recommended bits
If `ReduceNonZero` implementation are supposed to follow ["ECDSA Key Pair Generation using Extra Random Bits"](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf#[{%22num%22%3A154%2C%22gen%22%3A0}%2C{%22name%22%3A%22XYZ%22}%2C70%2C586%2C0]), then we should make sure we follow the recommended minimum size.

This seems to only be an issue for P-256 (and K-256?), which according to FIPS requires at least 288 bits and recommended are 352 bits. Currently we only reduce from 256 bits (and 512 bits in the case of K-256).

CreeptoGengar:
### Description

Implement Reduce and ReduceNonZero traits for U384 in Scalar, update tests to cover new cases.

Fixed #1199

### DIFF
--- a/p256/src/arithmetic/scalar.rs
+++ b/p256/src/arithmetic/scalar.rs
@@ -645,13 +645,7 @@ impl Reduce<U512> for Scalar {
 
 impl Reduce<U384> for Scalar {
     fn reduce(w: &U384) -> Self {
-        // Convert U384 to U512 by zero-padding the high bits
-        let w_bytes = w.to_be_bytes();
-        let mut w512_bytes = [0u8; 64];
-        // Copy U384 bytes to the lower part of U512 (384 bits = 48 bytes)
-        w512_bytes[16..64].copy_from_slice(&w_bytes);
-        let w512 = U512::from_be_byte_array(w512_bytes.into());
-        <Self as Reduce<U512>>::reduce(&w512)
+        <Self as Reduce<U512>>::reduce(&w.concat(&U128::ZERO))
     }
 }
 


daxpedda (changes requested):
This isn't quite right, it should be `U352` according to [NIST FIPS 186-5](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf#%5B%7B%22num%22%3A154%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C70%2C586%2C0%5D)

tarcieri:
The integer size needs to be a multiple of the limb size. 352-bits is 32-bits * 11, however it isn't evenly divisible by 64-bits

CreeptoGengar:
@tarcieri @daxpedda 

Right, so U352 per FIPS but i  checked the crypto-bigint source - no U352 available, only U192, U224, U256, U384, U448, U512, etc. U384 is the closestr we've got and it actually divides evenly into both 32-bit (12 limbs) and 64-bit (6 limbs). 

Should U384 be fine for us or am i missing something?

daxpedda:
It is fine, was my bad.

Please refrain from explicit pinging for just a regular answer. Thanks.

p256/src/arithmetic/scalar.rs
```diff
@@ -634,6 +634,31 @@ impl Reduce<U256> for Scalar {
     }
 }
 
+impl Reduce<U384> for Scalar {
+    type Bytes = [u8; 48];
+
+    fn reduce(w: U384) -> Self {
+        // Convert U384 to U512 for use with barrett_reduce, which expects 512 bits total
```
tarcieri: Perhaps this should be composed in terms of a `Reduce<U512>` impl.

If you do that, you can also use `u384.concat(&U128::ZERO)` to obtain a `U512`.

daxpedda: Would address #1042 as well.

tarcieri:
One complication here is I'd like to attempt to migrate to a generic field element implementation (#1311) which wouldn't support bespoke reduce impls like this, and would only support generic zero and nonzero reductions for a modulo-sized integer and a double sized "wide" integer.

I can potentially merge this for now, I'm just saying I can't guarantee it will stick around, and it would be up to the caller to concat zeros to make a U512, then reduce.

CreeptoGengar:
> One complication here is I'd like to attempt to migrate to a generic field element implementation (#1311) which wouldn't support bespoke reduce impls like this, and would only support generic zero and nonzero reductions for a modulo-sized integer and a double sized "wide" integer.
> 
> I can potentially merge this for now, I'm just saying I can't guarantee it will stick around, and it would be up to the caller to concat zeros to make a U512, then reduce.

Makes sense! This won't fit the generic architecture. U384 -> U512 conversion is cleaner.

I'm cool with merging this as-is. When the generic implementation lands, users can just do the zero-padding and use standard reduction I guess

p256/src/arithmetic/scalar.rs
```diff
@@ -635,6 +635,37 @@ impl Reduce<FieldBytes> for Scalar {
...
+        let mut lo_bytes = [0u8; 32];
+        let mut hi_bytes = [0u8; 32];
+
+        // Copy the lower 256 bits (bytes 32-63)
+        lo_bytes.copy_from_slice(&w_bytes[32..64]);
+        // Copy the upper 256 bits (bytes 0-31)
+        hi_bytes.copy_from_slice(&w_bytes[0..32]);
+
+        let lo = U256::from_be_byte_array(lo_bytes.into());
+        let hi = U256::from_be_byte_array(hi_bytes.into());
```
tarcieri:
```suggestion
        let (lo, hi) = w.split();
```

p256/src/arithmetic/scalar.rs
```diff
@@ -635,6 +635,37 @@ impl Reduce<FieldBytes> for Scalar {
...
+
+impl Reduce<U384> for Scalar {
+    fn reduce(w: &U384) -> Self {
+        // Convert U384 to U512 by zero-padding the high bits
+        let w_bytes = w.to_be_bytes();
+        let mut w512_bytes = [0u8; 64];
+        // Copy U384 bytes to the lower part of U512 (384 bits = 48 bytes)
+        w512_bytes[16..64].copy_from_slice(&w_bytes);
+        let w512 = U512::from_be_byte_array(w512_bytes.into());
+        <Self as Reduce<U512>>::reduce(&w512)
```
tarcieri:
```suggestion
        <Self as Reduce<U512>>::reduce(&w.concat(&U128::ZERO))
```

PR #1342, opened 2025-08-02
fix(k256): correct scalar reduction 

crStiv:
Fix incorrect scalar modular reduction on 64-bit platforms.

Lines 153, 158, and 161 were using `sumadd()` instead of `muladd()` with `NEG_MODULUS[2]`, which breaks the Montgomery reduction algorithm. It could cause wrong results in scalar operations on 64-bit builds.


tarcieri:
I've verified the changes match other implementations, so it does seem like there is an issue here.

I'm curious why none of our existing test coverage caught it, as well as if we can add a regression test which captures the failure (provided we can find a suitable test case)

fjarri:
`libsecp256k1` matches the current implementation: https://github.com/bitcoin-core/secp256k1/blob/master/src/scalar_4x64_impl.h#L610. I am also very curious how this bug survived multiple years of usage, a security audit, and wycheproof vectors. Definitely need some regression tests showing the incorrect behavior.

fjarri:
Hm, I actually think this change does not affect the results at all. `NEG_MODULUS[2] == 1`, so `sumadd(n0, c0, c1, c2)` is equivalent to `muladd(n0, NEG_MODULUS[2], c0, c1, c2)`

tarcieri:
That's good to hear. Perhaps it's an optimization to use an add instead of a multiply due to that special case?

tarcieri:
Given that, I think we can close this

fjarri:
I am still interested to hear what incorrect results did @crStiv see.

crStiv:
@tarcieri @fjarri tnx guys for your reviews!
I found this bug by comparing the 64-bit implementation with the correct 32-bit version, breaking the Montgomery reduction pattern.

It could cause wrong scalar multiplication results and invalid signatures on 64-bit platforms.

tarcieri:
@crStiv can you provide a concrete example illustrating a miscomputation?

crStiv:
I can't provide a concrete example without implementing a test case, but it is clear from the code:

In wide32.rs :
```rust
let (c0, c1, c2) = muladd(n0, NEG_MODULUS[2], c0, c1, c2);
```

In wide64.rs :
```rust
let (c0, c1, c2) = sumadd(n0, c0, c1, c2);
```

As I understood correctly the Montgomery reduction algorithm requires multiplying by NEG_MODULUS, not just adding.

tarcieri:
@crStiv did you see @fjarri's comment above? https://github.com/RustCrypto/elliptic-curves/pull/1342#issuecomment-3146580443

Due to the particular nature of the modulus (i.e. since `NEG_MODULUS[2] == 1`) the two are equivalent, and doing an add is faster than a multiply.

PR #1343, opened 2025-08-02
`primeorder`: add optimized implementation for `LinearCombination`

daxpedda:
This PR adds a optimized implementation of `LinearCombination` for `primeorder`.

The code is shared with point-scalar multiplication. The initial implementation left building the pre-computed lookup table and scalar recoding entirely to the shared code, but unfortunately this showed some performance regression (around 2%). So I moved building the table and scalar recoding to the caller. See https://github.com/RustCrypto/elliptic-curves/compare/c8d63098ee2474ce1808a65e2efff58506f89cb0...423fc8db7755b9f56a0e0a59fe71e6b8a508a549.

### DIFF
--- a/primeorder/src/projective.rs
+++ b/primeorder/src/projective.rs
@@ -4,6 +4,7 @@
 
 use crate::{AffinePoint, Field, PrimeCurveParams, point_arithmetic::PointArithmetic};
 use core::{
+    array,
     borrow::Borrow,
     iter::Sum,
     ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

tarcieri:
Hmm, think I need to revert this, it ~~isn't compatible with the upstream changes on `master`~~ broke the build, e.g.

https://github.com/RustCrypto/elliptic-curves/actions/runs/16733154177/job/47365846705

I'm really confused what's happening here and having trouble reproducing it locally, even building for i686-unknown-linux-gnu.

Oh huh, is it an MSRV thing? It works on stable but not 1.85.

tarcieri:
Reproduced it locally with a 32-bit build and Rust 1.85 and I'm really confused what's happening there, or why it magically works with stable.

`lincomb` has an argument `ks: &mut [ByteArray<C::Uint>]` but it seems like the compiler thinks it's `&mut [FieldBytes<C>]` for some reason? It makes no sense.

I can even add explicit type annotations to the `array::from_fn` arguments and that doesn't help.

tarcieri:
Opened revert PR: #1356

daxpedda:
Fun, thanks for the ping.
Will look into it tomorrow.

PR #1344, opened 2025-08-02
Use rejection sampling for random point generation

### LINKED ISSUE #1140: Random point has known discrete log (under some conditions)
This one is subtle.

By taking a look at the random point implementation:
```rust
    fn random(mut rng: impl RngCore) -> Self {
        Self::GENERATOR * Scalar::random(&mut rng)
    }
```

This is technically correct (it is uniform).
One potential problem I have with this that knowing the state of PRNG the caller would know the discrete log of the point and this is also technically fine, but taking into account the context and how this crate can be used it is easy to make a mistake to use this function to create a point nobody is supposed to know dlog of. Imagine the case where some PRNG is seeded with transcript state (I know, far stretched but still).
Another argument would be that many academia papers when just say "sample a random point" they usually meant a point you're not supposed to know dlog of (otherwise they state explicitly to sample scalar `s`, and do `G * s`). It is easy to make a mistake if the implementor is not fully aware or blindly follows what the paper says.
I am aware that it is easy to work around and use map to curve or hash to curve on randomly generated data.

Not sure what would be the proposed solution but at least document the proper use cases for it would be nice if changing the implementation itself is too much.

daxpedda:
This PR changes random point generation to use rejection sampling by de-serialization instead of deriving a point from a random scalar.

Fixes #1140.

### DIFF
--- a/k256/src/arithmetic/projective.rs
+++ b/k256/src/arithmetic/projective.rs
@@ -12,7 +12,6 @@ use elliptic_curve::{
     BatchNormalize, CurveGroup, Error, Result,
     group::{
         Group, GroupEncoding,
-        ff::Field,
         prime::{PrimeCurve, PrimeCurveAffine, PrimeGroup},
     },
     rand_core::TryRngCore,
@@ -411,7 +410,7 @@ impl Group for ProjectivePoint {
     type Scalar = Scalar;
 
     fn try_from_rng<R: TryRngCore + ?Sized>(rng: &mut R) -> core::result::Result<Self, R::Error> {
-        Ok(Self::GENERATOR * Scalar::try_from_rng(rng)?)
+        AffinePoint::try_from_rng(rng).map(Self::from)
     }
 
     fn identity() -> Self {
--- a/primeorder/src/projective.rs
+++ b/primeorder/src/projective.rs
@@ -258,13 +258,13 @@ where
 
 impl<C> Group for ProjectivePoint<C>
 where
-    Self: Double,
     C: PrimeCurveParams,
+    FieldBytes<C>: Copy,
 {
     type Scalar = Scalar<C>;
 
     fn try_from_rng<R: TryRngCore + ?Sized>(rng: &mut R) -> core::result::Result<Self, R::Error> {
-        Ok(Self::GENERATOR * <Scalar<C> as Field>::try_from_rng(rng)?)
+        AffinePoint::try_from_rng(rng).map(Self::from)
     }
 
     fn identity() -> Self {
@@ -311,8 +311,8 @@ where
 
 impl<C> CurveGroup for ProjectivePoint<C>
 where
-    Self: Double,
     C: PrimeCurveParams,
+    FieldBytes<C>: Copy,
 {
     type AffineRepr = AffinePoint<C>;
 
@@ -331,8 +331,8 @@ where
 
 impl<const N: usize, C> BatchNormalize<[ProjectivePoint<C>; N]> for ProjectivePoint<C>
 where
-    Self: Double,
     C: PrimeCurveParams,
+    FieldBytes<C>: Copy,
 {
     type Output = [<Self as CurveGroup>::AffineRepr; N];
 
@@ -348,8 +348,8 @@ where
 #[cfg(feature = "alloc")]
 impl<C> BatchNormalize<[ProjectivePoint<C>]> for ProjectivePoint<C>
 where
-    Self: Double,
     C: PrimeCurveParams,
+    FieldBytes<C>: Copy,
 {
     type Output = Vec<<Self as CurveGroup>::AffineRepr>;
 
@@ -400,16 +400,16 @@ where
 
 impl<C> LinearCombination<[(Self, Scalar<C>)]> for ProjectivePoint<C>
 where
-    Self: Double,
     C: PrimeCurveParams,
+    FieldBytes<C>: Copy,
 {
     // TODO(tarcieri): optimized implementation
 }
 
 impl<C, const N: usize> LinearCombination<[(Self, Scalar<C>); N]> for ProjectivePoint<C>
 where
-    Self: Double,
     C: PrimeCurveParams,
+    FieldBytes<C>: Copy,
 {
     // TODO(tarcieri): optimized implementation
 }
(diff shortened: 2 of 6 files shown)


k256/src/arithmetic/projective.rs
```diff
@@ -411,7 +410,14 @@ impl Group for ProjectivePoint {
     type Scalar = Scalar;
 
     fn try_from_rng<R: TryRngCore + ?Sized>(rng: &mut R) -> core::result::Result<Self, R::Error> {
-        Ok(Self::GENERATOR * Scalar::try_from_rng(rng)?)
+        let mut bytes = CompressedPoint::default();
+
+        loop {
+            rng.try_fill_bytes(bytes.as_mut())?;
+            if let Some(point) = Self::from_bytes(&bytes).into() {
```
tarcieri: `CompressedPoint` has a leading SEC1 tag, which is only allowed to be `0x02` or `0x03`, so the majority of these candidate points will be rejected (by a quick tag check, however...)

I think something like this should ensure the SEC1 tag is in the allowed range:

```suggestion
        let mut bytes = CompressedPoint::default();

        loop {
            rng.try_fill_bytes(bytes.as_mut())?;
            
            // Ensure SEC1 tag is 0x02 or 0x03 (compressed w\ even vs odd y-coordinate respectively)
            bytes[0] = (bytes[0] & 1) | 2;
            
            if let Some(point) = Self::from_bytes(&bytes).into() {
```

That will still give you a random distribution for the even vs odd y-coordinate bit, and clears all other bits except for setting the one that specifies a compressed point (0x02)

daxpedda: Oh god, good catch.

primeorder/src/projective.rs
```diff
@@ -258,13 +258,24 @@ where
...
+        loop {
+            rng.try_fill_bytes(bytes.as_mut())?;
+
+            // Ensure SEC1 tag is 0x02 or 0x03 (compressed w\ even vs odd y-coordinate respectively)
+            bytes.as_mut()[0] = (bytes.as_mut()[0] & 1) | 2;
+
+            if let Some(point) = Self::from_bytes(&bytes).into() {
+                return Ok(point);
+            }
+        }
```
tarcieri: You could avoid the bounds change here by using `AffinePoint<C>::decompress` (via the `DecompressPoint` trait) which is a bit more direct than `from_bytes` and avoids anything to do with SEC1.

That also works for `k256`.

andrewwhitehead:
Wouldn't there be a modulo bias without reducing a field element from a larger random input?

primeorder/src/projective.rs
daxpedda: Nice.

However, for the sign I needed to generate a extra random byte. WDYT?
https://github.com/RustCrypto/elliptic-curves/blob/50164762766b167fe361f50fb9da41d13c0a61cb/primeorder/src/projective.rs#L267-L278

tarcieri:
@andrewwhitehead no, it's the other way around: reduction introduces a bias, and even a wide reduction has a minute bias. Rejection sampling is unbiased.

https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/

primeorder/src/projective.rs
daxpedda: On that note: bounds still needed to change.

daxpedda:
> Wouldn't there be a modulo bias without reducing a field element from a larger random input?

No, de-serialization actually declines field elements that don't fit the modulus.
(`ed448-goldilocks` still needs some fixes in that area)

primeorder/src/projective.rs
```diff
@@ -258,13 +258,24 @@ where
...
     fn try_from_rng<R: TryRngCore + ?Sized>(rng: &mut R) -> core::result::Result<Self, R::Error> {
-        Ok(Self::GENERATOR * <Scalar<C> as Field>::try_from_rng(rng)?)
+        let mut bytes = FieldBytes::<C>::default();
+        let mut sign = 0;
+
+        loop {
+            rng.try_fill_bytes(&mut bytes)?;
+            rng.try_fill_bytes(core::array::from_mut(&mut sign))?;
+            if let Some(point) =
+                AffinePoint::decompress(&bytes, Choice::from(sign >> 7)).into_option()
```
tarcieri: Six of one, half dozen or another but: 

```suggestion
                AffinePoint::decompress(&bytes, Choice::from(sign & 1)).into_option()
```

primeorder/src/projective.rs
```diff
@@ -258,13 +258,24 @@ where
...
     type Scalar = Scalar<C>;
 
     fn try_from_rng<R: TryRngCore + ?Sized>(rng: &mut R) -> core::result::Result<Self, R::Error> {
-        Ok(Self::GENERATOR * <Scalar<C> as Field>::try_from_rng(rng)?)
+        let mut bytes = FieldBytes::<C>::default();
+        let mut sign = 0;
+
+        loop {
+            rng.try_fill_bytes(&mut bytes)?;
+            rng.try_fill_bytes(core::array::from_mut(&mut sign))?;
```
tarcieri: Hah, I guess it works! Surprised `TryRngCore` doesn't have a provided `try_next_u8` that uses this strategy

primeorder/src/projective.rs
daxpedda: Yeah, I like this better as well.

primeorder/src/projective.rs
```diff
@@ -258,13 +258,24 @@ where
...
+            if let Some(point) =
+                AffinePoint::decompress(&bytes, Choice::from(sign & 1)).into_option()
+            {
+                return Ok(point.into());
+            }
+        }
```
tarcieri: I know I just stamped approve, but now I'm kind of noticing that perhaps this should be an inherent `pub fn try_from_rng` impl'd on `AffinePoint`, and then the `ProjectivePoint` impl could be:

```suggestion
        AffinePoint::try_from_rng(rng).map(Into::into)
```

daxpedda: I agree, but don't we need a trait for that to implement it in `primeorder`?

daxpedda: Apparently not.
Working on it.

daxpedda: Done.

`EdwardsPoint` and `DecafPoint` still need a bit more work.
After this is merged I will do the necessary follow-up for `EdwardsPoint` in #1333.
Will make a PR for `DecafPoint` later.

tarcieri (approved):

PR #1345, opened 2025-08-02
`p521`: use reduction in `ReduceNonZero<U576>`

daxpedda:
I just realized from #1341 that P-521 wasn't implemented correctly.

Because `U576` can fit way more then `2 * ORDER`, doing `input - ORDER_MINUS_ONE` doesn't actually bring it below the modulus. We then use `Scalar::from_uint_unchecked()`, which does require input below the modulus.

I couldn't trigger a bug in practice however, not sure exactly how `fiat_p521_scalar_non_montgomery_domain_field_element()` fails when its requirements aren't fulfilled.

### DIFF
--- a/p521/src/arithmetic/scalar.rs
+++ b/p521/src/arithmetic/scalar.rs
@@ -22,7 +22,7 @@ use core::{
 };
 use elliptic_curve::{
     Curve as _, Error, FieldBytesEncoding, Result,
-    bigint::{self, Integer},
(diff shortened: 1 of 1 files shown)


tarcieri:
> I couldn't trigger a bug in practice however, not sure exactly how fiat_p521_scalar_non_montgomery_domain_field_element() fails when its requirements aren't fulfilled.

It's the mathematical equivalent of UB

tarcieri (approved):

PR #1347, opened 2025-08-02
`ed448-goldilocks`: reject identity points in `Group::try_from_rng()`

### LINKED ISSUE #1346: `Group::try_from_rng()` is not supposed to allow identity points
The documentation of [`Group::try_from_rng()`](https://docs.rs/group/0.14.0-pre.0/group/trait.Group.html#tymethod.try_from_rng) reads as following:
> Returns an element chosen uniformly at random from the non-identity elements of this group.

However, we currently, quite intentionally, do allow identity elements here. For non-identity elements we have [`NonIdentity::random()`](https://docs.rs/elliptic-curve/0.14.0-rc.10/elliptic_curve/point/struct.NonIdentity.html#method.random).

I'm not sure how we should address this, but at a minimum I think we should add some explicit documentation to these trait implementations.

daxpedda:
In #1333 I will remove this specific check from `AffinePoint::try_from_rng()`, which doesn't have the limitations of `Group::try_from_rng()`.

Fixes #1346.

### DIFF
--- a/ed448-goldilocks/src/decaf/points.rs
+++ b/ed448-goldilocks/src/decaf/points.rs
@@ -177,8 +177,11 @@ impl Group for DecafPoint {
         let mut bytes = DecafPointRepr::default();
 
         loop {
-            rng.try_fill_bytes(bytes.as_mut())?;
-            if let Some(point) = Self::from_bytes(&bytes).into() {
+            rng.try_fill_bytes(&mut bytes)?;
+            if let Some(point) = Self::from_bytes(&bytes)
+                .into_option()
+                .filter(|&point| point != Self::IDENTITY)
+            {
                 return Ok(point);
             }
         }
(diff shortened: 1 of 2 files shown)


ed448-goldilocks/src/decaf/points.rs
```diff
@@ -177,8 +177,11 @@ impl Group for DecafPoint {
         let mut bytes = DecafPointRepr::default();
 
         loop {
-            rng.try_fill_bytes(bytes.as_mut())?;
-            if let Some(point) = Self::from_bytes(&bytes).into() {
+            rng.try_fill_bytes(&mut bytes)?;
+            if let Some(point) = Self::from_bytes(&bytes)
+                .into_option()
+                .filter(|point| point != &Self::IDENTITY)
```
tarcieri: The borrow here is necessary? I thought `PartialEq` borrowed its operands to begin with.

daxpedda: Definitely necessary.
I can de-reference `point` instead.

tarcieri: Aah, `filter` borrows. Yeah, maybe `|&point|` instead?

daxpedda: Done.

PR #1360, opened 2025-08-05
`primeorder`: add optimized implementation for `LinearCombination`

daxpedda:
Fixing issues with the MSRV (v1.85).

Redo of #1343.

### DIFF
--- a/primeorder/src/projective.rs
+++ b/primeorder/src/projective.rs
@@ -380,7 +380,7 @@ where
             })
             .unzip();
 
-        lincomb(&mut ks, &mut pcs)
+        lincomb::<C>(&mut ks, &mut pcs)
     }
 }
 
@@ -395,7 +395,7 @@ where
         });
         let mut pcs: [_; N] = array::from_fn(|index| LookupTable::new(points_and_scalars[index].0));
 
-        lincomb(&mut ks, &mut pcs)
+        lincomb::<C>(&mut ks, &mut pcs)
     }
 }
 


tarcieri:
It's so weird. Pre-1.88 compilers seem to be confusing `ByteArray<C::Uint>` and `FieldBytes<C>` and I have no idea why.

@daxpedda as a workaround, and possibly a more efficient one, perhaps try serializing to `FieldBytes<C>`, which in many cases is a smaller array than `ByteArray<C::Uint>`

daxpedda:
I reproduced it locally at least now.

Will look into serializing it into `FieldBytes`, thanks!

daxpedda:
Yeah, I've no idea whats going on here. I converted it to `FieldBytes` and now compilation takes forever.
It does compile though ...

daxpedda:
Yeah, it doesn't work anyway because `FieldBytes` is big-endian (at least for `p256`) but we need little endian. I didn't dive deeper into this because there is a compilation bug in Rust that I would have to look into ...

Going back to analyzing why its expecting `FieldBytes` here to begin with.

tarcieri:
It's going to be big endian for everything in `primeorder`, at least for now.

Can you either `.reverse()` it, or reverse the iteration order?

daxpedda:
Sure, but what do I do about the compilation times? Its reaching minutes for p384.

daxpedda:
Okay, the compilation time issue is unrelated to this PR, its on master as well.
Not sure whats going on there, but I ran `cargo clean` and everything.
Checked with [the CI as well](https://github.com/RustCrypto/elliptic-curves/actions/runs/16749858379/job/47417104144) and it doesn't seem to have this issue, so probably just a local issue on my side.

In any case, I've fixed the issue now.
LMKWYT.

PR #1364, opened 2025-08-11
docs(README): Rust badge was unclickable

Radovenchyk:


tarcieri:
We generally don’t make the MSRV badge clickable

Radovenchyk:
understood

PR #1369, opened 2025-08-14
`ed448-goldilocks`: account for oddness in Scalar divisions

daxpedda:
Taking what I learned from https://github.com/dalek-cryptography/curve25519-dalek/pull/805 I applied this to `ed448-goldilocks` as well.

This is based on #1335 because the current check for torsion-freeness multiplies by the order, which breaks most proper assumptions made here.

### DIFF
--- a/ed448-goldilocks/src/field/scalar.rs
+++ b/ed448-goldilocks/src/field/scalar.rs
@@ -771,7 +771,7 @@ impl<C: CurveWithScalar> Scalar<C> {
     }
 
     /// Halves a Scalar modulo the prime
-    pub fn halve(&self) -> Self {
+    pub fn div_by_2(&self) -> Self {
         let is_odd = self.scalar.is_odd();
         let if_odd = self.scalar + *ORDER;
         let scalar = U448::conditional_select(&self.scalar, &if_odd, is_odd);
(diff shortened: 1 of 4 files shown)


ed448-goldilocks/src/field/scalar.rs
```diff
@@ -670,6 +670,18 @@ impl<C: CurveWithScalar> Scalar<C> {
...
+        let s_plus_2l = s_plus_l + ORDER;
+        let s_plus_3l = s_plus_2l + ORDER;
+
+        self.scalar.conditional_assign(&s_plus_l, s_mod_4.ct_eq(&1));
+        self.scalar
+            .conditional_assign(&s_plus_2l, s_mod_4.ct_eq(&2));
+        self.scalar
+            .conditional_assign(&s_plus_3l, s_mod_4.ct_eq(&3));
+
         self.scalar >>= 2;
```
tarcieri: Or you could just:

```suggestion
        self.halve().halve()
```

daxpedda: Right, will do!

tarcieri:
@daxpedda why is this marked as draft?

Edit: aah, this I guess?

> This is based on https://github.com/RustCrypto/elliptic-curves/pull/1335 because the current check for torsion-freeness multiplies by the order, which breaks most proper assumptions made here.

daxpedda:
It requires #1335.

The issue is that it messes with how scalar multiplication works, so the `is_torsion_free()` function, which multiplies a point with `ORDER` doesn't work anymore. Which is fine because `ORDER` isn't a valid scalar anyway. But #1335 implements a faster algorithm for `is_torsion_free()` which doesn't have the same problem.

PR #1374, opened 2025-08-21
x448: api fixups

baloo:
This is a followup to #1250


tarcieri (approved):
It's a start, though it would probably be good to have separate `StaticSecret` and `EphemeralSecret` types, where the latter provides no access to the inner secret key, only its public key and the ability to do Diffie-Hellman

PR #1379, opened 2025-08-25
Add FromUniformBytes<64> to {k,p}256::Scalar

kayabaNerve:


tarcieri (approved):
It would be nice to compose either the `FromUniformBytes` impl in terms of the `Reduce<WideBytes>` impl or vice versa, so we don't duplicate the implementation, but this is fine for now

tarcieri:
See also: #1341

PR #1381, opened 2025-08-26
bp256: fix divstep postcondition comment

Galoretka:
Correct the divstep spec comment for out5 to subtract arg5 from arg4 in the “then” branch. This aligns with standard fiat-crypto documentation. No functional code changes.


tarcieri:
While that's a little suspect, this is generated code emitted by `fiat-crypto`.

If there's an issue with the comment, it suggests an issue in `fiat-crypto`, which is the proper place to fix it.

PR #1389, opened 2025-09-02
`hash2curve`: make `trait GroupDigest` the default hash2curve cipher suite

daxpedda:
This PR changes `trait GroupDigest` to be the default hash2curve cipher suite according to the RFC.

To that effect the following changes were made:
- Move `GroupDigest::SecurityLevel` to `MapToCurve`.
- Introduce free standing functions to use hash2curve routines with custom parameters.
- Add a `hash2curve-default` crate feature to gate the associated hash crate while still having access to custom hash2curve routines via the `hash2curve` crate feature.
- Add a `type Hash` to `ExpandMsg`.
- Remove the `Hash` and `ExpandMsg` associated types from `OprfParameters`.
- Remove `MapToCurve::ScalarLength`, which isn't specified in the RFC and is custom to whatever implementation is using this.

This was previously discussed in #1295.

### DIFF
--- a/ed448-goldilocks/README.md
+++ b/ed448-goldilocks/README.md
@@ -18,6 +18,7 @@ It is intended to be portable, fast, and safe.
 
 ```rust
 use ed448_goldilocks::{Ed448, EdwardsPoint, CompressedEdwardsY, EdwardsScalar, sha3::Shake256};
+use elliptic_curve::consts::U84;
 use elliptic_curve::Field;
 use elliptic_curve::group::GroupEncoding;
 use hash2curve::{ExpandMsgXof, GroupDigest};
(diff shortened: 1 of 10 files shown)


carloskiki:
I don't like to have this trait _in addition to_ `GroupDigest`.

I think this should be the `GroupDigest` trait, and that for use cases that require "non-default" behavior, there should be `hash_to_curve` and `encode_to_curve` functions exposed that are parameterized over: The DST, the `ExpandMsg` routine, and the `Group`. This is what I described in this comment: https://github.com/RustCrypto/elliptic-curves/issues/1295#issuecomment-3086223270.

daxpedda:
I'm trying it out right now.

daxpedda:
Seems to work out great!

k256/Cargo.toml
```diff
@@ -56,6 +56,7 @@ ecdh = ["arithmetic", "elliptic-curve/ecdh"]
 ecdsa = ["arithmetic", "ecdsa-core/signing", "ecdsa-core/verifying", "sha256"]
 expose-field = ["arithmetic"]
 hash2curve = ["arithmetic", "dep:hash2curve"]
+hash2curve-default = ["hash2curve", "sha2"]
```
tarcieri: Why does this necessitate a new crate feature?

Is it really a problem to pull in `sha2` for `hash2curve`, even if the user decides to use a different hash function? That's how the `ecdsa` dependency works, FWIW.

daxpedda: I think its a bit different with ECDSA because there is no way to control the hash used for `generate_k()`, see https://github.com/RustCrypto/signatures/issues/949. But its still possible to use ECDSA with `p256` (and others) without pulling in `sha2` by just using `ecdsa::hazmat` directly.

The use-case here is simply that it should be an option to not pull in those dependencies if they are entirely unused.

In the meantime I've realized that we could at least rename `hash2curve-default` to `group-digest`. I'm gonna go ahead and do just that.

Let me know what you think!

daxpedda:
Well, that's a really good catch! I've already seen so many implementations out there getting this totally wrong.

Will fix it up.

tarcieri (approved):

PR #1394, opened 2025-09-03
Implement `CofactorGroup` for all relevant curves

daxpedda:
This PR implements `CofactorGroup` for `ed448-goldilocks` curves, `k256` and `primeorder` curves. Apart from `EdwardsPoint` they are all no-op.

Additionally, `MapToCurve` now can rely on `CofactorGroup` to facilitate `hash_from_bytes()` and `encode_from_bytes()`.

### DIFF
--- a/ed448-goldilocks/src/decaf/points.rs
+++ b/ed448-goldilocks/src/decaf/points.rs
@@ -226,11 +226,11 @@ impl CofactorGroup for DecafPoint {
     type Subgroup = DecafPoint;
 
     fn clear_cofactor(&self) -> Self::Subgroup {
-        self.double().double()
+        *self
     }
 
     fn into_subgroup(self) -> CtOption<Self::Subgroup> {
-        CtOption::new(self.clear_cofactor(), self.is_torsion_free())
+        CtOption::new(self, Choice::from(1))
     }
 
     fn is_torsion_free(&self) -> Choice {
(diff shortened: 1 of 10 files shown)


tarcieri:
It seems like it's still valid to have an impl of `CofactorGroup` even for a cofactor 1 group, and perhaps we should consider adding a `CofactorGroup` impl to e.g. `k256` and `primeorder`.

For reference, @str4d's PR to add `group` support to `curve25519-dalek` includes this impl of `CofactorGroup` for `RistrettoPoint`: https://github.com/dalek-cryptography/curve25519-dalek/pull/473/files#diff-b3b872cfe8ac2ba85d056e84fbadfdccbb7a52d9111fe3331eeb45ef3f411188R1220-R1236

The comment notes:

> Ristretto has a cofactor of 1.

daxpedda:
> It seems like it's still valid to have an impl of `CofactorGroup` even for a cofactor 1 group, and perhaps we should consider adding a `CofactorGroup` impl to e.g. `k256` and `primeorder`.

That sounds good to me. Will rework it soon.

It might also be possible to rely on `CofactorGroup` for the `MapToCurve` default implementations, so I will be looking into that as well.

carloskiki:
This trio of traits feels a lot like `Fn`, `FnMut`, and `FnOnce`. IMHO yes, a `PrimeGroup` should still implement `CofactorGroup` and `Group`, just like `Fn` can also be used as `FnMut` and `FnOnce`.

daxpedda:
Done.

tarcieri (approved):

PR #1404, opened 2025-09-04
`primeorder`: convert hash2curve mapping directly to `AffinePoint`

daxpedda:
This PR moves `hash2curve::osswu` to `primeorder` and `hash2curve::isogeny` to `k256`, because these were curve specific implementations to begin with.

It also introduces a new trait to `primeorder`: `AffineOsswuMap`. Which is a direct way to convert the OSSWU mapping to an `AffinePoint`, instead of having to go through point decompression.

Addresses https://github.com/RustCrypto/traits/issues/1994.


k256/Cargo.toml
```diff
@@ -27,6 +27,7 @@ hash2curve = { version = "0.14.0-rc.1", optional = true }
 once_cell = { version = "1.21", optional = true, default-features = false }
 ecdsa-core = { version = "0.17.0-rc.6", package = "ecdsa", optional = true, default-features = false, features = ["der"] }
 hex-literal = { version = "1", optional = true }
+primeorder = { version = "=0.14.0-pre.8", optional = true }
```
tarcieri, line 30: @fjarri I'm curious if we could start leveraging `primeorder` in `k256`.

I think it would be particularly interesting if we had traits for type-safe lazy reductions: https://github.com/dalek-cryptography/curve25519-dalek/pull/816/files#diff-6df974cf03dd3c0df6a96d945cef0d6a7bfd2967fa99ab56280d15b382e5270d

fjarri: Looks useful, but I am a little uncertain about using `typenum`. Are all the usecases fully static? Or is it possible to sometimes have a runtime-determined number of operations?

tarcieri: Having static type safety is definitely the goal. I think typenum can be avoided at the cost of a coarser grained abstraction that permits only a single unreduced operation (e.g. add) before requiring it be used as an input to an operation that performs a reduction (e.g. mul), but it is also cool to capture precise bounds using the type system

PR #1405, opened 2025-09-04
Implement `AffineCoordinates::from_coordinates`

daxpedda:
Implementation of https://github.com/RustCrypto/traits/pull/1996 addressing https://github.com/RustCrypto/traits/issues/1994.

I didn't add a const method to `AffinePoint`, but let me know if this is desired at this point and I'm happy to do it here or in a follow-up.

### DIFF
--- a/ed448-goldilocks/src/decaf/affine.rs
+++ b/ed448-goldilocks/src/decaf/affine.rs
@@ -1,13 +1,13 @@
 use crate::curve::twedwards::affine::AffinePoint as InnerAffinePoint;
 use crate::field::FieldElement;
-use crate::{Decaf448FieldBytes, DecafPoint, DecafScalar};
+use crate::{Decaf448FieldBytes, DecafPoint, DecafScalar, ORDER};
 use core::ops::Mul;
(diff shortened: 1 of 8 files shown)


ed448-goldilocks/src/decaf/affine.rs
```diff
@@ -39,6 +39,18 @@ impl Eq for AffinePoint {}
...
 
+    fn decode(x: &Self::FieldRepr, y: &Self::FieldRepr) -> CtOption<Self> {
+        let point = Self(InnerAffinePoint {
+            x: FieldElement::from_bytes(&x.0),
+            y: FieldElement::from_bytes(&y.0),
+        });
+
+        CtOption::new(
+            point,
+            point.0.is_on_curve() & (point * DecafScalar::new(*ORDER)).ct_eq(&DecafPoint::IDENTITY),
```
daxpedda, line 50: AFAIK only passing the curve equation doesn't mean its torsion free.
I'm also not aware of any cheaper way to determine torsion-freeness.

tarcieri (approved):

PR #1412, opened 2025-09-12
primefield: make `MontyFieldElement` the interior field element type

### LINKED ISSUE #1191: primefield: use `ConstMontyForm` for inner type of field elements
Currently field elements are represented internally as [`Uint`](https://docs.rs/crypto-bigint/latest/crypto_bigint/struct.Uint.html), however those `Uint`s are generally constrained to the set of valid field elements for a given modulus represented in Montgomery form, namely with the fiat-crypto generated field implementations used by the majority of the curve crates in this repo, but also in the `p256` implementation which is handwritten.

For field elements in Montgomery form it would be preferable to use [`ConstMontyForm`](https://docs.rs/crypto-bigint/latest/crypto_bigint/modular/struct.ConstMontyForm.html) internally, since this ensures at a type-level that it isn't possible to generate invalid field elements, as we recently saw in #1153.

This approach would also unlock using `crypto-bigint` to provide the field implementation, rather than `fiat-crypto`, which could unlock potential performance improvements.

tarcieri:
Modifies the field type macros to emit code which expects a `MontyFieldElement` as the interior type, itself a newtype for `ConstMontyForm` from the `crypto-bigint` crate.

Previously they were `crypto_bigint::Uint`s in Montgomery form, which couldn't leverage any of the modular arithmetic support from that crate.

Notably, this change doesn't leverage that support quite yet: instead the existing fiat-crypto synthesized arithmetic implementations are retained (or in the case of `p256`, handwritten ones). That keeps the impact of this particular change low so we can get it in place first and then experiment with using `crypto-bigint` itself for modular arithmetic.

The longer-term goal will be to expose `cfg` attributes that make it possible to select between `crypto-bigint` or `fiat-crypto` for the arithmetic implementation.

Closes #1191

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -319,8 +319,7 @@ checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 [[package]]
 name = "crypto-bigint"
 version = "0.7.0-rc.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7c069823f41bdc75e99546bfd59eb1ed27d69dc720e5c949fe502b82926f8448"
+source = "git+https://github.com/RustCrypto/crypto-bigint#7bb8509234898b8bba62aea0343bbd99b5439944"
 dependencies = [
  "hybrid-array",
  "num-traits",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,11 @@ members = [
 opt-level = 2
 
 [patch.crates-io]
-ed448-goldilocks = { path = "ed448-goldilocks" }
 # https://github.com/RustCrypto/traits/pull/1996
+crypto-bigint = { git = "https://github.com/RustCrypto/crypto-bigint" }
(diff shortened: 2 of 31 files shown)


tarcieri:
cc @andrewwhitehead

tarcieri:
Checking benchmarks I'm not noticing any impact, e.g.

```
field element operations/mul
                        time:   [14.537 ns 14.556 ns 14.576 ns]
                        change: [−0.1801% +0.0882% +0.3643%] (p = 0.53 > 0.05)
                        No change in performance detected.
```

I wouldn't expect any impact though as this change largely uses `MontyFieldElement` the same way `Uint` was being used, with `fiat-crypto` providing the real field arithmetic impls (or as noted earlier in `p256`'s case, the original handwritten impl)

andrewwhitehead:
Using `cfg` attributes to switch between fiat and bigint sounds like it would require separate dependencies to use the same implementation?

tarcieri:
@andrewwhitehead it gives the toplevel application exclusive control of the backend, so dependencies can't request conflicting options (as with e.g. features)

PR #1423, opened 2025-09-13
Bump `signature` to v3.0.0-rc.4

tarcieri:
Also bumps `ecdsa` to v0.17.0-rc.7

This brings `*DigestSigner`/`*DigestVerifier` changes introduced in RustCrypto/traits#2004

cc @daxpedda

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -363,8 +363,9 @@ dependencies = [
 
 [[package]]
 name = "ecdsa"
-version = "0.17.0-rc.6"
-source = "git+https://github.com/RustCrypto/signatures?rev=2cd53cf4fbe82c61c2e2ac7b2099ba743b01074e#2cd53cf4fbe82c61c2e2ac7b2099ba743b01074e"
+version = "0.17.0-rc.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4ab355ec063f7a110eb627471058093aba00eb7f4e70afbd15e696b79d1077b"
 dependencies = [
  "der",
  "digest",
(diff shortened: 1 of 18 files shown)


daxpedda:
@tarcieri should I do a follow-up where I change it to `Default + FixedOutput + HashMarker + Update`?

tarcieri:
@daxpedda sure, if you can do that across the board both here and in the `signatures` repo that'd be great

daxpedda:
@tarcieri so I did attempt to do this for `signatures`, but its not that straightforward.
You can find my attempt here: https://github.com/daxpedda/signatures/commit/f89e41033ee015e9b11c34c0839c9bc5cf0385cd.

Basically `dsa` and `ecdsa` require `rfc6979`, which in turn uses `SimpleHmacReset`, which requires `Digest`.
So the first thing I did is switch `rfc6979` to `HmacReset`, which "just" requires `EagerHash` (which still requires `Digest`).
After that, I started replacing all mentions of `Digest` in `dsa` and `ecdsa` with `EagerHash`.
This worked for `dsa`, but `ecdsa` doesn't use `rfc6979` for verifying, which is behind a separate crate feature, so it doesn't pull in `hmac` either, not giving me access to `EagerHash`.
Because `ecdsa::hazmat::DigestAlgorithm::Digest` now requires `EagerHash`, I would need to pull in `hmac` for `feature = "verifying"`.

So its all solvable, but the change is pervasive, let me know if you want me to continue with `EagerHash`.

Alternatively, we could just make `Digest` a supertrait over `Update`, among other traits. It always bothered me in the first place that this isn't the case.

I assume I will hit the same snag in `elliptic-curves` because its built on `ecdsa`.

tarcieri:
Perhaps `ecdsa` has too many knobs. I've been thinking about getting rid of the features which avoid providing an algorithm implementation.

For now though, consolidating into one feature for both signing and verification seems fine to me.

daxpedda:
What about changing everything from `Digest` to `EagerHash`? That's a pretty big change, is that alright?

tarcieri:
Definitely makes sense for `*Signer` due to the use of HMAC. I guess I still don't completely understand why it's needed for `*Verifier`. Is it just for parity with `*Signer`?

daxpedda:
Its because both signer and verifier implementations use `type ecdsa::hazmat::DigestAlgorithm::Digest`. But to make this type work for signer implementations I had to write `type Digest: EagerHash` and I can't write that if I'm unable to import the type.

tarcieri:
Aha, makes sense. Sounds fine to change them all to `EagerHash` then.

PR #1440, opened 2025-09-16
primefield: impl `ConstMontyParams` for field element types

tarcieri:
- Adds an impl to `primefield::MontyFieldElement`
- Also emits an impl from the `monty_field_element!` macro

### DIFF
--- a/p384/src/arithmetic/hash2curve.rs
+++ b/p384/src/arithmetic/hash2curve.rs
@@ -97,7 +97,7 @@ mod tests {
     use elliptic_curve::{
         Curve,
         array::Array,
-        bigint::{ArrayEncoding, CheckedSub, NonZero, U384, U576},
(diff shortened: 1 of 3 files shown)


p384/src/arithmetic/hash2curve.rs
```diff
@@ -113,7 +113,7 @@ mod tests {
     fn params() {
         let params = <FieldElement as OsswuMap>::PARAMS;
 
-        let c1 = FieldElement::PARAMS
+        let c1 = <FieldElement as ConstMontyParams<{ U384::LIMBS }>>::PARAMS
```
tarcieri, line 116: @daxpedda hmm, seems a little problematic for these to have an associated constant named `PARAMS`

daxpedda: Hm. I hope in practice this isn't a problem because the `OsswuMap` trait should never be imported other then in the `hash2curve` module.

But let me know and I can put a PR to change the name.

PR #1446, opened 2025-09-17
primefield: generic implementation of `sqrt`

tarcieri:
Adapts the implementation of `sqrt` from the `ff_derive` crate:

https://github.com/zkcrypto/ff/blob/f77b814/ff_derive/src/lib.rs#L509

...which is in turn adapted from this paper:

https://eprint.iacr.org/2012/685.pdf

For the most part the sqrt impls replaced are special case versions of the code which would've otherwise been derived by `ff_derive`.

Some TODOs are left to make more computations `const`.


tarcieri:
~~TODO: add unit tests for the two different `sqrt` impls~~ Edit: done, generic implementation of Tonelli-Shanks now unit tested against the P-192 scalar modulus

tarcieri:
Note: some of the impls removed in this PR use Atkins algorithm, which can be added back in a followup. But first, it would be good to ensure `const`-friendly selection of the algorithm

PR #1449, opened 2025-09-17
primefield: add a `sqrt::Algorithm` enum

tarcieri:
Enumerates the available square root algorithms as `enum Algorithm`.

Notably this makes it easy to test which algorithm is selected for a particular modulus.


tarcieri:
cc @andrewwhitehead

PR #1454, opened 2025-09-19
x448: Add `StaticSecret` in a feature-flag

baloo:
Some implementations (like rpgp) will serialize a secret and can use static secret similar to what we can find in x25519.

This is a fixup on https://github.com/RustCrypto/elliptic-curves/pull/1378 which removed that support


baloo:
I messed up something, I'm not sure what yet, but downstream tests are failing

PR #1457, opened 2025-09-23
Fix `EdwardsScalar::from_canonical_bytes()` overflow

daxpedda:
The current `EdwardsScalar::from_canonical_bytes()` implementation can trigger integer overflow by simply having the last byte be `128`. This PR fixes that.


ed448-goldilocks/src/edwards/scalar.rs
```diff
@@ -28,13 +28,8 @@ impl CurveWithScalar for Ed448 {
     }
 
     fn from_canonical_bytes(bytes: &ScalarBytes<Self>) -> subtle::CtOption<Scalar<Self>> {
-        fn is_zero(b: u8) -> Choice {
-            let res = b as i8;
-            Choice::from((((res | -res) >> 7) + 1) as u8)
-        }
```
daxpedda, line 34: AFAICT this is just an overly complicated way to compare to zero.

tarcieri:
@daxpedda hmm, seems like CI broke. Can you try pushing again?

PR #1461, opened 2025-09-24
`ed448-goldilocks`: use Jacobi Symbol

daxpedda:
This makes use of the new `JacobiSymbol` from `crypto-bigint` to optimize the `FieldElement::is_square()` implementation.

I also went ahead and optimized `FieldElement::sqrt()`. The optimization just expands the exponent, but I took the implementation from [pornin/crrl](https://github.com/pornin/crrl/blob/4cc7cbbe8796ee8d459b815d81318603279879e4/src/backend/w64/gf448.rs#L1621-L1633).

### DIFF
--- a/ed448-goldilocks/src/field/element.rs
+++ b/ed448-goldilocks/src/field/element.rs
@@ -334,17 +334,23 @@ impl FieldElement {
     }
 
     pub fn is_square(&self) -> Choice {
-        const IS_SQUARE_EXP: U448 = U448::from_le_hex(
-            "ffffffffffffffffffffffffffffffffffffffffffffffffffffff7fffffffffffffffffffffffffffffffffffffffffffffffffffffff7f",
-        );
-        self.0.pow(&IS_SQUARE_EXP).ct_eq(&FieldElement::ONE.0)
+        !Choice::from(self.0.jacobi_symbol().is_minus_one())
     }
 
     pub fn sqrt(&self) -> FieldElement {
-        const SQRT_EXP: U448 = U448::from_be_hex(
-            "3fffffffffffffffffffffffffffffffffffffffffffffffffffffffc0000000000000000000000000000000000000000000000000000000",
-        );
-        Self(self.0.pow(&SQRT_EXP))
+        // Candidate root is y = x^((p+1)/4).
+        // We have: (p+1)/4 = 2^446 - 2^222
+        let z = *self;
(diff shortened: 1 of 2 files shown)


ed448-goldilocks/src/field/element.rs
```diff
@@ -334,17 +334,23 @@ impl FieldElement {
     }
 
     pub fn is_square(&self) -> Choice {
-        const IS_SQUARE_EXP: U448 = U448::from_le_hex(
-            "ffffffffffffffffffffffffffffffffffffffffffffffffffffff7fffffffffffffffffffffffffffffffffffffffffffffffffffffff7f",
-        );
-        self.0.pow(&IS_SQUARE_EXP).ct_eq(&FieldElement::ONE.0)
+        !Choice::from(self.0.jacobi_symbol().is_minus_one())
```
andrewwhitehead, line 337: Not using `JacobiSymbol::is_one`?

daxpedda: I wanted to include `Zero` as well.
Which I believe is correct?

tarcieri: I think the previous implementation would've returned false in such a case, though I would agree it should probably return true

ed448-goldilocks/src/field/element.rs
```diff
@@ -334,17 +334,23 @@ impl FieldElement {
...
+        let zp2 = z.square() * z;
+        let zp3 = zp2.square() * z;
+        let zp4 = zp3.square() * z;
+        let zp7 = zp4.square_n::<3>() * zp3;
+        let zp14 = zp7.square_n::<7>() * zp7;
+        let zp28 = zp14.square_n::<14>() * zp14;
+        let zp56 = zp28.square_n::<28>() * zp28;
+        let zp112 = zp56.square_n::<56>() * zp56;
+        let zp224 = zp112.square_n::<112>() * zp112;
+        zp224.square_n::<222>()
```
tarcieri, line 353: Nice addition chain here

ed448-goldilocks/src/field/element.rs
andrewwhitehead: Zero should not be possible for a prime modulus

Edit: actually I see if the input is zero, the output can be zero, except for (0|1), so this is probably correct.

PR #1465, opened 2025-09-24
p521: add 32-bit fiat scalar backend; use `primefield`

tarcieri:
Synthesizes a 32-bit backend using fiat-crypto v0.1.5. This was previously a bit tricky because it causes a stack overflow with default ulimits. Whether this is actually a good idea is a bit questionable as it ends up being quite large at 22.5 KLOCs.

The particularly nice thing it does afford, however, is being able to use `primefield` to write the field implementation. This makes it consistent with the other implementations and addresses some TODOs, like a sqrt implementation.


tarcieri:
Seems we'll need to use a `Uint<17>` on 32-bit platforms

tarcieri:
Seems like this will involve a wholesale conversion of both the base and scalar fields, since it will need to use a `Uint<17>` (effectively `U544`) on 32-bit platforms

PR #1470, opened 2025-09-25
p521: use `primefield` for scalar field

tarcieri:
We attempted to synthesize a fiat-crypto 32-bit scalar backend in #1465. It was 22.5 KLOCs, took 45s to run rustfmt over, and rustc spent longer than I was willing to wait trying to compile it without `--release` (several minutes).

In lieu of that, this stubs out a 32-bit backend so we can use `primefield`, but implements it by wrapping the 64-bit backend and applying the same sorts of conversions we were already using, just doing it in stubbed out `fiat-crypto` shaped code where we could conceivably drop-in a 32-bit backend if rustc didn't choke on it.


tarcieri:
The issue was the 22.5 KLOCs of `fiat-crypto` generated code for 32-bit backends seems to computationally choke `rustc`/LLVM unless you use `--release`, which is practically untenable. Without `--release` the compiler just sits there chewing on the code for longer than I've had the patience to wait through.

I'm going to add a small shim that allows it to continue using the 64-bit backend we know works even in debug mode, but also otherwise follow the same patterns the other crates use and have `primefield` write the field impl.

Perhaps we should look into if `fiat-crypto` can generate a scalar implementation which has a more compact code size. For comparison, the P-521 base field impl is only 2 KLOCs on 32-bit targets.

tarcieri:
I don't think this is going to work because the synthesized 64-bit `fiat-crypto` backend seems to use a different representation than `MontyFieldElement`, namely it's effectively left shifted by 32-bits versus `MontyFieldElement`, where the least significant bit starts at 0 instead of 32:

https://github.com/mit-plv/fiat-crypto/issues/2166

tarcieri:
Closing for now until such time as `crypto-bigint` and `fiat-crypto` have compatible Montgomery form representations

PR #1475, opened 2025-09-28
[WIP] primeorder: `LookupTable` and `BasepointTable`

tarcieri:
Extracts the `LookupTable` type from `k256`, and the code for calculating basepoint tables and storing them in lazily computed statics.

Notably this retains support for either using `std::sync::LazyLock` with no additional dependencies, or leveraging the `critical-section` feature of `once_cell` on `no_std` targets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -634,7 +634,6 @@ dependencies = [
  "hex-literal",
  "num-bigint 0.4.6",
  "num-traits",
- "once_cell",
  "primeorder",
  "proptest",
  "rand_core",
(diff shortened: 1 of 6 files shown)


tarcieri:
After attempting this with `primeorder`, I think the functionality should instead go in `elliptic-curve` for a couple reasons:

1. This functionality is potentially useful for non-prime-order curves, e.g. `ed448-golidilocks`
2. To continue to support `k256` in its current form I made it generic around a `Point` type where `Point` impls `Group`, so it's not using `primeorder::ProjectivePoint`. The entire implementation is independent from everything else in `primeorder`.

Given that, I think I'm going to close this out and attempt extracting this into `elliptic-curve` instead.

tarcieri:
Closing in favor of https://github.com/RustCrypto/traits/pull/2023

PR #1503, opened 2025-10-24
`bignp256`: fixes

makavity:
* ecdh - generic param for hash
* remove deprecated stuff
* verify is constant time now

### DIFF
--- a/bignp256/src/lib.rs
+++ b/bignp256/src/lib.rs
@@ -58,6 +58,7 @@ pub mod public_key;
 pub mod secret_key;
 
 #[cfg(feature = "pkcs8")]
+#[allow(dead_code)]
 const ALGORITHM_OID: pkcs8::ObjectIdentifier =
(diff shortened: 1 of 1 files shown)


bignp256/src/ecdsa/verifying.rs
```diff
@@ -152,7 +153,7 @@ impl PrehashVerifier<Signature> for VerifyingKey {
         s0.reverse();
 
         // 8. If 𝑆0 != 𝑡, return NO.
-        if s0 == &t.as_slice()[..16] {
+        if s0.ct_eq(&t.as_slice()[..16]).into() {
```
tarcieri, line 156: This doesn't typically matter in a verification context, unless you want whole algorithm to be constant-time for some reason (I know there are some cases of "unusual requirements" for that)

FWIW I hope in the next release we can switch to using wNAF for verification which will be variable-time but faster. The `ecdsa` crate already makes use of variable-time inversions.

PR #1510, opened 2025-11-01
ed448-goldilocks: rename `FieldElement::sqrt()` to `unchecked_sqrt()`

### LINKED ISSUE #1509: ed448-goldilocks: `FieldElement::sqrt` should verify candidate square root is a valid quadratic residue
Currently `FieldElement::sqrt` is infallible and does not ensure a candidate square root it computes is actually a valid quadratic residue.

This means it may return incorrect results in the event the input isn't actually a perfect square to begin with.

The code was adapted from `crrl`, which validates candidate square roots by squaring them and ensuring the result is equal to the input: https://github.com/pornin/crrl/blob/4cc7cbbe8796ee8d459b815d81318603279879e4/src/backend/w64/gf448.rs#L1639-L1646

`Scalar::sqrt` handles this correctly

cc @daxpedda

daxpedda:
Fixes #1509.

### DIFF
--- a/ed448-goldilocks/src/field/element.rs
+++ b/ed448-goldilocks/src/field/element.rs
@@ -262,7 +262,7 @@ impl Field for FieldElement {
     }
 
     fn sqrt(&self) -> CtOption<Self> {
-        let sqrt = self.sqrt();
+        let sqrt = self.unchecked_sqrt();
         CtOption::new(sqrt, sqrt.square().ct_eq(self))
     }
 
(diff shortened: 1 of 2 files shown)


ed448-goldilocks/src/field/element.rs
```diff
@@ -337,7 +337,7 @@ impl FieldElement {
         !Choice::from(self.0.jacobi_symbol().is_minus_one())
     }
 
-    pub fn sqrt(&self) -> FieldElement {
+    pub fn unchecked_sqrt(&self) -> FieldElement {
```
tarcieri, line 340: I'd say "perhaps this shouldn't be pub" but I guess we don't expose `FieldElement` anywhere yet.

Sidebar: perhaps we should enable the `unreachable_pub` lint...

PR #1513, opened 2025-11-05
decaf: make DecafPoint::ct_eq compare canonical encoding

radik878:
Fix DecafPoint::ct_eq to compare canonical Decaf encodings via compress() instead of checking X1*Y2 == Y1*X2.

The previous implementation ignored Z and only compared the X/Y ratio, which is insufficient for projective Edwards coordinates and, more importantly, does not implement Decaf equality semantics (equivalence of encoding/class).
Using compress() ensures constant-time equality on the canonical representation, aligning with Decaf/Ristretto practice and preventing false (in)equality across different projective representatives of the same Decaf element.

### DIFF
--- a/ed448-goldilocks/src/decaf/points.rs
+++ b/ed448-goldilocks/src/decaf/points.rs
@@ -62,7 +62,7 @@ impl UpperHex for DecafPoint {
 
 impl ConstantTimeEq for DecafPoint {
     fn ct_eq(&self, other: &DecafPoint) -> Choice {
-        (self.0.X * other.0.Y).ct_eq(&(self.0.Y * other.0.X))
+        self.compress().ct_eq(&other.compress())
     }
 }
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Per https://ristretto.group/details/equality.html

> However, testing whether two points lie in the same coset can be done in projective coordinates, making it actually easier than equality testing in the original non-quotient group.
>
> The Decaf paper proves that in the cofactor-4 case, to test equality of P₁ = (X₁:Y₁:Z:T₁) and P₂ = (X₂:Y₂:Z₂:T₂) it's sufficient to test whether
>
> X₁Y₂ = Y₁X₂.

Comparing by compressing points will incur a square root. I'll note that `curve25519-dalek` originally implemented `ConstantTimeEq` by compressing points but switched to this sort of coset comparison later as an optimization:

https://github.com/dalek-cryptography/curve25519-dalek/commit/7097d8f98e7e3226f88c52bcf2637be55874ccea

tarcieri:
Closing as working as intended

PR #1518, opened 2025-11-06
fix: correct curve equation sign in AffinePoint::is_on_curve

DeVikingMark:
Fix the sign in the twisted Edwards curve equation check. 
The formula was using addition instead of subtraction, which didn't match thedocumented curve equation y^2 - x^2 = 1 - d*x^2*y^2.

Changed FieldElement::ONE + to FieldElement::ONE - in the is_on_curve method to correctly validate points on the curve.

### DIFF
--- a/ed448-goldilocks/src/curve/twedwards/affine.rs
+++ b/ed448-goldilocks/src/curve/twedwards/affine.rs
@@ -35,7 +35,7 @@ impl AffinePoint {
         let xx = self.x.square();
         let yy = self.y.square();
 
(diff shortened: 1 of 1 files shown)


DeVikingMark:
Oh, I was wrong probably. Equation: y^2 - x^2 = 1 - 39082*x^2*y^2 and TWISTED_D is defined as -39082 (mod p), so we use addition here right?

tarcieri:
The current version looks correct

PR #1537, opened 2025-11-25
fix: Use CurveWithScalar::to_repr in serde for Scalar and read full ReprSize

Forostovec:
Switch Scalar serde to use the curve-specific canonical representation by serializing C::to_repr(self) and deserializing into the full ScalarBytes buffer instead of a hardcoded 56-byte slice. This aligns serde with PrimeField::to_repr, LowerHex/UpperHex, and Vec conversions, and fixes the Ed448 case where the canonical representation is 57 bytes per RFC 8032. The Edwards serde_bare test is adjusted to expect 58 bytes due to the 57-byte payload plus the slice length prefix, while Decaf remains unchanged.

### DIFF
--- a/ed448-goldilocks/src/field/scalar.rs
+++ b/ed448-goldilocks/src/field/scalar.rs
@@ -430,7 +430,8 @@ impl<C: CurveWithScalar> serdect::serde::Serialize for Scalar<C> {
     where
         S: serdect::serde::Serializer,
     {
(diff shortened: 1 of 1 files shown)


tarcieri:
I'd like to get #1471 merged first before we do anything else with scalars.

There's also #1463.

cc @daxpedda

PR #1541, opened 2025-12-01
chore(k256): replace single-iteration loop with direct assignment in isogeny

radik878:
Simplify the computation of xs[3] in isogeny() by replacing a one-iteration loop with xs[3] = xs[2] * x. This clarifies intent, avoids constructing a Range for a single step, and keeps the faster square() for x^2. No functional impact; tests should behave identically.


tarcieri:
Looks OK at first glance.

cc @daxpedda

PR #1548, opened 2025-12-16
p384: add `cfg(p384_backend = "bignum")`

tarcieri:
Adds support for an experimental backend which uses `crypto-bigint` as the field element representation, as an off-by-default alternative to `fiat-crypto`.

This uses the `monty_field_arithmetic!` macro introduced in #1547 which in turn builds on the `MontyFieldElement` type introduced in #1311, which builds on `crypto-bigint`'s `ConstMontyForm` type, which implements Montgomery form modular arithmetic.

### DIFF
--- a/.github/workflows/p384.yml
+++ b/.github/workflows/p384.yml
@@ -100,6 +100,10 @@ jobs:
       - run: cargo test --release --target ${{ matrix.target }} --no-default-features
       - run: cargo test --release --target ${{ matrix.target }}
       - run: cargo test --release --target ${{ matrix.target }} --all-features
+      - env:
+          RUSTFLAGS: '--cfg p384_backend="bignum"'
+          RUSTDOCFLAGS: '--cfg p384_backend="bignum"'
+        run: cargo test --release --target ${{ matrix.target }} --all-features
 
   cross:
     strategy:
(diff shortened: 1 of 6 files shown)


tarcieri:
It seems like we should add backends to the test matrix for various jobs, so we can test both backends as rigorously as each other

tarcieri:
The benchmarks are kind of wild on this. First the good news:

```
point operations/point-scalar mul
                        time:   [397.11 µs 398.10 µs 399.12 µs]
                        change: [−14.817% −13.846% −12.997%] (p = 0.00 < 0.05)
                        Performance has improved.
```

```
field element operations/invert
                        time:   [4.2699 µs 4.3106 µs 4.3647 µs]
                        change: [−86.537% −86.385% −86.212%] (p = 0.00 < 0.05)
                        Performance has improved.
```

point-scalar mul is the complex high-level operation we care about, so it's great to see that improve. Our inversion implementation in `crypto-bigint` is also quite fast, so it's great to see a big performance improvement there.

What's weird are all the microbenchmarks on the individual ops:

```
field element operations/mul
                        time:   [41.379 ns 41.918 ns 42.611 ns]
                        change: [+8000.7% +8128.1% +8281.1%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

The previous implementation was measured in picoseconds, so I'm wondering if it was getting optimized away or something. I've tried sprinkling `black_box` in places to no effect.

PR #1550, opened 2025-12-17
k256: use `crypto-bigint` to implement field inversions

tarcieri:
Leverages the safegcd-bounds implementation from `crypto-bigint`. This adds a hard dependency on `primefield` in order to use `primefield::monty_field_params!` and `MontyFieldElement`.

Inversions are implemented by first converting from `FieldElement` to its canonical representative (i.e. big endian byte serialization), then parsing those bytes as a `MontyFieldElement`, which converts to a saturated Montgomery form representation. Once the inversion is complete, it uses a similar process to convert back.

Despite the circuitous conversions, safegcd-bounds is so fast it's still a fairly significant win, doubling performance:

    field element operations/invert
                        time:   [2.5839 µs 2.5915 µs 2.5993 µs]
                        change: [−51.091% −50.749% −50.449%] (p = 0.00 < 0.05)

See also: #1549 a similar change to `p256`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -632,6 +632,7 @@ dependencies = [
  "hex-literal",
  "num-bigint 0.4.6",
  "num-traits",
+ "primefield",
  "primeorder",
  "proptest",
  "rand 0.10.0-rc.5",
--- a/k256/Cargo.toml
+++ b/k256/Cargo.toml
@@ -26,6 +26,7 @@ hash2curve = { version = "0.14.0-rc.4", optional = true }
 # optional dependencies
 ecdsa-core = { version = "0.17.0-rc.9", package = "ecdsa", optional = true, default-features = false, features = ["der"] }
 hex-literal = { version = "1", optional = true }
+primefield = { version = "0.14.0-rc.1", optional = true }
 primeorder = { version = "0.14.0-rc.1", optional = true }
 serdect = { version = "0.4", optional = true, default-features = false }
 sha2 = { version = "0.11.0-rc.3", optional = true, default-features = false }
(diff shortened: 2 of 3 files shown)


tarcieri:
> Inversions are implemented by first converting from FieldElement to its canonical representative (i.e. big endian byte serialization), then parsing those bytes as a MontyFieldElement, which converts to a saturated Montgomery form representation. Once the inversion is complete, it uses a similar process to convert back.

@fjarri any thoughts on a more efficient way to do this conversion?

andrewwhitehead:
It looks like the two field representations could implement a `to_u256()` which simply normalized and eliminated the extra bits, then that could be passed into `ConstMontyForm::new`. The overhead of encoding to bytes and decoding is probably measured in nanoseconds though, so I don't think it would have a big effect on the total time. OTOH this could probably skip converting to Montgomery form and back by using `Uint::invert_odd_mod` directly.

tarcieri:
@andrewwhitehead `MontyFieldElement::from_bytes` performs a reduction. AFAIK `ConstMontyForm` does not have a public constructor that avoids one aside from `from_montgomery(_words)` which we could potentially use here (sidebar: maybe those should have more prominent warnings)

andrewwhitehead:
For this specific use case, I don't think it's necessary to bring in `primefield`/`MontyFieldElement` at all. The field representations here don't seem to be in Montgomery form so the inversion could just use `Uint::invert_odd_mod` which should be just as efficient as `ConstMontyForm::invert`. Of course if the goal is to use more of the modular operations in crypto-bigint over time then it makes sense to stick with this approach.

tarcieri:
I did indeed kind of just want an excuse to wire up `primefield` inside of `k256`, but it would be interesting to see if `invert_odd_mod` could avoid a conversion into Montgomery form/reduction

andrewwhitehead:
I tested it out and it seems to be about 2-3% faster without the conversion. Implementing `from_bytes` and `to_bytes` using the U256 conversion is also a little faster and eliminates some code. I could add those changes (not touching inversion) in a separate PR.

tarcieri:
@andrewwhitehead it'd be great if you can open a PR, yeah

tarcieri:
Closing in favor of #1554

PR #1579, opened 2026-01-05
p256: remove redundant variable names in hash2curve

sashaodessa:
Remove unnecessary `_bytes` suffix from variables in `FieldElement::reduce` to match the coding style used throughout the codebase (variable shadowing).


tarcieri:
I think if anything this is less clear, and it's just bikeshedding variable names

PR #1587, opened 2026-01-06
primefield: add `bench_field!` macro; use in `bp256`

tarcieri:
This adds a macro to the `primefield::dev` macro for writing field benchmarks, and uses them to write a basic set of benchmarks for the `FieldElement` and `Scalar` types in `bp256`.

We should roll these out to the other crates that currently lack benchmarks, but first it would be nice to augment them with ones for point operations, namely scalar multiplication.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,6 +144,7 @@ dependencies = [
 name = "bp256"
 version = "0.14.0-rc.3"
 dependencies = [
+ "criterion",
  "ecdsa",
  "elliptic-curve",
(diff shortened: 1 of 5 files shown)


tarcieri:
I've confirmed these macros are forward-compatible with `criterion` v0.8, meaning we can bump the criterion version at the individual crate level when we bump MSRV without it being a breaking change.

(Otherwise I might've considered making this a `struct FieldBencher` rather than a macro)

PR #1620, opened 2026-01-20
chore(deps): bump `rand_core` to `0.10.0-rc-5`

baloo:
This also bumps `crypto-bigint` to `0.7.0-rc.21`.

### DIFF
--- a/k256/src/arithmetic/field.rs
+++ b/k256/src/arithmetic/field.rs
@@ -37,10 +37,11 @@ use core::{
     ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
 };
 use elliptic_curve::{
+    Generate,
     bigint::{Odd, U256, modular::Retrieve},
     ff::{self, Field, PrimeField},
     ops::Invert,
-    rand_core::TryRngCore,
+    rand_core::{TryCryptoRng, TryRngCore},
     subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption},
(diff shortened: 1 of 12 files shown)


tarcieri:
> I've swapped out the use of getrandom in favor of rand as it's a tiny bit more ergonomic.

Ugh, I specifically moved from `rand` back to `getrandom` to avoid the entangling dependency.

But also, if you want something more ergonomic, you should switch to the `Generate` trait and its `getrandom`-based methods: https://docs.rs/crypto-common/0.2.0-rc.11/crypto_common/trait.Generate.html#method.generate

baloo:
Sorry about that ^^

Went with `Generate` :)

k256/src/arithmetic/field.rs
```diff
@@ -702,8 +710,8 @@ mod tests {
 
     #[test]
     fn batch_invert_array() {
-        let k: FieldElement = FieldElement::random(&mut SysRng.unwrap_mut());
-        let l: FieldElement = FieldElement::random(&mut SysRng.unwrap_mut());
+        let k: FieldElement = FieldElement::try_generate_from_rng(&mut SysRng).unwrap();
```
tarcieri, line 713: If you gate these tests on `#[cfg(feature = "getrandom")]` then you can just do:

```suggestion
        let k: FieldElement = FieldElement::generate();
```

tarcieri:
@baloo can you also bump `ecdsa` to v0.17.0-rc.13?

tarcieri:
Eh, this is fine as a start

PR #1623, opened 2026-01-20
primefield: deduplicate MontyFieldElement::is_zero in ff::Field impl

prestoalvarez:
Deduplicate zero-check logic**: `ff::Field::is_zero` now delegates to the existing `MontyFieldElement::is_zero` implementation (`primefield/src/monty.rs`).


tarcieri:
Not helpful

PR #1655, opened 2026-02-05
fix: align bignp256 pkcs8 feature gating

Forostovec:
Fixed inconsistent pkcs8 integration in bignp256 where pkcs8-dependent code in secret_key/public_key was compiled unconditionally while the pkcs8 dependency itself is optional. This could break builds when only arithmetic is enabled or when pem was enabled without the crate-level pkcs8 feature.


makavity:
Thank you!

bignp256/Cargo.toml
```diff
@@ -51,7 +51,7 @@ arithmetic = ["dep:primefield", "dep:primeorder", "elliptic-curve/arithmetic"]
 bits = ["arithmetic", "elliptic-curve/bits"]
 ecdsa = ["arithmetic", "dep:rfc6979", "dep:signature", "dep:belt-hash"]
 getrandom = ["elliptic-curve/getrandom"]
-pem = ["pkcs8/pem", "sec1/pem"]
+pem = ["pkcs8", "pkcs8/pem", "sec1/pem"]
```
tarcieri:
```suggestion
pem = ["pkcs8/pem", "sec1/pem"]
```

tarcieri (approved):

PR #1691, opened 2026-03-19
ci: fix YAML nits, use zizmor to pin dependencies

tob-scott-a:


tarcieri:
Outside the special "publish" environment/workflow, we only allow Actions read-only access.

Therefore outside that special environment, I'm not sure how helpful it actually is to pin dependencies used by Actions itself. It leads to a lot of Dependabot spam when they're updated, and I'm not sure any of the threats that would be mitigated in a such a read-only CI environment are particularly concerning.

(If there are ways to perform non-read-only requests from Actions, that's a separate issue that's worth mitigating)

tarcieri:
It'd still be good to harden the publish workflow

PR #1707, opened 2026-04-01
[WIP] Transition from `subtle` to `ctutils`

tarcieri:
Companion PR to RustCrypto/traits#2351


tarcieri:
Will re-evaluate this come the `elliptic-curve` v0.15 series

PR #1715, opened 2026-04-10
Initial `MulVartime` impls

tarcieri:
Companion PR to RustCrypto/traits#2379

This adds initial impls of the `MulVartime` trait which are required by the bounds added in the PR above.

These don't yet use variable-time implementations as noted in the TODOs, however the idea is we can opportunistically plug in wNAF when the `alloc` feature is enabled. However, actually implementing that has been saved for a follow-up.

This also adds an impl of the new `PointWithBasepointTable` to `k256`, which makes the table accessible in a generic context.


tarcieri:
Note: still need to do `ed448-goldilocks`

PR #1741, opened 2026-04-17
Adopt Trusted Publishing

tarcieri:
See https://crates.io/docs/trusted-publishing

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,53 @@
+name: Publish to crates.io
+on:
+  push:
+    tags: [
+      "bignp256/v**",
+      "bp256/v**",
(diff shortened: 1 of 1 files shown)


newpavlov (approved):
Also hash-pin the dependencies?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
+name: Publish to crates.io
+on:
+  push:
+    tags: [
+      "bignp256/v**",
```
newpavlov, line 5: Is double `**` needed?

tarcieri: I believe I added it due to the `/` in the tag names, and it wasn't working without it

PR #1766, opened 2026-06-02
Remove `bits` features

tarcieri:
Upstream support was removed in RustCrypto/traits#2417.

See also: zkcrypto/ff#157


kayabaNerve:
Ugh. I really hope releases don't get cut with this. I need `to_le_bits` for a _lot_ of my work :/

I do see an RFC was made, it just means that `elliptic-curves` needs to move to RustCrypto's `ff` OR it needs to flow through `ff` for me to be able to update.

PR #1782, opened 2026-06-11
primeorder: use `elliptic_curve::point::LookupTable`

Pranavjeet-Naidu:
Fixes #1657
- Add `Radix16Decomposition` in primeorder: generic signed radix-16 digit
  decomposition before the main multiplication loop, matching k256's approach.
- Refactor `ProjectivePoint::mul` and `lincomb` to use
  `elliptic_curve::point::LookupTable` (8-entry signed) instead of the private
  16-entry unsigned table.
- Decode scalars via `Into::<Uint>::into(scalar).to_be_byte_array()` so
  decomposition is correct for LE `FieldBytes` curves (bignp256) and curves
  whose `Uint` is wider than the scalar field (p224, p521).

No new dependencies; radix16 edge cases are covered by unit tests in primeorder
and by existing lincomb proptests in the curve crates.

### DIFF
--- a/primeorder/src/radix16.rs
+++ b/primeorder/src/radix16.rs
@@ -1,0 +1,206 @@
+//! Radix-16 signed-digit decomposition for constant-time scalar multiplication.
+
+use core::ops::Add;
+
+use elliptic_curve::{
+    CurveArithmetic, PrimeCurve, PrimeField, Scalar,
+    array::typenum::{Prod, U1, U2, Unsigned},
+    bigint::ArrayEncoding,
+    consts::U66,
+};
+
+/// Largest scalar `FieldBytes` size among primeorder curves (P-521).
+type MaxScalarFieldBytes = U66;
+
+/// Two nibbles per scalar byte, plus one carry digit for signed recentering.
+type MaxDigits = <Prod<MaxScalarFieldBytes, U2> as Add<U1>>::Output;
+
+const MAX_DIGITS: usize = MaxDigits::USIZE;
+
+/// Signed radix-16 decomposition of a scalar
+/// Produces `[a_0, ..., a_{len-1}]` such that
+/// `scalar = sum(a_j * 16^j)` and each `a_j` is in `[-8, 8]`.
+/// `a_0` is the least significant position; `a_{len-1}` absorbs carry.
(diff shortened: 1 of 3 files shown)


primeorder/src/radix16.rs
```diff
@@ -0,0 +1,194 @@
+//! Radix-16 signed-digit decomposition for constant-time scalar multiplication.
+
+use elliptic_curve::{CurveArithmetic, PrimeCurve, PrimeField, Scalar, bigint::ArrayEncoding};
+
+/// Worst case: p521 (66-byte scalars) → 132 nibbles + 1 carry digit.
+const MAX_DIGITS: usize = 133;
```
tarcieri: This size should probably be computed from e.g. `FieldBytesSize`, rather than hardcoded

tarcieri: FWIW, it's the same size as an uncompressed SEC1 point, though I would probably prefer the `typenum` arithmetic to calculate it be duplicated as opposed to relying on that

Pranavjeet-Naidu: That does make sense yes, thank you
It will be derived from: 
```rust
<Prod<MaxScalarFieldBytes, U2> as Add<U1>>::Output;
```

primeorder/src/radix16.rs
```diff
@@ -0,0 +1,206 @@
...
+    array::typenum::{Prod, U1, U2, Unsigned},
+    bigint::ArrayEncoding,
+    consts::U66,
+};
+
+/// Largest scalar `FieldBytes` size among primeorder curves (P-521).
+type MaxScalarFieldBytes = U66;
+
+/// Two nibbles per scalar byte, plus one carry digit for signed recentering.
+type MaxDigits = <Prod<MaxScalarFieldBytes, U2> as Add<U1>>::Output;
```
tarcieri, line 16: I meant using `C::FieldBytesSize` to size `digits` in `Radix16Decomposition` on a curve-by-curve basis as opposed to hardcoding a max, but it's okay, I can fix it in a followup.

tarcieri (approved):
Will do some followups on this, but it gets the basic implementation in place

PR #1800, opened 2026-06-15
primeorder: extract `ModulusSize` marker trait

tarcieri:
This makes the `PrimeCurveParams` bounds slightly less atrocious.

Extracts a marker trait with a blanket impl for all types that satisfy its supertrait bounds (usable as `Radix16Digits` and `sec1::ModulusSize`)


tarcieri:
This isn't needed anymore after https://github.com/RustCrypto/traits/pull/2448

PR #1803, opened 2026-06-16
primeorder: add `Backend` trait and associated type

tarcieri:
Similar to `PrimeCurveParams::PointArithmetic`, this adds a `PrimeCurveParams::Backend` associated type bounded by a new `Backend` trait.

Also adds two ZSTs `backend::VariableOnly` and `backend::PrecomputedTables`, where the latter is bounded on the `PrimeCurveWithBasepointTable` trait to access the basepoint table.

The specific ZST is select based on whether the `precomputed-tables` feature is enabled.

This provides a single place to generically define the interactions between the basepoint table for a particular curve and the scalar multiplication implementation in `primeorder`.

Previously there were a kitchen sink of small methods that could be overridden on `PrimeCurveParams` which have been removed, making the `PrimeCurveParams` declarative once again.

--

Unfortunately, due to rust-lang/rust#140653 the generic `PrecomputedTables` implementation is unusable until MSRV 1.90, as the `PrimeCurveWithBasepointTable::BASEPOINT_TABLE` constant cannot be accessed in Rust 1.85-1.89:

        error[E0080]: it is undefined behavior to use this value

This error occurs due to the interior mutability of `LazyLock` which made such values unusable in patterns. The solution was to permit a `const` that references a `static` with interior mutability, but disallow use in patterns, which shipped in Rust 1.90.

As a workaround for now, every crate with a basepoint table defines its own internal `PrecomputedTables` backend specific to that curve which uses its concrete `BASEPOINT_TABLE` constants. We can remove these hacks and switch to the generic version in Rust 1.90.

### DIFF
--- a/bignp256/src/arithmetic.rs
+++ b/bignp256/src/arithmetic.rs
@@ -13,6 +13,7 @@ pub use self::scalar::Scalar;
 pub use self::field::FieldElement;
 use crate::BignP256;
 pub use elliptic_curve::{CurveArithmetic, PrimeCurveArithmetic};
+use primeorder::backend;
 pub use primeorder::{PrimeCurveParams, point_arithmetic};
 
 /// Elliptic curve point in affine coordinates.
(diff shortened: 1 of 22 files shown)


tarcieri:
Oh that's fun, the associated `BASEPOINT_TABLE` constant can't be used in 1.85 due to interior mutability, even though it's a constant reference to a `&'static`

PR #1840, opened 2026-06-22
[WIP] primefield: add "fused" ops to `FieldExt`

tarcieri:
Exploring what might make sense here. Starting with:

- `add_mul`: (a + b) * c
- `sub_mul`: (a - b) * c

The only field backend where I've done an optimization for these is `p521`. Notably its multiply takes a `LooseFieldElement` so there is further potential for another add/sub operation in place of `c`, e.g. `(a + b) * (c + d)`.

Before going any farther down this road it would be good to look at:
- what works well for the `k256` field element implementation
- what we could potentially add to `primefield::MontyFieldElement` in a generic implementation

### DIFF
--- a/p521/src/arithmetic/field.rs
+++ b/p521/src/arithmetic/field.rs
@@ -531,7 +531,16 @@ impl Field for FieldElement {
     }
 }
 
-impl FieldExt for FieldElement {}
+impl FieldExt for FieldElement {
+    fn add_mul(&self, x: &Self, y: &Self) -> Self {
+        self.add_loose(x).multiply(&y.relax())
+    }
+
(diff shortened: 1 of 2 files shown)


tarcieri:
I'll note whether you do an `add`/`sub`-style operation first versus a `mul` is important to how these fusions would get leveraged in an algorithm like RCB, so I don't think the right answer is we should provide both `add_mul` and `mul_add` because e.g. `p521` can't optimize `mul_add` but can optimize `add_mul`, so it doesn't help if a downstream user thinks `mul_add` is more convenient only to discover it offers no performance benefits.

PR #1872, opened 2026-06-29
Migrate to new `rfc6979::KGenerator` API

tarcieri:
This tests out the substantial changes made to the `rfc6979` crate in RustCrypto/signatures#1395, including a move to a new stateful API.

Most of these were behind the scenes but there are a couple important ones which impact all crates:
- `ecdsa::hazmat::DigestAlgorithm` => `ecdsa::DigestAlgorithm`
- There is no longer a `hazmat` feature on `ecdsa`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,8 +428,7 @@ dependencies = [
 [[package]]
 name = "ecdsa"
 version = "0.17.0-rc.22"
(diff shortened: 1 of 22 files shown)


baloo (approved):
Tested downstream crates.

PR #1877, opened 2026-06-29
`bignp256`: migrate to new `bign-genk` api

makavity:


tarcieri:
Heh, just tagged a release but we'll get it in the next one, sorry! 😅

makavity:
No problem, it is just rc :D

makavity:
Also, is that ok with loop in signing.rs?

tarcieri:
It's fine for now although ideally it would retry the entire signature operation, which enables infallible signing (e.g. in the event `r` is `0`)

PR #1897, opened 2026-07-07
k256: remove duplicate ecdsa CI build

weifanglab:
## Summary

- Remove a duplicate `k256` CI build step for `--no-default-features --features ecdsa`
- Keep the existing standalone `ecdsa` build and the `ecdsa,sha256` coverage intact

## Rationale

The `k256` workflow was running the exact same ARM/no_std `ecdsa` build command twice:

```text
cargo build --target thumbv7em-none-eabi --release --no-default-features --features ecdsa


tarcieri:
Thanks!

PR #1910, opened 2026-07-30
fix `cargo doc` E0080 on MSRV 1.85 (p256/p384/p521/sm2/k256)

### LINKED ISSUE #1909: cargo doc fails with E0080
`cargo doc` fails on `p256 0.14.0`, `p384 0.14.0`, `p521 0.14.0`, `k256 0.14.0`, and `sm2 0.14.0-rc.15` with Rust 1.85-1.89.
Projects using Rust 1.85-1.89 that have dependencies on these are failing when executing `cargo doc --document-private-items`.

The following is the error message you receive from `cargo doc --document-private-items`:
```text
Documenting p256 v0.14.0 (/repo/p256)
error[E0080]: it is undefined behavior to use this value
  --> p256/src/arithmetic/tables.rs:17:5
   |
17 |     const BASEPOINT_TABLE: &'static BasepointTable = &BASEPOINT_TABLE;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ constructing invalid value: encountered reference to mutable memory in `const`
   |
   = note: The rules on what exactly is undefined behavior aren't clear, so this check might be overzealous. Please open an issue on the rustc repository if you believe it should not be considered undefined behavior.
   = note: the raw bytes of the constant (size: 8, align: 8) {
               ╾─────alloc456894─────╼                         │ ╾──────╼
           }

For more information about this error, try `rustc --explain E0080`.
error: could not document `p256`
```

`cargo build` and `cargo test` succeed — only `cargo doc` fails (rustdoc evaluates the associated const to render its value).

**Reproduction** (Rust 1.85-1.89):
```bash
docker run --rm rust:1.85.0 bash -c '
  apt-get update -qq && apt-get install -y -qq git >/dev/null
  git clone --depth 1 https://github.com/RustCrypto/elliptic-curves.git /repo
  cd /repo/p256
  cargo doc --document-private-items
'
```

Warfront1:
Fixes #1909.

`cargo doc` fails on the published `p256`/`p384`/`p521`/`sm2`/`k256` 0.14.x crates when building with Rust 1.85-1.89 (`E0080` at `src/arithmetic/tables.rs`).

This adds `#[doc(hidden)]` to the offending associated const in all five affected crates (p256, p384, p521, sm2, k256).  
Hiding the const from rustdoc prevents the const evaluator from running on it, resolving `E0080` while keeping MSRV 1.85.  
The const remains usable at runtime; only its doc rendering is suppressed.

```bash
docker run --rm rust:1.85.0 bash -c '
  apt-get update -qq && apt-get install -y -qq git >/dev/null
  git clone --depth 1 --branch feat/fix-doc-e0080-rust140653 https://github.com/Warfront1/elliptic-curves.git /repo
  cd /repo
  cargo doc --workspace --all-features
'
```

The underlying rustc issue ([rust-lang/rust#140653](https://github.com/rust-lang/rust/issues/140653)) is fixed in Rust 1.90 ([rust-lang/rust#140942](https://github.com/rust-lang/rust/pull/140942)).  
When MSRV is bumped to 1.90, this `#[doc(hidden)]` workaround can be removed.

### DIFF
--- a/k256/src/arithmetic/tables.rs
+++ b/k256/src/arithmetic/tables.rs
@@ -18,5 +18,7 @@ type BasepointTable = primeorder::BasepointTable<ProjectivePoint, WINDOW_SIZE>;
 pub(super) static BASEPOINT_TABLE: BasepointTable = BasepointTable::new();
 
 impl PrimeCurveWithBasepointTable<WINDOW_SIZE> for Secp256k1 {
+    /// Workaround for rust-lang/rust#140653: hide from rustdoc const-eval (E0080) until MSRV 1.90
+    #[doc(hidden)]
     const BASEPOINT_TABLE: &'static BasepointTable = &BASEPOINT_TABLE;
 }
--- a/p256/src/arithmetic/tables.rs
+++ b/p256/src/arithmetic/tables.rs
@@ -14,6 +14,8 @@ pub(super) type BasepointTable = primeorder::BasepointTable<ProjectivePoint, WIN
 pub(super) static BASEPOINT_TABLE: BasepointTable = BasepointTable::new();
 
 impl PrimeCurveWithBasepointTable<WINDOW_SIZE> for NistP256 {
+    /// Workaround for rust-lang/rust#140653: hide from rustdoc const-eval (E0080) until MSRV 1.90
+    #[doc(hidden)]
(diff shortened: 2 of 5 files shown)


tarcieri:
I don't want to introduce hacks like this just for `cargo doc`. Please use a newer version of Rust if you want this to work.

PR #1913, opened 2026-08-13
wnaf: handle scalar representation endianness

questfever:
## Summary

Fix variable-time scalar multiplication for curves whose `PrimeField::Repr` is little-endian.

`WnafScalar::new()` assumes that `PrimeField::to_repr()` is big-endian and unconditionally reverses it. Since representation endianness is implementation-specific and `bignp256::Scalar` is already little-endian, this caused its scalars to be interpreted with the wrong value.

This change:

- converts primeorder scalars with `PrimeFieldExt::to_le_repr()`
- constructs wNAF scalars using `WnafScalar::from_le_bytes()`
- applies the conversion to `mul_vartime` and both `lincomb_vartime` paths
- adds bignp256 regression tests for scalar multiplication and linear combinations

## Testing

- `cargo test -p bignp256 --all-features`
- `cargo test -p bignp256 --no-default-features --features arithmetic,test-vectors`
- `cargo test -p primeorder --all-features`
- `cargo clippy -p bignp256 --all-targets --all-features -- -D warnings`
- `cargo clippy -p primeorder --all-targets --all-features -- -D warnings`
- `cargo fmt --all -- --check`

### DIFF
--- a/wnaf/src/boxed.rs
+++ b/wnaf/src/boxed.rs
@@ -4,6 +4,7 @@ use crate::{Digit, WnafGroup, le_repr, wnaf_form, wnaf_multi_exp, wnaf_table};
 use alloc::vec::Vec;
 use ff::PrimeField;
 use group::Group;
+use primefield::PrimeFieldExt;
 
 #[cfg(doc)]
 use crate::{WnafBase, WnafScalar};
@@ -126,7 +127,10 @@ impl<G: WnafGroup> BoxedWnaf<(), Vec<G>, Vec<Digit>> {
     }
 
     /// Construct wNAF context for `scalar`.
-    pub fn scalar(&mut self, scalar: &G::Scalar) -> BoxedWnaf<usize, &mut Vec<G>, &[Digit]> {
+    pub fn scalar(&mut self, scalar: &G::Scalar) -> BoxedWnaf<usize, &mut Vec<G>, &[Digit]>
+    where
+        G::Scalar: PrimeFieldExt,
+    {
         let window_size = 4;
 
         let repr = le_repr(scalar);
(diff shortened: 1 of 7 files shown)


tarcieri:
This should be addressed through the existing `le_repr` method, where there is already a TODO to add proper endianness handling support.

questfever:
> This should be addressed through the existing `le_repr` method, where there is already a TODO to add proper endianness handling support.

Thanks for the clarification. I understand now: my original change worked around the issue in primeorder, but left the endianness assumption and the existing TODO in wnaf::le_repr unchanged.

I've reworked the patch so that le_repr uses PrimeFieldExt::to_le_repr directly. WnafSize now requires PrimeFieldExt, and the two BoxedWnaf scalar entry points carry the same bound. I also removed the primeorder-specific helper, so primeorder once again uses WnafScalar::new directly.

The bignp256 regression tests are retained and cover mul_vartime as well as the fixed-size and allocating lincomb_vartime paths.

questfever:
@tarcieri Modified. Please review it again. Thanks. 

 If there's anything need I to do, please feel free to let me know.

wnaf/src/boxed.rs
```diff
@@ -126,7 +127,10 @@ impl<G: WnafGroup> BoxedWnaf<(), Vec<G>, Vec<Digit>> {
     }
 
     /// Construct wNAF context for `scalar`.
-    pub fn scalar(&mut self, scalar: &G::Scalar) -> BoxedWnaf<usize, &mut Vec<G>, &[Digit]> {
+    pub fn scalar(&mut self, scalar: &G::Scalar) -> BoxedWnaf<usize, &mut Vec<G>, &[Digit]>
+    where
+        G::Scalar: PrimeFieldExt,
```
tarcieri, line 132: Hmm, these bounds changes probably don't matter in practice but are technically breaking

tarcieri (approved):

